PR #799, opened 2022-02-04
Axios catch and throw

TatisLois:
# Why

Axios currently will swallow any caught errors. This behavior prevents the application from properly reacting to happy/unhappy paths.

# How

1. Continued to catch in axios. Helpful for debugging + sending logs (in the future)
2. Throw after caught with the original error

# Test Plan

1. Forced errors to happen on local. 
2. Combed code and made sure that anywhere that was not catching correctly is also updated. (separate PR)

### DIFF
--- a/packages/app/lib/axios.ts
+++ b/packages/app/lib/axios.ts
@@ -101,8 +101,8 @@ const axiosAPI = async ({
     return await axios(request).then((res) => res.data);
   } catch (error) {
     console.log("failed request ", request);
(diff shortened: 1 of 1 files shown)


intergalacticspacehighway (approved):
Oh nice, here we have a log so I don't think we'll need [this](https://github.com/tryshowtime/showtime-frontend/pull/800#pullrequestreview-874237583)

PR #800, opened 2022-02-04
Update like + unlike to catch thrown errors

TatisLois:
# Why

With [axios](https://github.com/tryshowtime/showtime-frontend/pull/799) throwing errors that it's caught. The `like` and `unlike` functions need to update its error handling as axios won't swallow anymore and there is no guarantee the calls will succeed. 

# How

1. Wrap the calls in a try/catch
2. `like`
  - Happy path stays the same
  - Error path will mutate and return false.
3. `unlike`
  - refactored to have the same return style as `like` (true or false)
  - Only on confirmed happy path will it decrement by 1 
  - Error path will mutate and return false.

# Test Plan

On local
1. Forced fail adding a like for an NFT
2. Forced failed removing a life from an NFT

Small video of the force fail (happens fast because it's a force fail)

https://user-images.githubusercontent.com/11730651/152602547-4dbcccfa-e972-407c-8c86-387e445ff2ae.MP4

### DIFF
--- a/packages/app/hooks/api-hooks.ts
+++ b/packages/app/hooks/api-hooks.ts
@@ -350,25 +350,30 @@ export const useMyInfo = () => {
       }
 
       if (data) {
-        mutate(
-          queryKey,
-          {
-            data: {
-              ...data.data,
-              likes_nft: [...data.data.likes_nft, nftId],
+        try {
+          mutate(
+            queryKey,
+            {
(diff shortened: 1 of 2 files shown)


intergalacticspacehighway (approved):
LGTM. I think we should add a logger at all catch blocks. Any thoughts?

PR #805, opened 2022-02-07
fix: removed refresh token error screen

gorhom:
# Why

Currently when trying to refresh access token for unlogged in customer, the logic will throw an error screen _ONLY FOR DEBUG_, which is not necessary. 

https://showtime-rq88331.slack.com/archives/C02RX6JUKLY/p1644255140713819

# How

- change refresh token error from throwing an error to just print in app console.

# Test Plan

- install and launch the app
- app should print refresh token error in the console


axeldelafosse (approved):
LGTM ✨

PR #811, opened 2022-02-10
Fix current pill matching

TatisLois:
# Why

Quick fix to the `Current` pill to only display on the active address

# How

1. Update `isConnectedAddress` to check current address with the address being passed into the slot

# Test Plan

1. Log into multiple addresses connected addresses and make sure the current pill matches correctly


intergalacticspacehighway (approved):
LGTM ✨

PR #813, opened 2022-02-10
Listing An NFT

TatisLois:
# Why

Grant users the ability to list an NFT to the showtime marketplace. Under the constraint that it was minted on our contract, the user granted listing approval and are connected to the minted address.

# How

1. Create listing route and screen
2. Added the `ShowtimeV1Market` abi
3. Added as constants the supported currency contract addresses + symbols by chainId
4. Created types for listing and sellers
5. Created a few utilities to help with listing and possibly other web3 actions
  - findListingItemByOwner - returns the listing data of the listing owner
  - isUserAnOwner - checks if the _user_ owns and edition (across all their addresses)
  - findUserInOwnerList - returns an array of type `WalletAddressesV2` for all addresses a user has connected that own an edition
  - findAddressInOwnerList - returns an object of type `OwnersListOwner` if the passes in address owns an edition of the NFT
  - parseBalance - formatting for different currencies
6. Support for `Creator` title in design system creator element
7. In NFT dropdown the listing option only shows if -> User owns an edition + showtime contract that minted the nft is configured
8. In NFT dropdown added the inverted for delisting (empty for now)
9. Support for listing screen for when a user owns an edition but not on the active address
10. listing hook to manage listing flow and expose the current listing status
11. Redirect on successful listing similar to minting

# Test Plan

Some issues with testing until local minting is fixed

### DIFF
--- a/packages/app/abi/ShowtimeV1Market.json
+++ b/packages/app/abi/ShowtimeV1Market.json
@@ -1,0 +1,412 @@
+[
+  {
+    "inputs": [
+      { "internalType": "address", "name": "_nft", "type": "address" },
+      {
+        "internalType": "address",
+        "name": "_trustedForwarder",
+        "type": "address"
+      },
+      {
+        "internalType": "address[]",
+        "name": "_initialCurrencies",
+        "type": "address[]"
+      }
+    ],
+    "stateMutability": "nonpayable",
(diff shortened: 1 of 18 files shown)


intergalacticspacehighway (approved):
LGTM 💯

PR #831, opened 2022-02-15
Support magic testmode for a configurable email

TatisLois:
# Why

Google store rejection feedback highlighted that we can provide a demo account so that they can preview required UGC features like blocking. Since our flow is passwordless it would reduce approval friction if we just bypassed that part of the login flow _only_ under these conditions. We can always remove/approve it after approval. 

# How

1. Create two constants. The first is the input email to check enablement against. The second is the address we override the inputed one with (has a key/value pair of a burner address). We might not _need_ the key/value burner address but if they want to do a web3 thats how we can offer support.
2. Create a utility function that if the conditions are true will create a new magic instance with `testMode` true
3. In the email login in `packages/app/hooks/auth/use-magic-login.ts` we use the utility and constants

This should be backwards compatible with our current flow.

### Blocker

- [x] waiting on backend updates to support `/v1/login_magic`

# Test Plan

1. Tested current flow with these changes
2. Partial testing of the testMode flow (waiting on remaining changes)

## Test it out yourself
1. Log in with magic as normal, this should not change the flow on magic
2. Log in using `test+success@magic.link` you should not be prompt to check an e-mail and should be dropped right into the app. Some web3 features (mint) might be spotty but imo a non issue for the reviewers. 

https://user-images.githubusercontent.com/11730651/154333452-2bbbfe59-0bf2-49c5-8d43-c5d718d9a80a.MP4

### DIFF
--- a/packages/app/hooks/auth/use-magic-login.ts
+++ b/packages/app/hooks/auth/use-magic-login.ts
@@ -1,7 +1,9 @@
 import { useCallback } from "react";
 
+import { BYPASS_EMAIL_WITH_INSECURE_KEYS } from "app/lib/constants";
 import { magic } from "app/lib/magic";
 import { mixpanel } from "app/lib/mixpanel";
+import { overrideMagicInstance } from "app/utilities";
 
 import { useAuth } from "./use-auth";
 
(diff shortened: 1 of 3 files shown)


intergalacticspacehighway (approved):
🤩

axeldelafosse (commented):
How do we feel about this email being committed in git? If we open source it's not good 🤔

PR #842, opened 2022-02-17
fix(utilities): import magic correctly

TatisLois:
# Why

Incorrect imports break the web application

# How

update direct import to import from the `lib` that knows how to handle native vs web imports

# Test Plan

1. Retested regular magic logins
2. Retested demo account
3. Ran the react native web app


axeldelafosse (approved):
LGTM ✨

PR #854, opened 2022-02-25
WIP Feature/show2 511 GitHub action in staging to

TatisLois:
# Why

To help reduce mobile application release complexity and give back time to engineers while opening up releases to anyone this PR configured a script with logic to properly version, commit the latest version of the mobile application. When triggered the github action will preform the step before the Expo release step.

# How

This is a piece that is part of a larger release flow. For context 
1. A showtime [slack application](https://showtime-rq88331.slack.com/apps/A03660EPVV1-deploy?settings=1) was configured to accept a `/promote` command. The command takes inputs that map to action down stream. The full release command for mobile is `/promote mobile`
2. The slack promote command will trigger a google cloud function.
- That was setup as a stand alone project in this [git repo](/slack-promote). By default google cloud functions are written inline on the cloud console but by moving it to it's own file/repo we can iterate and improve it quicker and safer. Deployments are automated for it through it's own github actions.
- The cloud function has a secret token provision to trigger github action workflows. It maps the input from the promote command to an object with relevant context to trigger the right workflow
- Response back to slack using the slack block kit api for nicer formats 
3. Cloud function will trigger the `expo-production` workflow through the `on: workflow_dispatch` specifying the ref branch to be `staging`. 
4. `expo-production` workflow will continue to behave as-is except the `Set up repository` step will also update some config settings. This should only impact the version bump step not any other steps.
5. The `Application Version Bump` step will only run if the condition is met (skipped if not). It will run the script
6. The script will 
- Get the current and last release commit ID (if they are the same then bail)
- Diff the two commit ID's and check if the expo package.json was modified 
- If not then we assume all changes are JS related and patch/commit/push back up
- If the package.json was modified we do some cleanup to get a list of all the new `keys`
- We run the keys against a configuration that we generate from `react-native config` 
- The return of that check tells us if we should patch or major then commit/push back up
7. Any action that changes git/files etc.. is wrapped in a `inCI` variable that is only true on the CI.

Right now the cloud function is pointing to this branch and the `test-slack` workflow. I left the `test-slack` workflow for reference but we should remove it after the real expo flow is tested. I also have a PR open to point the cloud function to the right workflow.

Cloud function ref PR: https://github.com/tryshowtime/slack-promote/pull/5

# Flow sketch
flow

# Test Plan

Tested locally (safely with the inCI var)
Tested by pushing and having the cloud function pointed to this branch (needs to be updated)

### [Example Job On CI](https://github.com/tryshowtime/showtime-frontend/runs/5580998098?check_suite_focus=true)

Screen Shot 2022-03-17 at 12 38 37 AM

### Ran locally
Screen Shot 2022-03-16 at 11 50 10 PM

### Slack Screenshots

Screen Shot 2022-03-17 at 12 17 51 AM
Screen Shot 2022-03-17 at 12 24 22 AM


axeldelafosse (approved):
LGTM ✨Nice work!

One last thing missing is that we probably don't want to trigger a new build (last step in `expo-production` `Build on EAS`) if we just did a patch update, we just need an OTA update.

What happens if we want to manually trigger a major update (so we upgrade manually)?

I'm also wondering what would happen now because we manually upgraded to v15.

PR #861, opened 2022-03-03
feature(mint): set royalties for magic users to zero

TatisLois:
# Why

[Linear Task](https://linear.app/showtime/issue/SHOW2-554/enforce-0percent-royalties-on-magic-wallets)

Magic users do not have full ownership of their address. So while on mobile we allow web3 actions the team decided to minimize the security risk we are programmatically setting minting royalties to 0 for magic users. 

# How

1. hide the input from the mint options
2. set the value to zero on mint

# Test Plan

Ran it locally on my phone

Screen Shot 2022-03-02 at 10 45 54 PM

### DIFF
--- a/packages/app/components/create.tsx
+++ b/packages/app/components/create.tsx
@@ -11,6 +11,7 @@ import {
   UseMintNFT,
 } from "app/hooks/use-mint-nft";
 import { useUser } from "app/hooks/use-user";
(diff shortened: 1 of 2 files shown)


intergalacticspacehighway (approved):
LGTM 🚀

PR #864, opened 2022-03-07
fix(activity): format issue with timestamps

TatisLois:
# Why

Breaking the home page on safari, see [slack thread](https://showtime-rq88331.slack.com/archives/C02Q0E4PBD0/p1646669351711799) caused by a similar issue experienced on mobile as described on this [slack thread](https://showtime-rq88331.slack.com/archives/C02PTPFCZAA/p1646347601480709)

# How

1. remove additional char

# Test Plan

Ran the application locally on safari, firefox and chrome


fontanierh (approved):
Thank you 🙏  My new release might actually break without this fix !

PR #867, opened 2022-03-08
Listing Copy

TatisLois:
# Why

Per feedback updating the listing copy

# How

1. If the active address does not own the NFT but the user does, update the copy on that view with support for ens domain
2. Update the copy mapping object
3. Use a derived copy func to update the CTA copy 
4. When a listing is 0 use the word `Free` 
5. In nft dropdown updated a guarding variable name

# Test Plan

Tested each view locally to make sure copy was reflected


intergalacticspacehighway (approved):
LGTM ❇️

axeldelafosse (approved):

PR #905, opened 2022-03-17
Fix/lint yarn berry

TatisLois:
# Why

Resolve issue with lint checking failing builds

# How

1. Ensure that eslint runs
2. Artifact is created/uploaded correctly

# Test Plan

Ran locally and will see if build passes


axeldelafosse (approved):
LGTM

PR #915, opened 2022-03-21
Update market place env var for expo

TatisLois:
# Why

For web3 actions that depend on the market place contract the proper contract address has to be set as an env variable.

# How

1. Referenced the smart contract readme: [existing deployments](https://github.com/tryshowtime/smart-contracts-private#existing-deployments)
1. Update the value for local
1. Add the variable for staging and production

# Test Plan

Read the env output locally and went through a listing flow


intergalacticspacehighway (approved):
LGTM ❇️

PR #939, opened 2022-03-30
Update E-mail management within settings 

TatisLois:
# Why

E-mail tab in the settings page was no adhering to the correct behavior. For reference see the [linear issue](https://linear.app/showtime/issue/SHOW2-494/update-email-tab-settings-page). This PR resolves issues within the settings e-mail tab.

# How

1. Only display `add an email` button on accounts _without_ an e-mail connected
2. Disable removing an e-mail (and address) if it's the current method the user is logged in as. 
3. Refactored to use `useWeb3` to ensure logic checks are correct and aligned with the rest of the app
4. Refactored to use `useCurrentUserAddress` instead of duplicating wallet connect logic
5. Refactored `app/hooks/use-current-user-address.ts` because as it was structured it would in most cases *not* set the magic address even after login in through magic.
- In most cases we will only have an active wallet connect address *or* an active magic address we also get the wallets array from the backend (these can't sign but are associated to the user). 
- In some scenarios it's possible to have both providers existing at the same time. Sometimes wallet connect lingers after disconnecting or the magic instance becomes orphaned... In these scenario `use-current-user-address` should return the address in the preference of 1. Wallet connect 2. Magic.
- For some reason it was instead going by 1. Wallet connect 2. 1st item in backend wallets array 3. Magic. Creating an issue that a logged in user with both an address and magic account would never set the magic address in ``use-current-user-address` 
7. Small copy update

# Test Plan

Tested locally

### Add an e-mail + remove add option after an email is linked
https://user-images.githubusercontent.com/11730651/160753347-5478748d-8ae9-4108-8c73-a5aba46da9f0.MP4

### Remove an e-mail 
https://user-images.githubusercontent.com/11730651/160753388-b2d18fa2-8976-489e-80f5-69cc9f75fa4f.MP4

### DIFF
--- a/packages/app/components/settings/address-menu.tsx
+++ b/packages/app/components/settings/address-menu.tsx
@@ -12,16 +12,17 @@ import {
 import { MoreHorizontal } from "design-system/icon";
 
 type AddressMenuProps = {
-  address?: WalletAddressesExcludingEmailV2["address"];
+  address?: WalletAddressesExcludingEmailV2["address"] | undefined | null;
   email?: WalletAddressesV2["email"];
   ctaCopy: string;
+  isCurrent: boolean;
 };
 
 export const AddressMenu = (props: AddressMenuProps) => {
   const { removeAccount } = useManageAccount();
   const address = props.address;
   const ctaCopy = props.ctaCopy;
-  const disableDelete = !address;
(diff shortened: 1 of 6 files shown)


axeldelafosse (approved):
LGTM ✨

PR #941, opened 2022-03-31
feat(SHOW2-287): comment replies

gorhom:
# Why

As part of the comments feature, the comment replies now is ready to be integrated in the app with this PR. It will also introduce the local pagination for replies until the backend implements it on their side.

# How

https://user-images.githubusercontent.com/4061838/161155395-4940e666-90f7-4415-a1e0-7d7b83f9e34c.mov

# Test Plan

- Install the app
- Login
- Select any post
- Tap on comment icon
  - customer should see the reply to comment view
- Posting a comment will refresh the comments with the sub-comment view.


axeldelafosse (approved):
LGTM ✨ Nice work, as always!

The remaining TODOs are not blocking IMO, can be implemented in a follow up PR

PR #948, opened 2022-04-05
Integrate Adding A Wallet In Settings

TatisLois:
# Why

We've had the `Add Wallet` on the settings page but non-functional. This brings support to the feature, bringing mobile closer to web feature parity.

# How

Wallet connection management can be fickle with wallet connect and the variety of logged in scenarios we support (wallet connect, magic and BE authenticated). The lack of wallet connect documentation also made it tricky to make sure our use cases are covered. I tried to take special care of the _states_ a user could end up in unintentionally.

- Added an alert prompt to confirm adding a wallet
- Created a hook w/reducer to manage the `add wallet` flow

1. Before a _new_ session can be added through wallet connect the current one has to be killed. 
  - If a current session, then kill it
  - If there isnt one (magic, error etc..) then skip. Trying to kill a session when there is no session throws an error.
2. Request a new connection with wallet connect
  - We use the response from the wallet connect request to determine if it's successful or not. `walletConnector.connected` returns false even on success. It might be a race condition but not sure.
  - If the response is valid, we invoke connect again to update the wallet connect state. This does not prompt the modal again.
3. If the connected address is new
  - Send the address to the backend, connecting it to the profile
  - Disconnect (if the user logged in from magic) magic session
  - Display a toast confirming the new wallet
4. If the address is already connected, we inform that user through a toast and it becomes the new active connection
5. Use `mutate` so state can be reflected on the client
6. In the event of an `error` ie: closing the modal after disconnecting we update the CTA copy to retry and display a toast.
  - If the user ignores they will be able to still browse the app and other web3 actions would re-trigger the connect modal

# Test Plan

Tested locally with meta mask and rainbow

## Address added
https://user-images.githubusercontent.com/11730651/161837354-1da8d789-4538-4fd8-bca3-fc896aba638a.MP4

## Address already connected
https://user-images.githubusercontent.com/11730651/161837464-02775ac1-30eb-4947-bd7f-5fcbd7f9771f.MP4

## Add wallet from magic
https://user-images.githubusercontent.com/11730651/161837065-a7c4061f-2e28-410a-82a8-0b151a60e68b.MP4

## Issue with adding wallet (connection lost)
https://user-images.githubusercontent.com/11730651/161837312-6ce4f4d5-ad4b-47a0-8a6d-8cb021c90944.mov

### DIFF
--- a/packages/app/components/settings/settings-wallet-slot.tsx
+++ b/packages/app/components/settings/settings-wallet-slot.tsx
@@ -1,7 +1,12 @@
+import { useEffect } from "react";
+import { Alert } from "react-native";
+
 import Animated, { FadeIn } from "react-native-reanimated";
 
+import { useAddWallet } from "app/hooks/use-add-wallet";
 import { useCurrentUserAddress } from "app/hooks/use-current-user-address";
 import { formatAddressShort } from "app/lib/utilities";
+import { useWalletConnect } from "app/lib/walletconnect";
 import { WalletAddressesExcludingEmailV2 } from "app/types";
 
 import { View, Text, Button, Skeleton } from "design-system";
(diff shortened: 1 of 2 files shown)


intergalacticspacehighway (approved):
LGTM ❇️

axeldelafosse (approved):
LGTM ✨

PR #957, opened 2022-04-10
fix: added pathname to useRouter

gorhom:
# Why

#956 introduce a crash on when trying to access `useRouter() -> pathname` in `Creator` & ` LikedBy` components. 

# How

I have added pathname from react navigation and next router to the use router.

# Test Plan

- Launch tha app.
- Tap on any NFT post.
- App should not crash.


intergalacticspacehighway (approved):
LGTM. Good catch. I should've been more careful!

axeldelafosse (approved):
LGTM

PR #968, opened 2022-04-12
feature: nft activity modal (SHOW2-568)

enesozturk:
# Why
We are planning to implement NFT Activities modal as designed [here](https://www.figma.com/file/720aUothRCUZMtMEVHfJV0/Showtime?node-id=5811%3A87683).

# How

- Created NFT activities components (table, table head, rows, user item etc.)
- Created new navigation for this path 
- Added a new option to dropdown
- Create a hook to handle and store business logic. (Also created additional hook for modal which we use same logic in many place, it can be used for them also)

# Test Plan

-Not yet-

When looking at any NFT, press the settings icon (three-dot), in the dropdown menu, select the `Activities` option. The modal sheet should be opened.

## WIP

This PR is a work-in-progress. There are missing features and views, aware of them and waiting for some responses from both the back-end and design.


axeldelafosse (changes requested):

axeldelafosse (approved):
LGTM ✨

PR #976, opened 2022-04-13
Feature: add icon stories to storybook to see all (show2-680)

enesozturk:
# Why

It's helpful to see all icons as visually because sometimes the same icons in design might have different names and we think it does not exist in the repo but it exists.

# How

Created one icon story file and combined all the icons inside a grid. 
Additionally, Tailwind does not support grid features. It's probably because the grid doesn't supported on React Native.

# Test Plan

- Open Storybook and see `Components/Icon` story.


axeldelafosse (approved):
LGTM ✨

PR #983, opened 2022-04-14
feat: add bottom gap for modal and sheet

enesozturk:
# Why

When we open the NFT details modal and if the details are long, the bottom of the details goes under the notch.

# How

Gave extra gap with the safe area values to the both modal and modal sheet.

# Test Plan

Opening details with long content in both phones have notch and not.


axeldelafosse (approved):
LGTM ✨ Thanks for adding this!

PR #1005, opened 2022-04-19
feat(web): Link support href attrs.

alantoa:
# Why

Link component does not yet support `_blank` mode.
# How

- View the source code of [solito](https://github.com/nandorojo/solito/blob/ad5e470bf3da7aaa9e2a6f768d3bd6bf6116ef9d/example-monorepos/blank/packages/app/rnw-overrides.tsx#L17) and add according to nandorojo's example. 

# Test Plan

Web is work well!


axeldelafosse (approved):
LGTM ✨

PR #1020, opened 2022-04-21
refactor: refactor web navbar buttons (show2 724)

enesozturk:
# Why

Header buttons was not matching with the buttons on the design files. Needed to make some refactors.

# How

# Test Plan


axeldelafosse (changes requested):
The design is not correct on mobile web:

Screenshot 2022-04-21 at 15 38 12

It shouldn't use the background color

axeldelafosse (approved):
LGTM ✨

PR #1026, opened 2022-04-22
Various web improvements

axeldelafosse:
# Why

We need to prepare the web app for production

# How

- Added redirect, head meta, etc...
- Cleaned code and improved auth

# Test Plan

Run the web app


intergalacticspacehighway (approved):
LGTM ❇️

packages/app/lib/refresh-token.web.ts
```diff
@@ -1,11 +1,38 @@
+import { useEffect, useState } from "react";
+
+const REFRESH_TOKEN_STRING = "access-token";
```
edenjw, line 3: @axeldelafosse Is this intended?

PR #1042, opened 2022-04-25
fix: other users show own NFT.

alantoa:
# Why

Own NFT is also added to others' lists when minted.

# How

Check if `profile_id` is equal to `user id`. 

# Test Plan

No yet.


axeldelafosse (approved):
LGTM ✨

intergalacticspacehighway (approved):
LGTM ❇️

PR #1054, opened 2022-04-27
fix(web):  tabbar & layout.

alantoa:
# Why
Our web and Figma's current UI are too inconsistent, so I adjusted. 

This PR made a major optimization for the responsiveness of our website! 

Of course, there are still some minor problems, which I will fix later. (eg: shadow/select/color/more...)

**I‘m can't wait to see these changes in production now! 😆**  

# How

- Limit web maximum width, we need to limit the width of newly added pages in the future. eg: `tw="max-w-screen-xl w-full".

 - `RecyclerListView` on the desktop side, we use `layoutSize` to solve layout problems, which might be a little more troublesome, but it can solve our current problem.

# Test Plan
## Web
I have tested some commonly used pages,  but some pages I'm not sure if I need to modify some. ( If find a page layout problem, we will change it as soon as possible.)

https://user-images.githubusercontent.com/37520667/165513507-c0853977-e344-4d61-9691-3af4ab6ee5a2.mov

## iOS
I checked all the screens, no change

## Android
Not yet.

### DIFF
--- a/apps/next-react-native/src/pages/_app.tsx
+++ b/apps/next-react-native/src/pages/_app.tsx
@@ -1,6 +1,6 @@
 import "raf/polyfill";
 
-import { useState, useEffect } from "react";
+import { useEffect, useState } from "react";
 import { useColorScheme as useDeviceColorScheme } from "react-native";
 
 import { BottomSheetModalProvider } from "@gorhom/bottom-sheet";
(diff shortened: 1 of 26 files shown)


axeldelafosse (approved):
LGTM ✨ Nice work!

enesozturk (approved):

PR #1058, opened 2022-04-28
Feature/show2 751 profile route follow bugs

alantoa:
# Why

- open the “Followers / Following” modal, and press any user change router, but the modal doesn't close.  
- my own profile can't see “Followers / Following”
# How

- move `Follow` component.
- via `useEffect`, close Modal when `address` changed.

# Test Plan

- the web has been tested.
- iOS/Android has been seen,  not affected.

Video: 

https://user-images.githubusercontent.com/37520667/165758614-bd113f2a-2299-4aa0-8e9c-e3b9526c0a05.mov

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -5,7 +5,8 @@
   "eslint.format.enable": true,
   "editor.codeActionsOnSave": {
     "source.fixAll.eslint": true,
-    "source.organizeImports": true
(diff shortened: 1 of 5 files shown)


axeldelafosse (approved):
LGTM ✨

But I think you should disable import sorting in your VS Code settings

PR #1070, opened 2022-05-02
fix: react types error message

alantoa:
# Why

I can't bear to have a `type` error for every file, I searched for the reason because the type version is inconsistent.

截屏2022-05-02 13 59 28

# How

- move `@types/react` from `devDependencies ` to `resolutions`
# Test Plan

After: 

截屏2022-05-02 14 00 18


axeldelafosse (approved):
Nice catch! I think we can upgrade to [`18.0.8`](https://www.npmjs.com/package/@types/react/v/18.0.8) too

PR #1135, opened 2022-05-19
replace references to showtime.io with showtime.xyz

0xkarmacoma:
# Why

Sharing NFT links on showtime.xyz still generates showtime.io links

# How

$ git grep \\.io

# Test Plan

not tested


enesozturk (approved):
LGTM ✅
Just not sure if the URL changes are also made on the backend side to be sure this won't break something? @fontanierh

PR #1164, opened 2022-06-10
refactor: separate providers and clean up root files

enesozturk:
# Why

As much as we add new providers and third-party tools, app root files are growing and this creates a mess, causing hard to refactor issues. It would be better to separate some utilities, providers, and configurations into their own files and use them.

# How

Separated providers, and created platform-specific ones for some (SWR provider). 

⚠️ Just needed to keep theme configuration on the root file because of the `twrnc` structure, [they need to be on the root](https://github.com/jaredh159/tailwind-react-native-classnames/issues/97#issuecomment-982711080). Otherwise style changes not working properly. (This caused me many hours of debugging...) 

# Test Plan

Running apps on both web and mobile for theme changes is working as expected.


intergalacticspacehighway (approved):
LGTM ❇️  looking neat 💯

alantoa (approved):
Great PR!
LGTM! ✨

PR #1165, opened 2022-06-10
feature: improve webs dropdown

enesozturk:
# Why

Our dropdown should look awesome ✨

# How

⚠️ I configured stylings but for some reasons (high possibly because of the `twrnc`, it's not applying `hover:`, `dark:` states to the elements when I use these prefixes. Instead, I needed to write them to the `styles.css` file for the web. That is why no need to `tw="..."` prop usage on the components. Removed them. 

⚠️  Also realized that even if I use `@apply text-black dark:text white` on the `.css` file, it's not applying dark mode too. There are several issues about it but couldn't helped, upgraded Tailwind, but didn't solve too. (Had this on my side projects before as well) I just used pure CSS to detect dark mode (`data-color-scheme="dark"]`).

❓ While developing our universal components, I import from `design-system/` to see changes on my local, after finishing my work, converted them to `@universal-ui/` again. When should we release new components exactly? While (or just before) promoting the new web build?
 

# Test Plan

Checking dropdowns on the web. Styling doesn't apply to the native ones because it uses native dropdowns on the backend but I still checked to see if I broke something.


apps/next-react-native/src/styles/styles.css
```diff
@@ -12,11 +15,39 @@
   button:focus {
     @apply outline-none;
   }
-}
 
-@tailwind components;
+  [role="menu"] > div:nth-child(1) {
+    @apply w-60 rounded-2xl bg-white bg-opacity-60 p-2 shadow backdrop-blur-md backdrop-saturate-150;
```
alantoa: LFG! ✨

alantoa (approved):
LGTM! ✨
Thanks for your work!

intergalacticspacehighway:
> Also realized that even if I use @apply text-black dark:text white on the .css file, it's not applying dark mode too.

twrnc currently uses JS API for dark/light mode and doesn't support hover. It's not using pure tailwind css on web. 

> When should we release new components exactly? 

I think we should do 2 PRs when design-system change is required. 
e.g. 
- First PR includes design-system changes
- Publish after approval
- In follow up PR, we can have actual changes with new versions. 

wdyt?

btw, I am facing an [issue during release](https://showtime-rq88331.slack.com/archives/C02Q0E4PBD0/p1654869375775209?thread_ts=1654169414.171479&cid=C02Q0E4PBD0). Have you tried releasing it on npm?

intergalacticspacehighway (approved):
LGTM 🚀

packages/design-system/dropdown-menu/index.tsx
```diff
@@ -182,9 +185,10 @@ const DropdownMenuItemIndicator = DropdownMenu.menuify(
 );
 
 const DropdownMenuSeparator = DropdownMenu.menuify(
-  (props: { tw?: TW } & ComponentProps<typeof DropdownMenu.Separator>) => (
-    <DropdownMenu.Separator {...props} style={tailwind.style(props.tw)} />
-  ),
+  (props: { tw?: TW } & ComponentProps<typeof DropdownMenu.Separator>) =>
+    IS_WEB ? null : (
+      <DropdownMenu.Separator {...props} style={tailwind.style(props.tw)} />
+    ),
```
axeldelafosse: We should remove the `Separator`s from our implementation instead of doing this. We should be able to display a `Separator` on web

axeldelafosse (changes requested):

packages/app/components/activity/activity-dropdown.tsx
```diff
@@ -17,6 +20,9 @@ import { useMyInfo } from "app/hooks/api-hooks";
 import { useReport } from "app/hooks/use-report";
 import { useUser } from "app/hooks/use-user";
 
+const isWeb = Platform.OS === "web";
+const DropdownMenuSeparator = isWeb ? Fragment : ShowtimeDropdownMenuSeparator;
```
axeldelafosse: Why not simply remove the `<DropdownMenuSeparator />`? IIRC we don't need them on native

axeldelafosse (changes requested):

packages/app/components/activity/activity-dropdown.tsx
enesozturk: Oh don't we? I didn't know that, sure that would be definitely better. I will handle it now 👍🏽

axeldelafosse (approved):
LFG ✨

PR #1168, opened 2022-06-13
fix: design session fixes

intergalacticspacehighway:
# Why
- fixes [bugs](https://linear.app/showtime/issue/SHOW2-880/design-session-dropclaim-feedbackbugs) from last drop design session. 

# Test Plan
- Test drop/claim


packages/app/screens/login.tsx
```diff
@@ -42,4 +42,5 @@ export const LoginScreen = withModalScreen(LoginModal, {
   matchingPathname: "/login",
   matchingQueryParam: "loginModal",
   snapPoints: ["90%"],
+  disableBackdropPress: true,
```
axeldelafosse, line 45: Not sure about the login modal

axeldelafosse (approved):
LFG ✨

packages/app/screens/login.tsx
intergalacticspacehighway: haha yeah I wasn't sure either 😂 . Should I remove it?

PR #1169, opened 2022-06-13
Simplify storybook config

dannyhw:
# Why

Hey I was checking out the repo after seeing the talk at appjs conf and I came accross some storybook config that I thought could be simplified.

Congrats on launching the project, you've made something really cool here!

A lot of packages are included by default in addon-react-native-web so you don't need them listed explicitly. I've also changed the addon to use a partial match for package names, so unless there are showtime packages you don't want to include then just `@showtime` is enough

# How

I updated the main.js to remove packages from the module to transpile list if they are already included by default

# Test Plan

I ran the storybook after making changes and it seems to run fine. If you have chromatic setup (which I think you do) then the visual tests will be a good way to validate this.


intergalacticspacehighway:
Thank you @dannyhw ❤️

PR #1170, opened 2022-06-13
feat: handling profile 404/error/loading & bug fix

alantoa:
# Why

### profile page has issues: 

- we are not handling not found user cases.
- loading/error has some bug.

### Setting page: 
We don't have a clear cache feature now: 
1. sometimes when the amount of data in the simulator is too much, if the cache is not clear, it will be stuck.
2. sometimes you don't know if it's a cache issue.

# How

- use the correct `loading` state for `useSWR`.
- remove the hack code, and move the `loading`/`block`/`error` state component to `TabScene`, it should not appear in `rowRenderer`.
- add the social-link feature.
- add the clear-cache feature(only support native), at present, it will only be displayed in `dev`, and we will open it to users when we are sure that we need to.

# Test Plan

- Check the profile page if it works when 404/error/loading.
- Check the profile social link if it works.


packages/app/components/profile/index.tsx
```diff
@@ -21,9 +20,11 @@ import { useHeaderHeight } from "app/lib/react-navigation/elements";
 
 import { HeaderTabView } from "design-system/tab-view/index";
 import { Route } from "design-system/tab-view/src/types";
+import { TabSpinner } from "design-system/tab-view/tab-spinner";
 
 import { ErrorBoundary } from "../error-boundary";
 import { FilterContext } from "./fillter-context";
+import { Profil404 } from "./profile-404";
```
axeldelafosse: `Profile404`

axeldelafosse (approved):
LFG ✨

packages/app/components/profile/index.tsx
```diff
@@ -153,17 +150,30 @@ const Profile = ({ address }: { address: string | null }) => {
...
+    profileData?.data,
+    isLoading,
+    isError,
+  ]);
 
   const routes = useMemo(
     () =>
       data?.tabs?.map((item, index) => ({
-        title: item?.name,
+        title: item?.name?.replace(/^\S/, (s) => s.toUpperCase()),
```
intergalacticspacehighway, line 176: we can use textTransform style with capitalize

alantoa: oh, I forgot it, thank you!

intergalacticspacehighway (approved):
LGTM 🔥

packages/app/components/profile/index.tsx
alantoa: oh, I remembered, at the time I was considering the situation is that sometimes you need to capitalize the first sentence, just like `This today` instead of `This Today`, via CSS it's not good, so I use `JS` to do this!


figma design snapshot: 

image

intergalacticspacehighway: makes sense 👍

PR #1171, opened 2022-06-14
twitter intent changes and continue transaction post switching

intergalacticspacehighway:
# Why
- Implements twitter intent changes
- Continue drop/claim transaction after switching network

# Test Plan
- Try drop/claim with wrong network.
- Switch popup should appear
- Transaction should continue after switching network instead of showing error

### DIFF
--- a/packages/app/hooks/use-drop-nft.ts
+++ b/packages/app/hooks/use-drop-nft.ts
@@ -149,7 +149,8 @@ export const useDropNFT = () => {
         await pollTransaction(relayerResponse.relayed_transaction_id);
       } else {
         Alert.alert(
-          "Wallet seems to be disconnected. Try signing out and signing in."
+          "Wallet disconnected",
+          "Please logout and login again to complete the transaction"
         );
       }
     } catch (e: any) {
--- a/packages/app/providers/wallet-provider.web.tsx
+++ b/packages/app/providers/wallet-provider.web.tsx
@@ -1,17 +1,13 @@
 import { getDefaultWallets, RainbowKitProvider } from "@rainbow-me/rainbowkit";
 import "@rainbow-me/rainbowkit/styles.css";
-import { chain, configureChains, createClient, WagmiConfig } from "wagmi";
+import { configureChains, createClient, allChains, WagmiConfig } from "wagmi";
 import { alchemyProvider } from "wagmi/providers/alchemy";
 import { publicProvider } from "wagmi/providers/public";
 
-const { chains, provider } = configureChains(
-  [chain.polygon],
-  [
-    alchemyProvider({ alchemyId: process.env.NEXT_PUBLIC_ALCHEMY_ID }),
-    publicProvider(),
-  ]
-);
-
+const { chains, provider } = configureChains(allChains, [
+  alchemyProvider({ alchemyId: process.env.NEXT_PUBLIC_ALCHEMY_ID }),
+  publicProvider(),
+]);
 const { connectors } = getDefaultWallets({
   appName: "Showtime",
   chains,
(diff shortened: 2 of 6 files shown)


packages/app/providers/wallet-provider.web.tsx
```diff
@@ -1,17 +1,13 @@
...
 
-const { chains, provider } = configureChains(
-  [chain.polygon],
-  [
-    alchemyProvider({ alchemyId: process.env.NEXT_PUBLIC_ALCHEMY_ID }),
-    publicProvider(),
-  ]
-);
-
+const { chains, provider } = configureChains(allChains, [
```
enesozturk, line 7: Is it (`allChains`) because the transactions are network agnostic?

intergalacticspacehighway: No, added it to handle a case when wallet doesn't have polygon mumbai chain configured. It was [throwing error](https://github.com/tmm/wagmi/blob/main/packages/core/src/connectors/injected.ts#L203) on switching.

intergalacticspacehighway: we can also just add polygon and polygonMumbai. I don't think this would matter much. wdyt? 🤷‍♂️

axeldelafosse (approved):
LFG ✨

packages/app/providers/wallet-provider.web.tsx
enesozturk: Yeah I remember I added polygonMumbai too but cannot see now, maybe missed when rebasing :/ Because when they see the switch network modal, all the networks will show up, might this confuse the users? Not sure

packages/app/hooks/use-drop-nft.ts
```diff
@@ -149,7 +149,8 @@ export const useDropNFT = () => {
         await pollTransaction(relayerResponse.relayed_transaction_id);
       } else {
         Alert.alert(
-          "Wallet seems to be disconnected. Try signing out and signing in."
+          "Wallet disconnected",
+          "Please logout and login again to complete the transaction"
```
enesozturk, line 153: I think somehow we should handle and solve this later because it doesn't look like a good UX to say logout and log in :/

enesozturk (approved):
LGTM ❇️

packages/app/hooks/use-drop-nft.ts
enesozturk: Also will you handle Twitter colored button like Alex mentioned? @intergalacticspacehighway 
https://showtime-rq88331.slack.com/archives/C02PXGK3V8D/p1655138669237659

packages/app/providers/wallet-provider.web.tsx
intergalacticspacehighway: > switch network modal, all the networks will show up.

It only shows the network that we're switching to

packages/app/hooks/use-drop-nft.ts
intergalacticspacehighway: > I think somehow we should handle and solve this later because it doesn't look like a good UX to say logout and log in :/


yeah working on a hook that'll handle it


> Also will you handle Twitter colored button like Alex mentioned? @intergalacticspacehighway


Yes, added it here https://github.com/showtime-xyz/showtime-frontend/pull/1171/files#diff-be9253eac967b37574365765760354328be2b22357666e4aa7d6b55c3f46add2R105

enesozturk: Oh sorry missed 👍🏽

packages/app/providers/wallet-provider.web.tsx
intergalacticspacehighway: we won't show switch network modal but trigger the switching programmatically using switchNetwork from wagmi hook. So I think it won't be an issue

PR #1174, opened 2022-06-14
Use `tailwindcss-react-native`

axeldelafosse:
# Why

We want to use `tailwindcss-react-native` to use CSS on web. We will be able to do server-side rendering for the user profiles for example

# How

- Install and use `tailwindcss-react-native`
- TODO: replace `twrnc`

# Test Plan

Run the apps


packages/design-system/tailwind/colors.ts
```diff
@@ -1,4 +1,4 @@
-export const colors = {
```
enesozturk, line 1: Isn't it the same or it should be `module.exports`, maybe because of compiling issues? Just curious 🤔

axeldelafosse: Yeah the content is the same but we would like to import this file from the Tailwind config and we can't import an ES module from it

PR #1175, opened 2022-06-15
fix(comment): disable btn when content is empty

alantoa:
# Why 

- publish content even if the comment input is empty.
- comment input avatar style is broken.

image

# How

- disable btn when content is empty.
- adjust the avatar position.

# Test Plan

check if the comment feature is working.


intergalacticspacehighway (approved):
LGTM ❇️

axeldelafosse (approved):
LFG ✨

PR #1176, opened 2022-06-15
chore: upgrade wallet packages and clean command

enesozturk:
# Why

With the newest version of the RainbowKit and Wagmi, some deep linking issues and improvements had made.  Also updated clean command because every time we need to delete `.next` build folder too. And a small styling issue on minting successful modal.

# How

# Test Plan


intergalacticspacehighway (approved):
LGTM ❇️

axeldelafosse (approved):
LFG ✨

PR #1178, opened 2022-06-15
fix: ToolTip onPress

nandorojo:
This memoization is unnecessary. Also, it's better to pass a dispatch function to a state updater when using the existing state.

Does `onTouchStart` still work on Views btw?

# Why

Saw some unnecessary memoization.

# How

🤷‍♂️

# Test Plan


alantoa:
Thank you! @nandorojo 🙏

onTouchStart is work on the previous RN version, now I didn't check it for the latest version, because we don't use it on native at the moment.

I'll check again tomorrow!

it was mainly for the web at the time, and I also supported native by the way.

PR #1185, opened 2022-06-16
fix: remove video previewOnly from web

intergalacticspacehighway:
# Why
- we disabled videos on grid cards but we can enable it on web. It was also causing to videos not playing in detail screen

# How
- set previewOnly to false in video context for web

# Test Plan
- run app and go to profile where there are video NFTs


alantoa (approved):
LGTM! ❤️

PR #1189, opened 2022-06-20
fix: limit max size to 50 mb during drops

intergalacticspacehighway:
# Why
to show an alert if user tries to upload a media > 50mb

# Test Plan
- try uploading a media > 50mb during drops


alantoa (approved):
LGTM! ✨

PR #1190, opened 2022-06-20
fix: display error message from API

intergalacticspacehighway:
# Why
When there's an error, user is shown generic axios message like request failed with 500 code.

# How
Get the message from API and throw a new error with that message.

# Test Plan
Tested on drop, tried upload >50mb and verified if appropriate error message is shown.


axeldelafosse (approved):
LFG ✨

PR #1191, opened 2022-06-20
fix: drop feedback

intergalacticspacehighway:
# Why
includes some feedback regarding drop https://showtime-rq88331.slack.com/archives/C02PXGK3V8D/p1655418694706589

# Test Plan
- Open drop detail screen. 
- Requires button component release, will do in next PR.


axeldelafosse (approved):
LFG ✨

PR #1193, opened 2022-06-21
chore: upgrade rainbowkit

enesozturk:
# Why

Rainbow has some improvements for deep linking issues with the newest releases. I'll continue testing the flows with the newest version.

# How

# Test Plan


axeldelafosse (approved):
LFG ✨

PR #1194, opened 2022-06-21
Hide price and Buy button

axeldelafosse:
# Why

The App Store requires to enable "in-app purchases" if we let people buy NFTs so let's hide the price and buy button for now


intergalacticspacehighway (approved):
LGTM ❇️

PR #1196, opened 2022-06-21
fix: skip redirection on native

intergalacticspacehighway:
# Why
- skip for now button redirects to web app

# How
we don't have deeplink support on native yet, so it should just close the modal instead  of redirecting.

# Test Plan
- create a drop, press skip for now on success screen


axeldelafosse (approved):
LFG ✨

PR #1197, opened 2022-06-21
revert: remove edition size frontend hack (do not merge this yet)

intergalacticspacehighway:
# Why
- this seems to be an indexer issue. Claimed count won't be always less by 1 than actual count. So the hack may not work for all cases. https://showtime-rq88331.slack.com/archives/C02PXGK3V8D/p1655598050596599

# How
- remove the less by 1 hack


alantoa (approved):
LGTM! ✨

PR #1198, opened 2022-06-21
fix: button text color in drop screen

intergalacticspacehighway:
# Why
fixes button text not visible in sold out

# How
upgrade button and it's dependent version

# Test Plan
open a drop that has `sold out` text in it


axeldelafosse (approved):
LFG ✨

PR #1200, opened 2022-06-22
refactor: blur views on mobile safari

enesozturk:
# Why

Blur backgrounds are not working as expected on mobile safari browsers because we are not targeting `webkit` browsers, I tried even using Tailwind but its not adding these prefixes, needed to update the blur background hook.

# How

# Test Plan

Checking header, feed item details, bottom nav bar background on mobile web Safari

### DIFF
--- a/.bitmap
+++ b/.bitmap
@@ -11,7 +11,7 @@
 {
     "accordion": {
         "scope": "showtime.universal",
-        "version": "0.0.16",
+        "version": "0.0.17",
         "mainFile": "index.tsx",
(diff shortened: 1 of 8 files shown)


axeldelafosse (approved):
LFG ✨

Can you publish a new version of `@showtime-xyz/universal.hooks` and use it instead of `design-system/hooks` please?

enesozturk:
Yeap, of course, wanted to get approve first 👍🏽

intergalacticspacehighway (approved):
LGTM ❇️

PR #1202, opened 2022-06-22
Fix notifications in header

axeldelafosse:
# Why

We have a weird bug happening only in production build where Next.js `Link` is navigating to a page. Can't understand why so decided to just use a `Pressable` instead for this use-case

# How

- Used `Pressable` instead of `Link` for `TabBarIcon` if `onPress` callback

# Test Plan

Run the web app in Vercel PR preview and check if clicking on the Notifications icon in the header is working correctly


intergalacticspacehighway (approved):
LGTM ❇️

PR #1203, opened 2022-06-22
fix: tab list & spinner component

alantoa:
# Why
- The Profile page switching tab the list height does not change.
- The spinner component doesn't work on the web.
- Profile header shadow is weird.

# How
- set tab `display: none` when the tab is not focused.
- refactor Spinner component for weh.
- hide shadow when before loading.

# Test Plan

- check if the Profile/Trending page works.
- check new Spinner component works.


intergalacticspacehighway (approved):
LGTM ❇️

PR #1204, opened 2022-06-22
add meta tags for drops

intergalacticspacehighway:
# Why
add meta tags for drops

# How
use getServerSide props to generate metatags

# Test Plan
Try sharing drop link, it should display og and twitter card


enesozturk (approved):
LGTM ❇️

PR #1205, opened 2022-06-22
fix: email login issues on web

enesozturk:
# Why

On the web, we were not able to do transactions dues to missing web3 setter.

# How

# Test Plan

Testing minting or drop flows on the web with email login.


enesozturk:
This is urgent and would be better to merge now. Let's discuss tomorrow if you find any issue on the PR @axeldelafosse @intergalacticspacehighway

PR #1206, opened 2022-06-22
refactor: nft card goes under bottom navigation

enesozturk:
# Why

In the NFT details screen on mobile web, the NFT details card was placed under the bottom navigator. Needed to give bottom value.

# How

# Test Plan

Checking NFT details page on the mobile web.

### DIFF
--- a/packages/app/components/swipe-list.tsx
+++ b/packages/app/components/swipe-list.tsx
@@ -430,7 +430,7 @@ export const FeedItem = memo(
 
           <Reanimated.View
             style={[
-              tw.style(`z-1 absolute bottom-0 right-0 left-0`),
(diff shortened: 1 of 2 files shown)


packages/app/components/swipe-list.tsx
```diff
@@ -428,8 +430,9 @@ export const FeedItem = memo(
 
           <Reanimated.View
             style={[
-              tw.style("z-1 absolute bottom-0 right-0 left-0"),
```
enesozturk, line 431: Here, `bottom-[${}]` was getting an absurd number event the parameter is not its. Needed to pass it as inline style.

intergalacticspacehighway (approved):
LGTM ❇️

PR #1209, opened 2022-06-23
fix: handle wallet disconnect during drop/claim

intergalacticspacehighway:
# Why
we don't handle a case when a wallet is disconnected during drop/claim

# How
During drop/claim
- if user is not logged in, we show a sign in button
- if user's wallet is not connected, we show a connect wallet button
https://showtime-rq88331.slack.com/archives/C02PXGK3V8D/p1655946595060959

This is a workaround due to the lack of imperative connect API in rainbowkit. I'll try to implement it and raise a PR in rainbowkit. https://github.com/rainbow-me/rainbowkit/discussions/386

# Test Plan
- try drop/claim and disconnect wallet (I did it by setting autoConnect false)

# Aside
we need to refactor some web3 hooks  https://linear.app/showtime/issue/SHOW2-917/clean-up-web3-action-hooks

### DIFF
--- a/packages/app/components/claim/claim-form.tsx
+++ b/packages/app/components/claim/claim-form.tsx
@@ -8,9 +8,11 @@ import { tw } from "@showtime-xyz/universal.tailwind";
 import { Text } from "@showtime-xyz/universal.text";
 import { View } from "@showtime-xyz/universal.view";
 
+import { ConnectButton } from "app/components/connect-button";
 import { Media } from "app/components/media";
 import { PolygonScanButton } from "app/components/polygon-scan-button";
 import { useMyInfo } from "app/hooks/api-hooks";
+import { useWallet } from "app/hooks/auth/use-wallet";
 import { useClaimNFT } from "app/hooks/use-claim-nft";
 import {
   CreatorEditionResponse,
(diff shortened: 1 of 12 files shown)


packages/app/components/claim/claim-form.tsx
```diff
@@ -71,6 +78,25 @@ export const ClaimForm = ({ edition }: { edition: CreatorEditionResponse }) => {
...
+  if (!isAuthenticated) {
+    return (
+      <View tw="p-4">
+        <Button onPress={navigateToLogin}>Please login to continue</Button>
+      </View>
+    );
+  }
+
+  // TODO: remove this after imperative login modal API in rainbowkit
+  if (!connected) {
```
enesozturk, line 90: I guess we are gonna need to update connected conditions because if the user login with email and refresh the page, they will be in this status: authenticated and not connected (but should be connected). I'll do some checks and push to this branch, wdty? @intergalacticspacehighway

intergalacticspacehighway: yes, good catch. actually, I didn't handle that because it was a bit unsure whether we're supporting magic wallet for claiming. Feel free to add it!

axeldelafosse (approved):
LFG ✨

PR #1210, opened 2022-06-23
chore: upgare wallet depencencies on web

enesozturk:
# Why

There is an issue that we are not redirected to the browser app when we try to log in with Coinbase Wallet. It was opening our app in in-app browser then does nothing. 
Rainbow team shows that if we upgrade it, that will be solved:
- https://github.com/rainbow-me/rainbowkit/issues/531

# How

# Test Plan


intergalacticspacehighway (approved):
LGTM ❇️

PR #1211, opened 2022-06-23
fix: profile metatags

intergalacticspacehighway:
# Why
add og and twitter metatags in user profile

# Test Plan
try opening user profile and verify if metatags are coming from server


apps/next-react-native/src/pages/nft/[chainName]/[contractAddress]/[tokenId]/index.tsx
```diff
@@ -7,25 +7,27 @@ import { getMediaUrl } from "app/utilities";
...
+            description: nft.token_description,
+            image: imageUrl,
+          },
         },
-      },
-    };
-  }
+      };
+    }
+  } catch (e) {}
```
Comment (intergalacticspacehighway) [inline apps/next-react-native/src/pages/nft/[chainName]/[contractAddress]/[tokenId]/index.tsx:28]: need to improve error handling here, didn't want to spend too much time as we're using it just for metatags for now. will revisit, we can hydrate this data in swr.

enesozturk (approved):
LGTM ❇️

PR #1212, opened 2022-06-23
fix: profile blocked text

alantoa:
# Why

- profile blocked text is obscured.
- on native, `ProfileListFilter` is inserted in the list now, the code is not pure.

# How

- use`TabScrollView` to display the blocked text.
- native and web use the same props as we build new a `TabView`, we don't need hacks.

# Test Plan

- check the profile page if it works.


axeldelafosse (approved):
LFG ✨

PR #1213, opened 2022-06-23
fix: return checksum address

intergalacticspacehighway:
# Why
we need checksumed address of the user address but due to some reason wallet connect is returning it in lower case on native.
```
[Error: Request failed with status code 500]
 [Error: ('Web3.py only accepts checksum addresses. The software that gave you this non-checksum address should be considered unsafe, please file it as a bug on their platform. Try using an ENS name instead. Or, if you must accept lower safety, use Web3.toChecksumAddress(lower_case_address).', '0x000')]
```

# How
use getAddress from ethers util that converts address to checksumed address

# Test Plan
- test drop/claim works on native


axeldelafosse (approved):
LFG ✨

PR #1214, opened 2022-06-23
feat: add web3modal to connect other wallets to profile

enesozturk:
# Why

In profile settings, there is add wallet section to merge accounts. Currently, due to Rainbow Kit's structure, we cannot get the wallet connection data like the web3modal. I added back the web3modal for just this feature otherwise users won't be able to do it on the web. 

# How

Added web3modal and separated `useAddWallet` hook for the web.

# Test Plan

Adding new wallets to the profile on the web.


intergalacticspacehighway (approved):
LGTM ❇️

PR #1216, opened 2022-06-24
fix: imporve NFT detail & profile filter

alantoa:
# Why
- NFT detail page is not yet adapted to the large screen.
- Profile NFT filters missing on the large screen.
# How

- adjust UI.

# Test Plan

just fix the style, no need to test.


alantoa:
Just fixing urgent bugs, and worked on the preview.

PR #1221, opened 2022-06-27
fix: metamask issue on add wallet 

enesozturk:
# Why

There is an issue with MetaMask when trying to add a wallet in settings. 

# How

Needed to handle MetaMask response another way.

# Test Plan

Adding new wallet to profile in settings with MetaMask


intergalacticspacehighway (approved):
LGTM ❇️

PR #1222, opened 2022-06-27
feat/fix keyboard aware scrollview

catalinmiron:
# Why

Profile/Drop NFT or any screen that lives outside of BottomSheet and have multiple inputs inside a list should always keep the focused inputs visible. 

# How

Patch `scroll-view` component with `react-native-keyboard-aware-scrollview`

# Test Plan

-


packages/app/components/drop/drop-form.tsx
```diff
@@ -31,6 +30,7 @@ import { useRouter } from "app/navigation/use-router";
 import { getTwitterIntent, getUserDisplayNameFromProfile } from "app/utilities";
 
 import { useFilePicker } from "design-system/file-picker";
+import { ScrollView } from "design-system/scroll-view";
```
axeldelafosse: We can switch back to `@showtime-xyz/universal.scroll-view` once we release the new version of `universal.scroll-view` via Bit

axeldelafosse (approved):
LFG ✨

intergalacticspacehighway (approved):
LGTM 🥳

packages/app/components/drop/drop-form.tsx
catalinmiron: Done

intergalacticspacehighway (approved):
LGTM ❇️

packages/app/components/drop/drop-form.tsx
```diff
@@ -3,14 +3,14 @@ import { Linking, Platform } from "react-native";
...
 import { Button } from "@showtime-xyz/universal.button";
 import { Checkbox } from "@showtime-xyz/universal.checkbox";
-import { ErrorText } from "@showtime-xyz/universal.fieldset";
-import { Fieldset } from "@showtime-xyz/universal.fieldset";
-import { Image as ImageIcon, FlipIcon } from "@showtime-xyz/universal.icon";
+import { ErrorText, Fieldset } from "@showtime-xyz/universal.fieldset";
+import { FlipIcon, Image as ImageIcon } from "@showtime-xyz/universal.icon";
 import { Pressable } from "@showtime-xyz/universal.pressable";
+import { useSafeAreaInsets } from "@showtime-xyz/universal.safe-area";
 import { ScrollView } from "@showtime-xyz/universal.scroll-view";
```
alantoa, line 14: just a heads up, I noticed that this component was updated, but didn't see its component version update, so maybe the old version was used.

alantoa (approved):
LGTM! ✨

packages/app/components/drop/drop-form.tsx
catalinmiron: Oh, the latest version is `0.0.15` but because it's a patch, when you'll install the packages it will take the latest change. Or am I missing something? Shall I bump the package version (package.json) as well? I see the latest version inside package.json is `0.0.6`

PR #1223, opened 2022-06-27
Remove `<Collection />` component

axeldelafosse:
# Why

This component is not relevant anymore and it is even misleading for the drops

# How

- Removed the component
- Made sure the feed and modals looked good

# Test Plan

Run the apps and check if the feed is looking good


intergalacticspacehighway (approved):
LGTM ❇️

PR #1224, opened 2022-06-27
fix: video and placehoder image overlapping

alantoa:
# Why

- video and image overlapping.
- some components have type error/warning
- we removed the Collection component, so we need to adjust item height.
- the style of the card is inconsistent with Figma.
- mint modal style has some broken.
# How

- blur the background image and center the video, and add a patch for `expo-av`.
- fix many types of errors.
- adjust card item height.
- improve card style.
- fix mint modal style.
# Test Plan

- check display normally if the video and image overlapping.
- check if the card style is looking good.
- check if the mint modal is looking good.


packages/design-system/spinner/index.web.tsx
```diff
@@ -1,11 +1,10 @@
-import { useEffect, useRef, useCallback, useMemo } from "react";
+import { useEffect, useCallback, useMemo } from "react";
 import { Animated, Easing } from "react-native";
 
 import { getSpinnerSize, SpinnerView, SpinnerProps } from "./spinner-view";
 
 export const Spinner = ({ size, duration = 750, ...rest }: SpinnerProps) => {
-  const transition = useRef(new Animated.Value(0)).current;
-
+  const transition = useMemo(() => new Animated.Value(0), []);
```
catalinmiron: It's better to use `useRef` for Animated.Values. It's safe to revert this change. 
Also, what do you think about using `moti` instead? 

```javascript
<MotiView 
  from={{rotate: '0deg'}}
  animate={{rotate: '360deg'}}
  transition={{
    type: 'timing',
    loop: true,
    repeatReverse: false,
    easing: Easing.linear
  }}
  style={{
    height: getSpinnerSize(size),
    width: getSpinnerSize(size),
  }}
/>
```

packages/app/hooks/use-nft-cards-list-layout-provider.ts
```diff
@@ -14,7 +14,7 @@ const useNFTCardDimensions = () => {
     return { height: width / 3, width: width / 3 };
   }
   return {
-    height: 594 + LIST_MARGIN_TOP_LARGE,
+    height: 548 + LIST_MARGIN_TOP_LARGE,
```
catalinmiron: can we move this to a constant?

packages/app/components/feed/comment-button.tsx
```diff
@@ -50,8 +52,9 @@ export function CommentButton({ nft }: CommentButtonProps) {
       size="regular"
       tw="h-auto p-0"
       onPress={handleOnPress}
+      accentColor={textColors}
     >
-      <Message height={24} width={24} />{" "}
+      <Message height={24} width={24} color={iconColor} />{" "}
       {nft?.comment_count > 0 ? getRoundedCount(nft.comment_count) : ""}
```
catalinmiron, line 58: I am not sure if this is semantically correct, should we wrap the "content" into a `<Text>` element? This should work on web, but React Native likes to keep the "text" wrapped into a <Text> element.

packages/app/components/claim/gift-button.tsx
```diff
@@ -1,26 +1,28 @@
...
-        // @ts-ignore
-        color={tw.style("bg-gray-900 dark:bg-white").backgroundColor}
-      />{" "}
+    <Button
+      variant="text"
+      size="regular"
+      tw="h-auto p-0"
+      accentColor={textColors}
+    >
+      <Gift height={20} width={20} color={iconColor} />{" "}
```
catalinmiron, line 25: Same

catalinmiron (changes requested):

intergalacticspacehighway (approved):
LGTM 💯

packages/app/components/card/rows/elements/ownership.tsx
```diff
@@ -26,7 +26,7 @@ function OwnershipContainer({
   ) : (
     <View
       tw={[
-        "flex h-[30px] w-[30px] flex-row flex-wrap justify-between",
+        "flex h-[32px] w-[32px] flex-row flex-wrap justify-between",
```
intergalacticspacehighway: nit: here i think we can use h-8, w-8

alantoa: yeah agree!

packages/design-system/spinner/index.web.tsx
alantoa: thank you for your review!
it's just a test, I'm not sure if it's causing every update [this](https://www.chromatic.com/build?appId=619f8881c3b4c2003a08edf7&number=2092), `moti` I'll try again, I can't remember what problem I had before without using it.

packages/app/hooks/use-nft-cards-list-layout-provider.ts
alantoa: yeah that's awosome!

packages/app/components/feed/comment-button.tsx
alantoa: yeah, it's works on native/web, I very much agree with your suggestion!  I also like the `${text} ` or `&nbsp;` way of writing, I will modify it.
It's just that it was written like this before, I didn't change it, maybe we define a specification.

packages/app/components/feed/feed.md.tsx
```diff
@@ -30,7 +30,7 @@ import { Hidden } from "design-system/hidden";
 import { Tabs } from "design-system/tabs";
 import { CARD_DARK_SHADOW } from "design-system/theme";
 
-const CARD_HEIGHT = 848;
+const CARD_HEIGHT = 825;
```
enesozturk, line 33: Any reason to change this? We sometimes needed to change these values but they are causing UI bugs. I think we should use normal the web approach instead rendering the lists with FlatList or VirtualizedList etc. I'll open a discussion on this. Just wanted to ping you if this won't break something.

catalinmiron (approved):

catalinmiron:
LGTM

enesozturk (approved):
LGTM ❇️

packages/app/components/feed/feed.md.tsx
alantoa: yeah, we have removed the Collection component, so we need to adjust item height.
I know now fix height it's very hacking, It will be resolved once we get rid of the `recyclerlistview`.

packages/design-system/spinner/index.web.tsx
alantoa: just try it:
image
it's working on the web but does not work on native. :(

alantoa: but thank you tips!
so let's try using `moti` on the web, native code is temporarily unchanged, see [this](https://www.chromatic.com/build?appId=619f8881c3b4c2003a08edf7&number=2092) if it disappears

PR #1225, opened 2022-06-27
nit: drop feedback

intergalacticspacehighway:
# Why
small changes based on feedback in the feedback channel.

# Test Plan
- test drop/claim


axeldelafosse (approved):
LFG ✨

PR #1227, opened 2022-06-27
refactor: improve mint, drop and claim flow for deeplinking issue

enesozturk:
# Why

On mobile web, due to deepling issues of Rainbow and iOS, we needed to do same logic that users need to press sign message when they want to do any transaction (mint, drop, claim).

# How

Refactored `useDropNFT`, `useClaimNFT`, `useMintNFT` hooks and added separated states which only used on the mobile web then handled that state on the views. 

# Test Plan

Testing minting, dropping, claiming on mobile web browsers.


packages/app/components/claim/claim-form.tsx
```diff
@@ -223,17 +227,43 @@ export const ClaimForm = ({ edition }: { edition: CreatorEditionResponse }) => {
...
-            >
-              {state.status === "loading"
-                ? "Claiming..."
-                : state.status === "error"
-                ? "Failed. Retry!"
-                : "Claim for free"}
-            </Button>
+            {shouldShowSignMessage ? (
+              <View tw="px-2">
+                {!isSignRequested ? (
```
intergalacticspacehighway, line 232: i think claim works without manual signature request, so we can prevent this prompt here

intergalacticspacehighway: I guess the claim should work without a signature prompt. During drop/mint, it takes a bit longer to upload the file so it would need it!

enesozturk:
@intergalacticspacehighway Yeah man I tried and it works as expected on Claim flow. Updated, thanks!

intergalacticspacehighway (approved):
LGTM 💯

axeldelafosse (commented):
LFG ✨

PR #1230, opened 2022-06-28
chore: separate rudderstack platform keys

enesozturk:
# Why

Rudderstack keys are different depending on the platform. We might need to add web keys to env files and use those ones on the web source code as well.
@axeldelafosse 

# How

# Test Plan


enesozturk:
@axeldelafosse any comments on this? or should I decline?

PR #1231, opened 2022-06-28
chore: update mobile check method usage

enesozturk:
# Why

Drop and claim success modals are not showing because the mobile web checker methods were wrong.

# How

# Test Plan


enesozturk:
@intergalacticspacehighway sorry that I should have notified you I changed their name.

intergalacticspacehighway (approved):
LGTM ❇️

PR #1232, opened 2022-06-28
fix: notifications error boundary

intergalacticspacehighway:
# Why
notification can crash on some profile. It's using suspense but lacked error boundary.
Fix has been pushed on BE https://showtime-rq88331.slack.com/archives/C02QD3J7DJM/p1656411292367589

# How
add suspense and error boundary to Notifications

# Test Plan
- tapping on notification should not lead to crash


axeldelafosse (approved):
LFG ✨

PR #1236, opened 2022-06-28
fix: hide drop option in profile

intergalacticspacehighway:
# Why
user should be able to hide a drop after creation

# How
Enable `hide` option in NFT dropdown and `hide` all other options for drop nfts 😛  

# Test Plan
Go to profile and tap dropdown in NFT drop card. Hide option should be visible


axeldelafosse (approved):
LFG ✨

PR #1238, opened 2022-06-28
Feature/show2 944 nft card sizing

alantoa:
# Why

- we removed the <Collection /> component, so the Trending NFT Card height is incorrect.
- `Suggested` sticky now is not working on the web.

# How
- minus the height of the `Collection` component.
- use `react-stickynode` lib to fix sticky.

# Test Plan

- check if the height of the Trending Card height is correct.
- check home `Suggested` sticky if it works.


axeldelafosse (approved):
LFG ✨

PR #1240, opened 2022-06-29
Feature/show2 940 remove the collection filter

alantoa:
# Why

- now we only display the `Showtime` collection, so remove the Collection filter.
- Select the component the height has 2 more pixels.

# How
- use list count instead of `Collection` filter.
- subtract 2px's `padding-y`

# Test Plan

check if the profile looks good!


axeldelafosse (approved):
LFG ✨

PR #1241, opened 2022-06-29
fix: update edition cache after claiming and notification error 

intergalacticspacehighway:
# Why
- Claim button is not disabled after successful claiming.
- Notification crash had one more issue apart from API failure. The `View cannot have Text RNW error` was being caught in `ErrorBoundary` leading to show `Something went wrong`. This error was happening due to the usage of `TextLink` as it uses `View` on web and mounting Text in it was causing error.

# How
- Update swr cache
- Notification crash fix - Use Link from solito to pass href and a Text node for styling. 

# Test Plan
- Test notifications and claim on web

### DIFF
--- a/packages/app/components/card/card.tsx
+++ b/packages/app/components/card/card.tsx
@@ -71,8 +71,6 @@ function Card({ listId, nft, numColumns, tw, onPress, hrefProps = {} }: Props) {
     );
   }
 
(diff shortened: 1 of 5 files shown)


axeldelafosse (approved):
LFG ✨

PR #1243, opened 2022-06-29
Better font handling on web

axeldelafosse:
# Why

The fonts are not really well handled

# How

- Use Next.js font optimization
- Use `font-display: fallback` instead of `swap`

# Test Plan

Run the web app


enesozturk (approved):
LGTM ❇️

PR #1244, opened 2022-06-29
refactor: scroll to top issues

enesozturk:
# Why

Due to some default configurations of current router methods, we were scrolling the page to the top on page loads or URL changes. This was causing a bad experience. During this task I was able to disable scrolling for the following cases:

- Opening NFT dropdown modals (activity, details, etc)
- Opening comments modal
- Tabs (profile, feed)

> **Note**
> Still couldn't figure out why it's scrolling top when navigating the with browser's native buttons.
> Also header buttons are scrolling top but disabling these causes bad UX too, like the following case: if I scroll the page, then go to the profile, it starts from the scrolled position. **It shouldn't**, because I'm opening that page for the first time...
> I'll continue to investigate this.

# How

Overrided. the router methods that inherited from Solito router and set the scroll to false. Needed to update some more methods too.

# Test Plan

Testing modals and page navigations on web.

### DIFF
--- a/apps/next-react-native/src/styles/styles.css
+++ b/apps/next-react-native/src/styles/styles.css
@@ -53,10 +53,6 @@
     @apply bg-red-500 bg-opacity-20;
   }
 
-  [data-color-scheme="dark"] [role="menuitem"].danger:focus {
-    @apply outline-red-800;
-  }
-
   [role="menuitem"] > div > div:nth-child(1) {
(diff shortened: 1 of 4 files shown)


axeldelafosse (approved):
LFG ✨

intergalacticspacehighway (approved):
LGTM 💯

alantoa (approved):
LGTM! ✨

PR #1249, opened 2022-06-30
refactor: cannot burn nft on web

enesozturk:
# Why

When we merge the accounts, we see the NFTs which created with the wallet we added. Then when trying to delete these NFTs, we got an error due to wrong ownership. Added an alert to make users log in with the wallet the NFT is created with.

# How

Compared the nft creator address with the current user address and called alert if needed.

# Test Plan

- Login
- Add wallet to account in settings
- Try to remove (burn) the NFT from secondary wallet.

### DIFF
--- a/packages/app/components/delete.tsx
+++ b/packages/app/components/delete.tsx
@@ -7,6 +7,7 @@ import { formatDistanceToNowStrict } from "date-fns";
 import { Controller, useForm } from "react-hook-form";
 import useSWR from "swr";
 
+import { useAlert } from "@showtime-xyz/universal.alert";
 import { Button } from "@showtime-xyz/universal.button";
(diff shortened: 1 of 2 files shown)


packages/app/components/delete.tsx
```diff
@@ -125,16 +129,27 @@ function Delete({ nft }: { nft: NFT }) {
     );
   }
 
+  const handleBurnNFT = () => {
+    if (nft?.creator_address !== userAddress) {
+      alert(
+        "Switch Wallet",
+        "Please log in with the wallet that created this NFT."
+      );
```
axeldelafosse, line 137: 💯

axeldelafosse (approved):
LFG ✨

PR #1254, opened 2022-06-30
Typography

axeldelafosse:
# Why

We saw some issues with our typography 

# How

- Used Space Grotesk capsize metrics
- Updated some Text components to avoid setting manual line height
- Added a manual line height for Title (in Card) on large screens. For some reason I think the Link component is messing up the positioning

# Test Plan

Run the web app on desktop and check if the Card Title is OK


alantoa (approved):
LGTM! ✨

PR #1256, opened 2022-07-01
fix: peerDependencies & update packages

alantoa:
# Why

- universal components `Reanimated` & `GestureHandler` package dependencies have some problems.
- GestureHandler `v2.5.0` & `Reanimated` `v2.9.0` looks good to us.

# How

- move `Reanimated` & `GestureHandler` packages to `peerDependencies`.
- upgrade GestureHandler version, but `Reanimated` `v2.9.0` have some issues for `next.js`, will dig it later.

# Test Plan

check the Native/Web if it works.


axeldelafosse (approved):
LFG ✨

PR #1257, opened 2022-07-01
Force a major version bump

axeldelafosse:
# Why

We want to be able to run `/promote frontend major` to force a major version bump that triggers a new build and submit to the app stores


alantoa (approved):
LGTM! ✨

PR #1258, opened 2022-07-04
fix: Trending swipe to trigger refresh

alantoa:
# Why

- issue from: https://linear.app/showtime/issue/SHOW2-960/trending-swipe-to-trigger-refresh

# How

- lazy load the Trending list and ensure only one list in each tab.

# After

https://user-images.githubusercontent.com/37520667/177147856-6b0360a6-cdca-4e11-ba90-9302850195b6.MP4

# Test Plan

- check if the Trending list works.


axeldelafosse (approved):
LFG ✨

PR #1259, opened 2022-07-05
feat: profile default tab

alantoa:
# Why

- If Created list counts = 0, it should go to the Owned tab directly. 
- If the parmas of the current tab exceeds the range, will get an empty list. 

# How

- do some judgment handing the default tab.
- when the params don't match, enter the default tab.

# Test Plan

check if the Profile tab load works.

### DIFF
--- a/packages/app/components/profile/index.tsx
+++ b/packages/app/components/profile/index.tsx
@@ -65,16 +65,16 @@ const Profile = ({ address }: { address: string | null }) => {
     if ((router.query as any)?.tab) return (router.query as any)?.tab;
     const createdIndex =
       data?.tabs.findIndex((item) => item.type === "created") ?? 0;
-    const onwedIndex =
+    const ownedIndex =
       data?.tabs.findIndex((item) => item.type === "owned") ?? 0;
-    if (onwedIndex === createdIndex || createdIndex === -1 || onwedIndex === -1)
+    if (ownedIndex === createdIndex || createdIndex === -1 || ownedIndex === -1)
       return 0;
     const createdCount = data?.tabs[createdIndex].displayed_count ?? 0;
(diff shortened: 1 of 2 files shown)


packages/app/components/profile/index.tsx
```diff
@@ -48,20 +49,43 @@ const Profile = ({ address }: { address: string | null }) => {
...
   });
+  /**
+   * defalut tab index.
+   * if Created list = 0, it should go to the Owned section directly.
+   * */
+  const defaultIndex = useMemo(() => {
+    if ((router.query as any)?.tab) return (router.query as any)?.tab;
+    const createdIndex =
+      data?.tabs.findIndex((item) => item.type === "created") ?? 0;
+    const onwedIndex =
```
intergalacticspacehighway: nit: typo

intergalacticspacehighway (approved):
LGTM ❇️

packages/app/components/profile/index.tsx
alantoa: thank you! my bad! 🙏

packages/app/components/profile/index.tsx
```diff
@@ -48,20 +49,43 @@ const Profile = ({ address }: { address: string | null }) => {
...
+  const router = useRouter();
+
   const { width } = useWindowDimensions();
   const isDark = useIsDarkMode();
   const contentWidth = useContentWidth();
   const { data } = useProfileNftTabs({
     profileId: profileData?.data?.profile.profile_id,
   });
+  /**
+   * defalut tab index.
```
axeldelafosse: Typo. `defalut` -> `default`

axeldelafosse (approved):
LFG ✨

packages/app/components/profile/index.tsx
alantoa: Thank you! My bad!

PR #1260, opened 2022-07-05
fix: description multiline and quotes 

intergalacticspacehighway:
# Why
if we have quotes or new line in description, collection doesn't appear on OpenSea. 

# How
- Hack - pass JSON escaped strings to contract. A clean fix would have to be done at the contract level.

# Test Plan
- test creating collection with multiline description or quotes


axeldelafosse (approved):
LFG ✨

PR #1264, opened 2022-07-05
fix: profile tab router

alantoa:
# Why

- profile tab load does not work.

# How

- give it the initial router params will fix it!

# Test Plan

- check the tab list is working.


intergalacticspacehighway (approved):
LGTM ❇️

PR #1265, opened 2022-07-06
Feature/show2 951 hide follow button when the account is

alantoa:
# Why

- `Follow` button always show even if your current account is yourself.
- `SegmentedControl` default index incorrect on mobile web.

# How

- hide `Follow` button when the account is yourself.
- fixed `TabView` 's renderSceneHeader props.

# Test Plan

check if displayed normally when  `Follow` button is removed.


intergalacticspacehighway (approved):
LGTM 💯

PR #1266, opened 2022-07-06
fix: handle disconnected wallet

intergalacticspacehighway:
# Why
Sometimes if a wallet is disconnected but user is logged in, it still won't show it's disconnected during claim/drop.

# How
Use connected wallet address instead from our API.

# Test Plan
Test drop/claim

# Aside
Need to prioritise cleaning up web3 hooks. Right now it has some files for different platforms and it's making it hard to debug.


axeldelafosse (approved):
LFG ✨

PR #1268, opened 2022-07-06
fix: ledger wallet signature

intergalacticspacehighway:
# Why
Ledger wallet seems to be subtracting 27 in `v` value during signature https://github.com/LedgerHQ/ledgerjs/issues/466 which leads to an invalid signature. 

# How
Add 27 in the `v` value. Thanks to @karmacoma-eth 💯 

# Test Plan
Test drop/claim works with ledger wallet

### DIFF
--- a/packages/app/utilities.ts
+++ b/packages/app/utilities.ts
@@ -695,7 +695,7 @@ export const ledgerWalletHack = (signature?: string) => {
   if (signature) {
     const lastByteOfSignature = signature.slice(-2);
     if (lastByteOfSignature === "00" || lastByteOfSignature === "01") {
-      const temp = parseInt(lastByteOfSignature, 16) + parseInt("1b", 16);
+      const temp = parseInt(lastByteOfSignature, 16) + 27;
       const newSignature = signature.slice(0, -2) + temp.toString(16);
       return newSignature;
     }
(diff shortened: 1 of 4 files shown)


axeldelafosse (approved):
LFG ✨

packages/app/utilities.ts
```diff
@@ -689,3 +689,17 @@ export function isIOS(): boolean {
...
   return isAndroid() || isIOS();
 }
+
+// TODO: https://github.com/LedgerHQ/ledgerjs/issues/466
+export const ledgerWalletHack = (signature?: string) => {
+  if (signature) {
+    const lastByteOfSignature = signature.slice(-2);
+    if (lastByteOfSignature === "00" || lastByteOfSignature === "01") {
+      const temp = parseInt(lastByteOfSignature, 16) + parseInt("1b", 16);
+      const newSignature = lastByteOfSignature + temp.toString(16);
```
0xkarmacoma:
```suggestion
      const newSignature = signature?.slice(0, -2) + temp.toString(16);
```

packages/app/utilities.ts
```diff
@@ -689,3 +689,17 @@ export function isIOS(): boolean {
 export function isMobileWeb(): boolean {
   return isAndroid() || isIOS();
 }
+
+// TODO: https://github.com/LedgerHQ/ledgerjs/issues/466
+export const ledgerWalletHack = (signature?: string) => {
+  if (signature) {
+    const lastByteOfSignature = signature.slice(-2);
+    if (lastByteOfSignature === "00" || lastByteOfSignature === "01") {
+      const temp = parseInt(lastByteOfSignature, 16) + parseInt("1b", 16);
```
0xkarmacoma:
```suggestion
      const temp = parseInt(lastByteOfSignature, 16) + 27;
```

intergalacticspacehighway: optimization instincts. I like it 😂

PR #1269, opened 2022-07-06
fix: profile default tab

alantoa:
# Why

- set profile default tab index we need from backend API.
- profile tab data has a cache issue.

# How

delete the previous default index code and fix the cache.

# Test Plan

check if the profile default tab works.


axeldelafosse (approved):
LFG ✨

PR #1271, opened 2022-07-06
fix: drop/claim username in twitter intent

intergalacticspacehighway:
# Why
twitter intent shows wrong username.

# How
prepend `@` only when twitter link is available.

# Test Plan
test sharing drop/claim on twitter


axeldelafosse (approved):
LFG ✨

PR #1272, opened 2022-07-07
Chores

axeldelafosse:
# Why

We need to do some chores

# How

- Added some docs in the READMEs
- Renamed `next-react-native` to `next`
- Removed some legacy packages used in the previous web app

# Test Plan

Run the web app


intergalacticspacehighway (approved):
LGTM ❇️

PR #1274, opened 2022-07-07
fix: sentry on web

intergalacticspacehighway:
# Why
Sentry error logging is not working on web

# How
Add DSN env on vercel and follow https://docs.sentry.io/platforms/javascript/guides/nextjs/

# Test Plan
test if errors are getting logged on sentry. (try throwing error manually in app)


axeldelafosse (approved):
LFG ✨

PR #1275, opened 2022-07-08
fix: trending nft tap leads to open previous nft

intergalacticspacehighway:
# Why
 trending nft tap leads to open previous nft

# How
correction in calculating index

# Test Plan
tap an NFT from trending NFTs


alantoa (approved):
LGTM! ✨

PR #1277, opened 2022-07-08
fix: wallet disconnected but user stays logged in state

intergalacticspacehighway:
# Why
https://linear.app/showtime/issue/SHOW2-975/wallet-sometimes-gets-disconnected-while-user-stays-logged-in

# How
Replaces previous condition with a better approach

# Test Plan
Test drop and claim after disconnecting the wallet

# Aside
We're dealing with 4 different cases for web3 actions. Magic, web, mobile web and native. Will be prioritising to clean some stuff.


axeldelafosse (approved):
LFG ✨

PR #1278, opened 2022-07-08
Use new Trending NFTs endpoint

axeldelafosse:
# Why

We have a new Trending NFTs endpoint which includes drops

# How

- Decided to show "NFT" tab first instead of "CREATOR"
- Used the new endpoint and removed the pagination (the endpoint doesn't support pagination yet)

# Test Plan

Check the Trending screen on web and native


intergalacticspacehighway (approved):
LGTM ❇️

alantoa (approved):
LFG!🔥
Can't wait for new Trending endpoint!

PR #1279, opened 2022-07-08
waldo upload action after merge

intergalacticspacehighway:
waldo upload action should trigger after merging the PR in staging


axeldelafosse (approved):
LFG ✨

PR #1280, opened 2022-07-08
fix: TabView init error

alantoa:
# Why

TabView init error, just quickly fix patch, not sure useParam get `undefined` on Native, will dig it later.
# How

use `useState` replace to `useParam` which will be fixed later.

# Test Plan

check  Profile/Trending/Setting is working.


axeldelafosse (approved):
LFG ✨

PR #1281, opened 2022-07-11
fix: trending NFT press

intergalacticspacehighway:
# Why
Trending NFT press on the web doesn't navigate to NFT detail and on the native it opens incorrect NFT

# How
Add the navigation link on web, fix `days` param parsing on native

# Test Plan
Press trending NFT items on web and native


alantoa (commented):
LFG!✨

alantoa (approved):

PR #1282, opened 2022-07-11
fix: android sign in with email button is truncated

intergalacticspacehighway:
# Why
- Android sign in with email button is truncated https://showtime-rq88331.slack.com/archives/C02PXGK3V8D/p1657532536941539
- Some clean up, add types in `useWallet` hook

# How
Increase container height on android. @alantoa should we move to the new tab view for login? didn't want to block this fix so added a quick solution.

# Test Plan
Go to login form on android.

### DIFF
--- a/packages/app/components/login/index.tsx
+++ b/packages/app/components/login/index.tsx
@@ -22,7 +22,7 @@ interface LoginProps {
 }
 
 const CONTENT_HEIGHT = Platform.select({
-  android: [281, 397],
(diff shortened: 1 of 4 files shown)


alantoa (approved):
LGTM! ✨

alantoa:
>should we move to the new tab view for login? didn't want to block this fix so added a quick solution. 

@intergalacticspacehighway  yeah, let me do this!

PR #1284, opened 2022-07-12
Feature/show2 752 keyboard navigation support focus

catalinmiron:
# Why

# How

# Test Plan


catalinmiron:
Mobile | Desktop
-- | --
image | image


Preview:

https://user-images.githubusercontent.com/2805320/178475314-f5085e7a-33e8-4a2a-a5e8-0d614000be0d.mp4

packages/design-system/modal/modal.container.web.tsx
```diff
@@ -22,7 +24,7 @@ const MODAL_CONTAINER_TW = [
   "bg-white dark:bg-black",
   "shadow-xl shadow-black dark:shadow-white",
   "rounded-t-[32px] rounded-b-0 sm:rounded-b-[32px] pb-4",
-  "max-h-100vh",
+  "max-h-70vh sm:max-h-90vh",
```
alantoa: just tips, not blockers, last time we discussed using `max-h-82vh` on the desktop.
and could you publish this component on npm! thank you! 🙏

packages/design-system/modal/modal.container.web.tsx
```diff
@@ -1,6 +1,8 @@
 import { memo, useMemo } from "react";
 import { StyleSheet } from "react-native";
 
+import { useEscapeKeydown } from "@radix-ui/react-use-escape-keydown";
+import FocusTrap from "focus-trap-react";
```
alantoa, line 5: this component seems to only use on the web, could you move it to `app/lib/focus-trap-react/index.web.ts`? we need to unify.

alantoa (approved):
LGTM! ✨

axeldelafosse (approved):
LFG ✨

Same comments as Alan -- wrote some docs on Bit here: https://github.com/showtime-xyz/showtime-frontend/tree/staging/packages/design-system#bit

packages/app/lib/focus-trap-react/index.ts
```diff
@@ -0,0 +1,7 @@
+import FocusTrapWeb from "./index.web";
```
axeldelafosse: Not sure about this. This is what we want to avoid actually. If you are loading this from the native app, I'm afraid it won't work.

Can you do `import type { Props } from "focus-trap-react"` instead?

intergalacticspacehighway (approved):
LGTM ❇️

PR #1285, opened 2022-07-12
video mute/unmute button

intergalacticspacehighway:
# Why
Video lacks ability to mute/unmute audio

# How
add a mute/unmute button in video component

# Test Plan
Test video mute/unmute on all the platforms

# Aside
Follow up PR would be to experiment with FlashList and auto unmute videos based on user preference. I think we should try FlashList because viewability tracking with RecyclerListView on Android is not accurate and has some issues.

https://user-images.githubusercontent.com/23293248/178455250-c62a491c-6e59-4a91-b48d-ce5ee776bedd.mov


packages/app/components/mute-button/mute-button.tsx
```diff
@@ -0,0 +1,44 @@
+import { Platform, Pressable, StyleSheet } from "react-native";
+
+import { Muted, Unmuted } from "design-system/icon";
```
intergalacticspacehighway: replacing it with showtime universal imports

axeldelafosse (approved):
LFG ✨

alantoa (approved):
LGTM! ✨

alantoa:
> I think we should try FlashList  

@intergalacticspacehighway  cool! 
I'll replace `TabView`'s `RecyclerListView` with `FlashList` when you have done the feed!

alantoa:
btw do we need to use `FlashList` on the web?
tbh on the web, I really don't like how it behaves, I would prefer a virtual list.

intergalacticspacehighway:
@alantoa i'll try it today and let you know the findings!

PR #1286, opened 2022-07-12
Feature/show2 979 use default tab type

alantoa:
# Why
- types have some warning/error tips
- profile tab is not use `default_tab_type`.

# How

- fix ts type and clear code.
- use `default_tab_type` tab field.

# Test Plan

- check if the profile default tab works.


intergalacticspacehighway (approved):
LGTM ❇️

PR #1287, opened 2022-07-12
fix: disconnected wallet on web

intergalacticspacehighway:
# Why
sometimes wallet seems to be disconnected but `isConnected` from wagmi is true. Seems like this issue https://github.com/tmm/wagmi/issues/444

# How
Test if signer and chain are present along with isConnected. 

# Test Plan
Issue is not easily reproducible, we can test it if it happens again after the fix.


alantoa (approved):
LGTM! ✨

PR #1288, opened 2022-07-13
fix: burn nft broken

intergalacticspacehighway:
# Why
Burn NFT is not working

# How
- react hook form's `handleSubmit` returns a function, it needs to be called. https://github.com/showtime-xyz/showtime-frontend/compare/feature/show2-984-burn-nft?expand=1#diff-9e91a028c06aa860cf2f03cf298d448407027e4b76fb152dc4c5ca87fde06017R153
- we should check for `owner_address` instead of `creator_address` as owner is allowed to burn/transfer the NFT. Better solution would be -> https://showtime-rq88331.slack.com/archives/C02QD3J7DJM/p1657698754581359

# Test Plan
- Try deleting an NFT
- wanted to add E2E test case for burn but we need to get the indexer working on dev.

### DIFF
--- a/packages/app/components/create.tsx
+++ b/packages/app/components/create.tsx
@@ -19,8 +19,10 @@ import { tw } from "@showtime-xyz/universal.tailwind";
 import { Text } from "@showtime-xyz/universal.text";
 import { View } from "@showtime-xyz/universal.view";
 
+import { ConnectButton } from "app/components/connect-button";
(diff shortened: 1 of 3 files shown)


axeldelafosse (approved):
LFG ✨

alantoa (approved):
LGTM! ✨

PR #1290, opened 2022-07-13
fix: rainbowkit buttons not working

intergalacticspacehighway:
replace focus trap with a custom one.


yarn.lock
```diff
@@ -19807,7 +19806,7 @@ __metadata:
   languageName: node
   linkType: hard
 
-"focus-trap-react@npm:9.0.2, focus-trap-react@npm:^9.0.2":
+"focus-trap-react@npm:9.0.2":
```
axeldelafosse, line 19809: I think it's still included because the `modal` package wasn't updated via Bit

axeldelafosse (approved):
LFG ✨

packages/design-system/modal/modal.container.web.tsx
```diff
@@ -78,4 +77,67 @@ const styles = StyleSheet.create({
...
+      } else {
+        contentRef.current.focus();
+      }
+    }
+  }, [contentRef]);
+
+  return (
+    <>
+      <div
+        onFocus={useCallback(
```
catalinmiron: can we move this outside the render?

packages/design-system/modal/modal.container.web.tsx
```diff
@@ -78,4 +77,67 @@ const styles = StyleSheet.create({
...
+        tabIndex={0}
+      />
+      <div
+        ref={contentRef}
+        style={{ outline: "none" }}
+        tabIndex={-1}
+        {...props}
+      />
+      <div
+        onFocus={useCallback(
```
catalinmiron: same

catalinmiron (approved):
nit: please move `useCallback` outside the render

packages/design-system/modal/modal.container.web.tsx
intergalacticspacehighway: done!

PR #1291, opened 2022-07-13
Bring back backdrop filter to the dropdowns on web

catalinmiron:
# Why

We lost the backdrop filter on web

# How

The backdrop css styles were there but they were added to the wrong container, instead we should have added it to the [role="menu"] container (the dropdown menu main wrapper)

# Test Plan

Light | Dark
-- | -- 
image | image


axeldelafosse (approved):
LFG ✨

PR #1292, opened 2022-07-13
Feature/show2 645 error boundary

alantoa:
# Why

we need to improve `ErrorBoundary` components to make errors appear more friendly!

# How

improve the `ErrorBoundary` component, and add a generic `Fallback` component. supports: 

- [x] Retry
- [x] Message
- [x] Use a fallback component to display something

WX20220713-203800 

# Test Plan

check this component if is looks good!


packages/app/components/error-boundary/index.tsx
```diff
@@ -0,0 +1,114 @@
...
+    this.reset();
+  };
+
+  reset() {
+    this.setState(initialState);
+  }
+
+  componentDidCatch(error: Error, info: React.ErrorInfo) {
+    if (!this.props.disableCaptureException) {
+      // Todo: maybe we can reported the error log to the server
```
alantoa: do we need to send a report to the `sentry` or other options?

axeldelafosse: Good idea!

axeldelafosse (approved):
LFG ✨

PR #1293, opened 2022-07-13
fix: prevent scroll on focus

intergalacticspacehighway:
# Why
fix scroll on focusing item in focus trap

# How
`preventScroll` in focus call

# Test Plan
modals should trap focus and should not scroll the body


axeldelafosse (approved):
LFG ✨ Need to release the `modal` component

catalinmiron (approved):

PR #1298, opened 2022-07-14
fix: screen click not working when Toast appears

alantoa:
# Why

- screen click not working when Toast appears on web
- Storybook toast not working on web.

# How
- set `pointerEvents = none` on web.
- use `@showtime-xyz/universal.safe-area`'s `SafeAreaProvider`.

# Test Plan

check toast appears screen click event if works.


axeldelafosse (approved):
LFG ✨

intergalacticspacehighway (approved):
LGTM 🚀

PR #1300, opened 2022-07-15
feat: add shareNFTOnTwitter to dropdown

alantoa:
# Why

we want open the Share feature again.

# How

open Share again and add `shareNFTOnTwitter` hooks easy to share web pages to Twitter.

image

# Test Plan

check if the `Share` feature works.

### DIFF
--- a/packages/app/hooks/use-share-nft.ts
+++ b/packages/app/hooks/use-share-nft.ts
@@ -41,7 +41,7 @@ export const useShareNFT = () => {
         message: ``,
       })
     );
-    track("NFT Shared on Twitter", { type: "Twitter" });
+    track("NFT Shared", { type: "Twitter" });
   };
 
   return { shareNFT, shareNFTOnTwitter };
--- a/packages/app/hooks/use-share-nft.ts
+++ b/packages/app/hooks/use-share-nft.ts
@@ -4,8 +4,7 @@ import { useShare } from "app/hooks/use-share";
 import { track } from "app/lib/analytics";
 import { CHAIN_IDENTIFIERS } from "app/lib/constants";
 import { NFT } from "app/types";
-
-import { getTwitterIntent } from "../utilities";
+import { getTwitterIntent } from "app/utilities";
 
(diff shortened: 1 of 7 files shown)


packages/app/hooks/use-share-nft.ts
```diff
@@ -1,29 +1,49 @@
+import { Linking } from "react-native";
+
 import { useShare } from "app/hooks/use-share";
 import { track } from "app/lib/analytics";
 import { CHAIN_IDENTIFIERS } from "app/lib/constants";
 import { NFT } from "app/types";
 
+import { getTwitterIntent } from "../utilities";
```
intergalacticspacehighway: nit: app/utilities (ocd)

packages/app/hooks/use-share-nft.ts
```diff
@@ -1,29 +1,48 @@
...
+      nft?.chain_identifier
+    )}/${nft?.contract_address}/${nft?.token_id}`;
+    // Todo: add share Claim/Drop copytext
+    Linking.openURL(
+      getTwitterIntent({
+        url,
+        message: ``,
+      })
+    );
+    track("NFT Shared on Twitter", { type: "Twitter" });
```
axeldelafosse: `NFT Shared on Twitter` should stay `NFT Shared`. Adding the `type` is perfect, don't need to create another event.

axeldelafosse (approved):
LFG ✨

packages/app/hooks/use-share-nft.ts
axeldelafosse: Haha it's not a nitpick, no worries. Can we find a way to get absolute imports with auto import on VS Code? I suspect this was added by auto import, right?

intergalacticspacehighway (approved):
LGTM ❇️

packages/app/hooks/use-share-nft.ts
alantoa: yeah it's auto import, using absolute paths will solve it!

PR #1301, opened 2022-07-15
improve mint UX

intergalacticspacehighway:
# Description

This PR makes the mint experience consistent on all the platforms (for now)

The snackbar UX is better in long term but right now it's a bit weird because wallet redirect happens unexpectedly.  This PR keeps the user in create form and updates UI similar to create drop experience (polygon scan link followed by success screen).

https://user-images.githubusercontent.com/23293248/179249719-404d977e-af3a-46b0-bad4-4b8bee4e259f.mp4

# Bugs
On mobile web, tapping on sign button during mint leads to the web link of the wallet.

Before

https://user-images.githubusercontent.com/23293248/179242878-1d6b7a3a-2665-477d-9f58-ad78fac37eb0.MP4

After

https://user-images.githubusercontent.com/23293248/179247231-6594daf6-cd44-42a7-b29b-8ec4ad9655c9.MP4

# How
- pre populate gas limit https://github.com/showtime-xyz/showtime-frontend/compare/mint-improv?expand=1#diff-1d76b368767dfd437cccac6ecb23189c95438c1ec6c9049ea0c2bed6571cc09fR29

# Test Plan
- Test minting on all platforms (including web mobile) with magic and normal wallet.
- Will add E2E testcases on this build


packages/app/components/create.tsx
```diff
@@ -463,65 +480,46 @@ function Create() {
...
+              <Button
+                tw={`mt-4 ${isSignRequested ? "opacity-60" : ""}`}
+                size="regular"
+                variant="primary"
+                disabled={isSignRequested}
+                onPress={() => {
+                  signTransaction(signMessageData.data);
+                }}
+              >
+                {isSignRequested ? "Signing..." : "Sign Message"}
```
axeldelafosse: Maybe we could simplify `Sign Message` to `Sign`?

packages/app/components/create.tsx
```diff
@@ -463,65 +480,46 @@ function Create() {
...
-      </CreateScrollView>
 
-      <View tw="mt-8 w-full px-4">
-        {shouldShowSignMessage ? (
-          <View tw="px-2">
-            {!isSignRequested ? (
+        <View tw="mt-8 w-full px-4">
+          {shouldShowSignMessage ? (
+            <View tw="px-2">
               <Text tw="text-center text-lg dark:text-gray-400">
```
axeldelafosse: Not sure about `text-lg` here: the font is different and the full bold is pretty hard on the eyes. What do you think?

axeldelafosse (approved):
LFG ✨

alantoa (approved):
LFG!! ✨

packages/app/components/create.tsx
axeldelafosse: Would be the case for `Sign Message` in `drop` and also `Sign the message` in `login`

packages/app/components/create.tsx
axeldelafosse: Same comment applies to `drop` and `login` too

intergalacticspacehighway: yeah, wdyt about below? (`text-base` and `Sign`)

Screenshot 2022-07-15 at 8 48 42 PM

intergalacticspacehighway: cc - @axeldelafosse

PR #1304, opened 2022-07-18
Feature/show2 980 Twitter video card preview

alantoa:
# Why

Twitter video cards can't preview

# How

when the NFT type is video, we turn on the preview.

# Test Plan

check if the Twitter video card works.


catalinmiron (approved):
LGTM

PR #1305, opened 2022-07-18
fix: notification issues

alantoa:
# Why

- notification list Avatar uses the NFT address.
- text overlapped on Native.
- we not setting max-width on Desktop.
- loading of the notifications list look like a bug

# How

- use `actor` field to the avatar.
- use `programmatically` router instead of declarative.
- adjust style.

# Test Plan

check if the notification looks good!

### DIFF
--- a/packages/app/components/header.tsx
+++ b/packages/app/components/header.tsx
@@ -38,7 +38,11 @@ import {
 import { useNavigateToLogin } from "app/navigation/use-navigate-to";
 import { useNavigationElements } from "app/navigation/use-navigation-elements";
 
-import { breakpoints } from "design-system/theme";
+import {
+  breakpoints,
+  CARD_DARK_SHADOW,
+  CARD_LIGHT_SHADOW,
+} from "design-system/theme";
 
 import { withColorScheme } from "./memo-with-theme";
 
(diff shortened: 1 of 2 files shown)


packages/app/components/header.tsx
```diff
@@ -180,9 +184,12 @@ const NotificationsInHeader = () => {
         onCloseAutoFocus={(e) => e.preventDefault()}
       >
         <View
-          tw="mt-2 overflow-hidden rounded-3xl bg-white shadow-lg shadow-black dark:bg-black dark:shadow-white"
+          tw="mt-2 overflow-hidden rounded-3xl bg-white dark:bg-black "
           style={Platform.select({
-            web: { maxHeight: "calc(50vh - 64px)" },
+            web: {
+              maxHeight: "calc(50vh - 64px)",
+              boxShadow: isDark ? CARD_DARK_SHADOW : CARD_LIGHT_SHADOW,
```
axeldelafosse, line 191: I don't remember: why are we doing this instead of customizing the shadows in the Tailwind config?

axeldelafosse (approved):
LFG ✨

packages/app/components/header.tsx
alantoa: because now the tailwind default shadow looks very bad and config seems to not available.

I also don't like that, when replacing to `NativeWind`, I'll try to fix it.

PR #1306, opened 2022-07-18
Fix/notifications max width

alantoa:
# Why
- notification list item is not auto warp.
- usePlatformBottomHeight does not work on the web.

# How

- adjust style.
- use `usePlatformBottomHeight` replace  to `useBottomHeight`.

# Test Plan

check if the notification looks good!


alantoa:
just fix the last PR's bug, will merge!

PR #1309, opened 2022-07-21
fix: report NFT

alantoa:
# Why

report the NFT should be use `nft.nft_id` not `nft.token_id`

# How

use `nft.nft_id` replace with  `nft.token_id`

# Test Plan

check if the report NFT works.


alantoa:
I have checked and it works, will merge!

PR #1310, opened 2022-07-21
selectors on elements for e2e tests

intergalacticspacehighway:
# Why
adds some selectors on elements required for e2e tests on web

# Test Plan
run web app and visit profile


axeldelafosse (approved):
LFG ✨

PR #1314, opened 2022-07-22
fix: notification issue

alantoa:
# Why

- when the `actors` is empty, we need to use the `link_to_profile` field.
- claimed copy text needs improvement.

# How

- use `link_to_profile` field.
- replace "@X claimed the drop Y" with "@X claimed Y by @Z".

# Test Plan

check that the fields are correct.


axeldelafosse (approved):
LFG ✨

PR #1316, opened 2022-07-25
fix: notifications copy text

alantoa:
# Why

we miss the `followed you` text.

# How

move the `type` text and `actors` code.

# Test Plan
check if the copy of the notification looks good!


axeldelafosse (approved):
LFG ✨

PR #1321, opened 2022-07-26
fix: activity modal should fallback to user address if username is null

intergalacticspacehighway:
# Why
https://linear.app/showtime/issue/SHOW2-993/activity-modal-should-fallback-to-use-address-if-username-is-null

- make it a link because why not.

# How
- use address as fallback if username is null

# Test Plan


axeldelafosse (approved):
LFG ✨

PR #1322, opened 2022-07-26
feat: deeplinking

intergalacticspacehighway:
## Why?
Tapping nft links should open in native app if the app is installed.

## How?
For now, we'll handle the nft detail deep linking to the mobile app. After testing, will add profile deeplinking.

## Test
Once web app is deployed, tapping on nft link should open in native app if the app is installed.


axeldelafosse (approved):
LFG ✨

PR #1323, opened 2022-07-26
Install `react-native-quick-crypto` and `react-native-bignumber`

axeldelafosse:
# Why

@mrousavy just released `react-native-quick-crypto`: https://github.com/margelo/react-native-quick-crypto

# How

- Removed `react-native-crypto`
- Installed `react-native-quick-crypto`
- Installed `react-native-bignumber`

# Test Plan

Run the app and check Wallet Connect for example


patches/react-native+0.69.3.patch
```diff
@@ -0,0 +1,163 @@
...
+         return {
+diff --git a/node_modules/react-native/Libraries/Blob/FileReader.js b/node_modules/react-native/Libraries/Blob/FileReader.js
+index c58dfba..2f3aa15 100644
+--- a/node_modules/react-native/Libraries/Blob/FileReader.js
++++ b/node_modules/react-native/Libraries/Blob/FileReader.js
+@@ -10,6 +10,7 @@
+ 
+ const Blob = require('./Blob');
+ const EventTarget = require('event-target-shim');
++const { toByteArray } = require('base64-js');
```
mrousavy: why not react-native-quick-base64 here? they have converters for that

axeldelafosse: Good catch! I'm glad I updated this patch in this PR haha 🤣

PR #1325, opened 2022-07-26
Feature/show2 1023 handle the new notification schema

alantoa:
# Why

Jorge handles the new backend data schema, we need to use it!

# How

handle the new notification schema and clean the notification code. 

# Test Plan

it needs to deploy the backend and then test it!


intergalacticspacehighway (approved):
LGTM ❇️

PR #1328, opened 2022-07-27
Drop it like it's hot

axeldelafosse:
# Why

Let's focus on drops instead of regular mints. We might want to unify the user flows at some point but for now let's focus

# How

Used `/drop` instead of `/camera`

# Test Plan

Run the apps and make sure the flows are working


intergalacticspacehighway (approved):
LGTM ❇️

PR #1329, opened 2022-07-27
Mark a drop as NSFW

axeldelafosse:
# Why

Some users started to drop NSFW content

# How

- Added a switch to mark content as NSFW
- Added `pinataContent` attribute on media

# Test Plan

Run the app, click on (+) then toggle the switch "Explicit content"


intergalacticspacehighway (approved):
LGTM ❇️

PR #1330, opened 2022-07-27
fix: missing /t/* links in deeplinks

intergalacticspacehighway:
### Why?

Missed adding `/t` links support

### Test
Links with `/t` should open in the native app


axeldelafosse (approved):
LFG ✨

PR #1331, opened 2022-07-27
fix: video bg-blur not working

alantoa:
# Why

video background blur not working on native because the video preview doesn't have the `blurhash` field.

# How

we use `BlurView` to replace it when `blurhash` is null.

# Test Plan

check if the feed video looks good!


axeldelafosse (approved):
LFG ✨

PR #1332, opened 2022-07-28
Feature/show2 1036 we don’t update the like and comment

catalinmiron:
# Why

Comments and likes are not reflecting the real count stored on the server.

# How

For likes, we don't rely on optimistic updates anymore 
For comments, we're getting the total count from the `use-comments` hook and keeping the hook as single source of truth. The reason that comments count was not updating was that the fetched `nft` (either from feed or directly from nft details page) was not updating.

# Test Plan

https://user-images.githubusercontent.com/2805320/181483447-464882fe-951c-4a92-833e-1c7c975f2514.mp4


alantoa (approved):
LGTM! ✨

axeldelafosse (approved):
LFG ✨

PR #1333, opened 2022-07-28
Feature/show2 989 remove percentage on the reloading

alantoa:
# Why

Alex feedback we need to remove the percentage on the pull refresh because it diverts user's attention.

# How

clean the code and remove the percentage on the pull refresh, I'll add our design-system refresh control in next PR(because pull refresh doesn't work now in my local, will dig it).

# Test Plan

check pull-refresh it looks good on native!


axeldelafosse (approved):
LFG ✨

intergalacticspacehighway (approved):
LGTM ❇️

PR #1334, opened 2022-07-28
Show `Hidden` tab on profile + Unhide NFT

axeldelafosse:
# Why

We are now displaying a `Hidden` tab on your own profile. This PR enables users to `Unhide` their NFTs

# How

Displayed `Unhide` button in dropdown menu instead of `Hide` when displaying a NFT from the `Hidden` tab

# Test Plan

Run the app, go to your profile, hide some NFTs, go to the Hidden tab then unhide some NFTs


alantoa (approved):
LFG! ✨

PR #1335, opened 2022-07-28
feat: claimers list

alantoa:
# Why
- claim icon is not recognizable, we need to make a distinction between claimed and unclaimed.
- we can't watch the claimers' user list.

# How

- use 🎁 replace claimed icon.
- support watch claimers user list.

# Test Plan

check if the claim icon and claimer user list look good!

# Video 

https://user-images.githubusercontent.com/37520667/181526335-d96db627-0f0e-439c-9e66-d2b24068c56c.mov


apps/next/src/pages/_app.tsx
```diff
@@ -18,6 +18,7 @@ import { AppProviders } from "app/providers/app-providers";
 import { ActivitiesScreen } from "app/screens/activities";
 import { BuyScreen } from "app/screens/buy";
 import { ClaimScreen } from "app/screens/claim";
+import { ClaimersScreen } from "app/screens/cliamers";
```
axeldelafosse: `cliamers` -> `claimers`

packages/app/components/claim/gift-button.tsx
```diff
@@ -1,23 +1,59 @@
...
+            native: as,
+            web: router.asPath,
+          }),
+          { shallow: true }
+        );
+      }}
+      accentColor={textColors}
+    >
+      {edition?.is_already_claimed ? (
+        <Text tw="h-5 w-6 text-center text-xl">🎁</Text>
```
axeldelafosse: Not sure about the use of an emoji here but it's better than what we have today!

axeldelafosse (approved):
LFG ✨

apps/next/src/pages/_app.tsx
alantoa: damn, my bad. 🤣

packages/app/components/claim/gift-button.tsx
alantoa: It would be better if there is a dedicated icon!

PR #1338, opened 2022-07-29
fix: attempt to fix missing sign requests

intergalacticspacehighway:
# Why
noticed we're using web3 object instead of wagmi's sign method. I am not sure if this will fix the issue, but worth the try.

# How
use web3 object for signing only when `isMagic` is true.

# Test Plan
test claim and drop is working with magic and normal wallet


alantoa (approved):
LGTM! ✨

PR #1341, opened 2022-08-01
allow 0 in royalty percentage

intergalacticspacehighway:
# Why
https://linear.app/showtime/issue/SHOW2-1050/drop-screen-allow-royalties-to-be-0percent

# Test Plan
Try putting 0 in royalty percentage while creating drop


alantoa (approved):
LGTM!🔥

PR #1342, opened 2022-08-01
Feature/show2 998 trending list

alantoa:
# Why

- the trending list performance is very low on desktop, user experience is not good.
- many of our file types have many warnings/errors.
- recyclerlistview is not good for the web.

# How

### implement a InfiniteScrollList
- web: use `react-virtuoso`, the API will same as FlatList.
- native: use `FlatList`, we can use `FlashList` or wait for `WishList` later, their APIs are similar to `FlatList` IMO.
 
----
- fix many files type.

# Test Plan

- check the new Trending list performance and bug, and make sure it running well!


intergalacticspacehighway (approved):
LGTM ❇️

axeldelafosse (approved):
LFG ✨

PR #1343, opened 2022-08-01
fix: turn autoConnect off till we identify signature requests bug

intergalacticspacehighway:
# Why
Wallets sometimes do not respond to signature requests even If it's in connected state. 

# How
Set `autoConnect` to false, user will be shown connect button on trying to drop/claim. 

# Test Plan
Login and test claim and drop, kill the app, try to claim and drop

# Aside

This issue seems to be related https://github.com/rainbow-me/rainbow/issues/3870, there are also some instances when the wallet is shown to be in a connected state but it doesn't respond to signature requests.


alantoa (approved):
LGTM! ✨

axeldelafosse (approved):
LFG ✨

PR #1344, opened 2022-08-01
fix: android create drop ui issues

intergalacticspacehighway:
# Why
Found UI issues in drop form on android

# How
Add header and bottom padding in drop form

# Test Plan
Test creating drop on android

# Aside
- Will add e2e for android
- We should make the header/footer consistent on android and iOS.


axeldelafosse (approved):
LFG ✨

PR #1345, opened 2022-08-01
fix: notification nft name copy

alantoa:
# Why

we didn't divide the notification's NFT display copy text.

# How

splice the NFTs list's name.

# Test Plan

check if the notification NFTs‘ copy looks good!


alantoa:
just fixed the copy, will merge it first.

PR #1346, opened 2022-08-02
flashlist ⚡ ⚡ ⚡ 

intergalacticspacehighway:
# Why
- Recyclerlistview's API is bad. (too much hardcoding is needed)
- Recyclerlistview has issues with viewable events on android. 
- FlashList is working super smooth for feed

# How
- Replace recyclerlistview with FlashList for swipe-list component

# Test Plan
- Test feed/swipe-list on android, iOS and web


packages/app/hooks/use-infinite-list-query.ts
```diff
@@ -39,7 +39,7 @@ export const useInfiniteListQuerySWR = <T>(
     isLoading,
   } = useSWRInfinite<T>(urlFunction, fetcher, {
     revalidateFirstPage: true,
-    suspense: true,
+    // suspense: true,
```
axeldelafosse, line 42: We might want to revisit this later

axeldelafosse (approved):
LFG ✨

alantoa (approved):
LFG! 🔥

packages/app/hooks/use-infinite-list-query.ts
intergalacticspacehighway: added it back. I think this solves it. https://github.com/showtime-xyz/showtime-frontend/pull/1346/files#diff-8bad9b025a5c54b5d6ca8cd87337cb619e02274a68ceb62d22d4959dacc92d76R50

There's some part of me that doesn't like it very much. 😖

intergalacticspacehighway: really makes it hard to do stuff like show the stale content along with a  loading spinner.  Show the error along with the stale content

intergalacticspacehighway: reverted it, the list was flashing (literal flash-list) due to suspense. Will revisit!
https://github.com/showtime-xyz/showtime-frontend/pull/1346/commits/e5d7c6ce1d4ff9c9c047103d087c12ccae73e831

PR #1347, opened 2022-08-02
Feature/show2 1049 home feed

alantoa:
# Why
- we want to get rid of `recyclerlistview` on the web, and `VirtuosoList` is better on the web.
- `useInfiniteListQuerySWR` hooks' load more has a bug, we seem to haven't noticed.
- notifications list loads-more have bug.

# How

- improve `VirtuosoList`, on prop `numColumns` === 1. use `Virtuoso` component, on prop `numColumns` > 1. we use `VirtuosoGrid` component. will be better for performance and features. 

- add `isReachingEnd` field on `useInfiniteListQuerySWR` hooks, and the `fetchMore` method will not be running when the list is loading completed.
- replace `FlatList` of in `notifications` with new `InfiniteScrollList`.

# InfiniteScrollList On Native Progress ⌛️
- will try to replace `FlatList` in `infinite-scroll-list` with `FlashList`, I need to do some tests before this, make sure it work well!

# Test Plan

- Test if the new Home/Trending list performs well.
- Check if the notification list looks good!


packages/app/components/feed/feed.md.tsx
```diff
@@ -162,109 +155,69 @@ export const FeedList = () => {
...
-  const { width: screenWidth, height } = useWindowDimensions();
-  let dataProvider = useMemo(
-    () =>
-      new DataProvider((r1, r2) => {
-        return r1.nft_id !== r2.nft_id;
-      }).cloneWithRows(data),
-    [data]
-  );
-
-  const _layoutProvider = useMemo(
```
intergalacticspacehighway, line 187: mah man. This API used to give me PTSD 😌

packages/app/components/trending/trending-creators-list.md.tsx
```diff
@@ -0,0 +1,69 @@
+import { useState, useCallback } from "react";
+import { ListRenderItemInfo } from "react-native";
+
```
intergalacticspacehighway, line 3: no problem with this file but I think we won't be using trending creators for the time being.

packages/app/components/feed/feed.md.tsx
alantoa: +1!! especially on the web, too many weird issues!
I'm very empathetic and battle with this API! 🤣

packages/app/global.d.ts
```diff
@@ -0,0 +1 @@
+declare module "@biconomy/mexa";
```
intergalacticspacehighway, line 1: 💯

intergalacticspacehighway (approved):
LGTM 🔥 🔥 🔥

packages/app/components/trending/trending-creators-list.md.tsx
alantoa, line 3: yeah I know, I have been tangled for a while to change this file 🤣, but I thought I might use it in the future, so I improved. (maybe it's my OTD I want consistency.)

packages/app/global.d.ts
axeldelafosse: Unrelated but we can probably get rid of this package altogether

packages/app/providers/user-provider.tsx
```diff
@@ -60,7 +60,7 @@ export function UserProvider({ children }: UserProviderProps) {
     const identifyAndRegisterPushNotification = async () => {
       if (data) {
         // Identify user
-        mixpanel.identify(data.data.profile.profile_id.toString());
+        mixpanel.identify(data.data?.profile.profile_id.toString());
```
axeldelafosse, line 63: We can remove this `mixpanel.identify` -- we don't use it anymore

axeldelafosse (approved):
LFG ✨

PR #1349, opened 2022-08-02
Install `react-native-quick-crypto` and `react-native-bignumber`

axeldelafosse:
# Why

Same as https://github.com/showtime-xyz/showtime-frontend/pull/1323. I had to remove the libraries while debugging an issue when building the app for release.

# How

- Set env `"FLIPPER_DISABLE": true` in prod on EAS
- Removed `react-native-crypto`
- Installed `react-native-quick-crypto`
- Installed `react-native-bignumber`

# Test Plan

Run the app and check Wallet Connect for example


intergalacticspacehighway (approved):
LGTM ❇️

PR #1350, opened 2022-08-02
feat: make profile required

intergalacticspacehighway:
# Why
Make profile required before claim/drop

https://linear.app/showtime/issue/SHOW2-1043/make-required-the-profile-screen-of-edit-profile-except-for-cover

# Test Plan
Try claiming/drop with incomplete profile

Need to discuss which fields should be required

https://showtime-rq88331.slack.com/archives/C02S5PQ1XJT/p1659442006883739?thread_ts=1659441560.369599&cid=C02S5PQ1XJT


intergalacticspacehighway:
gonna mess up e2e cases 😖

packages/app/components/drop/drop-form.tsx
```diff
@@ -147,6 +150,28 @@ export const DropForm = () => {
...
   }
 
+  if (
+    !userProfile?.data.profile.username ||
+    userHasIncompleteExternalLinks(userProfile?.data.profile) ||
+    !userProfile?.data.profile.bio ||
+    !userProfile?.data.profile.img_url
+  ) {
+    return (
+      <View tw="flex-1 items-center justify-center px-10 text-center">
```
alantoa, line 160: it would be better if encapsulate this component, we might use it in more places.

alantoa (approved):
LGTM! ✨

packages/app/components/drop/drop-form.tsx
intergalacticspacehighway: yeah, intentionally kept it like this. I don't like doing early abstractions 😂 . Will move it once we finalise if we're keeping this functionality.

PR #1351, opened 2022-08-02
fix: initial switch network prompt

intergalacticspacehighway:
# Why
Fixes the initial switch network prompt to mainnet. We'll try to switch to polygon, rainbowkit also handles adding the network if it's not added.

# How
There's a prop to specify `initialChain` in `RainbowKitProvider`

# Test Plan
Try login in


alantoa (approved):
LGTM! ✨

axeldelafosse (approved):
LFG ✨

PR #1354, opened 2022-08-03
show reconnect wallet message on drop/claim

intergalacticspacehighway:
# Why
- Added some findings regarding wallet connect missing signature bug here. https://github.com/rainbow-me/rainbow/issues/3870
- Until we find a better solution, we'll show a message to reconnect their wallet in case signature requests do not arrive in wallet app

# Test Plan
- Try claim/drop


axeldelafosse (approved):
LFG ✨

PR #1355, opened 2022-08-03
Fix/misc issues

alantoa:
# Why
- notification list not working on the web small screen.
- view NFT can't go back when haven't history.

# How

- notification add use window scroll props.
- use `history.length` judge to can go back.

# Test Plan


axeldelafosse (approved):
LFG ✨

PR #1356, opened 2022-08-03
Send `nsfw` query param to `poll-edition` endpoint

axeldelafosse:
# Why

We want to flag NSFW drops in our backend directly

# How

- Sent `nsfw` query param to `poll-edition` endpoint
- Improved types

# Test Plan

Flag a drop as NSFW while creating it and check if the backend is flagging it correctly


intergalacticspacehighway (approved):
LGTM ❇️

alantoa (approved):
LGTM! ✨

PR #1357, opened 2022-08-03
scroll to bottom when signature request missing message is shown

intergalacticspacehighway:
# Why
- Noticed that on web mobile, it's easy to miss the message. So, we scroll to bottom when it's shown. Also, claim modal height is small in web mobile, will increase it in next PR.
- Fix not able to pass `ref` in scrollview

# How
- Scroll to bottom when signature request missing message is shown

# Test Plan
- Test drop/claim

# Aside
this message UI is really a draagg.


packages/app/components/claim/claim-form.tsx
```diff
@@ -256,7 +257,11 @@ export const ClaimForm = ({ edition }: { edition: CreatorEditionResponse }) => {
           </View>
 
           {state.signaturePrompt && !isMagic ? (
-            <MissingSignatureMessage />
+            <MissingSignatureMessage
+              onMount={() => {
+                scrollViewRef.current?.scrollToEnd();
+              }}
```
axeldelafosse, line 263: 💯

axeldelafosse (approved):
LFG ✨

PR #1360, opened 2022-08-03
Pick first `libcrypto.so` and `libssl.so`

axeldelafosse:
# Why

We need to fix the Android build

# How

- Added `libcrypto.so` and `libssl.so` to our pick first config plugin

# Test Plan

Run `yarn android`


intergalacticspacehighway (approved):
LGTM ❇️

PR #1361, opened 2022-08-03
fix: edit-profile text align & validate username

alantoa:
# Why

- edit-profile link text hasn't aligned.
- username hasn't validated the special character.

# How

- fixed the above issues.

# Test Plan

- check if the text of the edit-profile link looks good!


axeldelafosse (approved):
LFG ✨

PR #1362, opened 2022-08-03
Use latest image for EAS Build

axeldelafosse:
# Why

We are experiencing weird issues with App Store returning an "Invalid Binary" error from time to time. Let's try to upgrade to Xcode 13.4 https://docs.expo.dev/build-reference/infrastructure/#image--macos-monterey-124-xcode-134--alias--latest

# How

Used `latest` image in `eas.json`


intergalacticspacehighway (approved):
LGTM ❇️

PR #1364, opened 2022-08-05
Feature/show2 985 resizing and cropping images

alantoa:
# Why

- many users' avatar sizes are not good, we need to support crop images on avatars & banners.

# How
- currently, only support the web, native will later.
- use `react-easy-crop` (the size of the package is very small) and `Modal` to implement a new cropper page, and use canvas and `File` to save when the image is cropped.

# Test Plan

- check if the cropper looks good!

### On Desktop:
https://user-images.githubusercontent.com/37520667/183034704-21e20b5c-c351-4ffd-9c77-2993b58ea348.mov

### On Mobile:
image


packages/app/components/preview/index.tsx
```diff
@@ -78,7 +84,18 @@ export const getLocalFileURI = (file?: string | File) => {
...
   return file;
 };
+/**
+ * Browsers will release object URLs automatically when the document is unloaded;
+ * for optimal performance and memory usage
+ * if there are safe times when you can explicitly unload them, you should do so.
+ * @param uri
+ * @returns
+ */
+export const revokeObjectURL = (uri: any) => {
```
intergalacticspacehighway, line 98: interesting 💯

packages/app/lib/canvas-utils.ts
```diff
@@ -0,0 +1,121 @@
...
+  // paste generated rotate image at the top left corner
+  ctx.putImageData(data, 0, 0);
+
+  // As Base64 string
+  // return canvas.toDataURL('image/jpeg');
+
+  // As a blob
+
+  return new Promise((resolve, reject) => {
+    canvas.toBlob((file) => {
```
intergalacticspacehighway, line 88: 💯

packages/design-system/modal/modal.header.tsx
```diff
@@ -31,9 +32,13 @@ function ModalHeaderComponent({
   );
   return (
     <View style={containerStyle}>
-      <Button variant="tertiary" size="regular" onPress={onClose} iconOnly>
-        <Close width={20} height={24} />
-      </Button>
+      {StartContentComponent ? (
```
intergalacticspacehighway, line 35: 💯

intergalacticspacehighway (approved):
LGTM 🔥 🔥 🔥

PR #1366, opened 2022-08-05
Fix/kill multiple owner list

intergalacticspacehighway:
# Why
We need to migrate to unified NFT schema in all NFT APIs. This schema won't have `multiple_owner_list`. So we'll 🔫  it (violence is bad, don't do it 💟 )

# How
- Remove usage of `multiple_owner_list` from all APIs and replace with `is_user_owner` new flag

# Test Plan
Test if anything is not crashing


alantoa (approved):
LFG! 🔥

PR #1367, opened 2022-08-05
feat: improve claim button copy

alantoa:
# Why

- we need to improve the claim button copy text. 
https://linear.app/showtime/issue/SHOW2-1018/claim-button-copy-text

# How

- clean code and add some logic.

# Test Plan

- check if the new claim button copy looks good!

### DIFF
--- a/packages/app/utilities.ts
+++ b/packages/app/utilities.ts
@@ -171,10 +171,8 @@ export const overrideMagicInstance = (email: string) => {
 
 // Format big numbers
 export function formatNumber(number: number) {
-  if (!number) return 0;
-  // for the edge case of 100k, our max supply, put “100k”, no decimals
-  if (number >= 100000) {
-    return `100k`;
+  if (number > 1000000) {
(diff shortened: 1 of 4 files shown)


packages/app/utilities.ts
```diff
@@ -171,8 +171,10 @@ export const overrideMagicInstance = (email: string) => {
 
 // Format big numbers
 export function formatNumber(number: number) {
-  if (number > 1000000) {
-    return `${(number / 1000000).toFixed(1)}m`;
+  if (!number) return 0;
+  // for the edge case of 100k, our max supply, put “100k”, no decimals
+  if (number >= 100000) {
```
axeldelafosse: We also use this function for likes and comments, even if this is not the case it might introduce a bug later on

alantoa: yeah my bad, will add a new method to use for claims.

intergalacticspacehighway (approved):
LGTM ❇️

PR #1368, opened 2022-08-05
upgrade moti

intergalacticspacehighway:
# Why
- no change in bundle size, but upgraded it.
- move moti to peer dependency for design system


alantoa (approved):
LGTM! ✨

PR #1369, opened 2022-08-08
drop nft set default supply to 100 and remove some web3 actions

intergalacticspacehighway:
# Why
 - remove list, unlist, delete and transfer options from dropdown
 - set default drop nft supply to 100


alantoa (approved):
LGTM! ✨

PR #1370, opened 2022-08-08
feat: resize profile img on native

alantoa:
# Why

last week I noticed that expo has the `allowsEditing` option, it supports crop images on native.

https://docs.expo.dev/versions/latest/sdk/imagepicker/#imagepickeroptions

# How

- on iOS, it uses [UIImagePickerController](https://developer.apple.com/documentation/uikit/uiimagepickercontroller), so currently, custom cropping aspect areas don't support it. maybe we can add 3rd lib later if it is necessary.
- on Android, it uses [Android-Image-Cropper](https://github.com/ArthurHub/Android-Image-Cropper).

# Test Plan

check if the cropper looks good on native!

### DIFF
--- a/packages/app/components/edit-profile.tsx
+++ b/packages/app/components/edit-profile.tsx
@@ -284,7 +284,13 @@ export const EditProfile = () => {
                   render={({ field: { onChange, value } }) => (
                     <Pressable
                       onPress={async () => {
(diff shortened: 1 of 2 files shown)


intergalacticspacehighway (approved):
LGTM ❇️

axeldelafosse (approved):
LFG ✨

PR #1371, opened 2022-08-08
fix: improve reconnect wallet flow

intergalacticspacehighway:
# Why
- Improves reconnect flow, fixes some quirks on disconnect and reconnecting on wallet connect and rainbowkit

# How
- show the user a message to retry claiming after reconnecting, ideally it should auto start reclaiming, but I think this is good for now, will do that after testing this for a while.

# Test Plan
- Testing is a bit tricky on this one as we don't know when the session ends, but you can try something like the below on web. Native has a similar flow.

https://user-images.githubusercontent.com/23293248/183385144-11a2cadd-2aae-4af5-9954-4e3abe1badfe.mov

### DIFF
--- a/packages/app/components/claim/claim-form.tsx
+++ b/packages/app/components/claim/claim-form.tsx
@@ -35,7 +35,9 @@ import {
 } from "app/utilities";
 
 export const ClaimForm = ({ edition }: { edition: CreatorEditionResponse }) => {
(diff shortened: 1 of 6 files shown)


axeldelafosse (approved):
LFG ✨

alantoa (approved):
LGTM! ✨

PR #1372, opened 2022-08-08
fix: nft detail height

alantoa:
# Why

- NFT detail height not fill on screen.

### Before
image

# How

- fix the NFT height.

# Test Plan

- check if the mobile feed looks good.

### After

image


intergalacticspacehighway (approved):
LGTM ❇️

PR #1374, opened 2022-08-08
fix: ModalScreen when in BottomTab

alantoa:
# Why

https://linear.app/showtime/issue/SHOW2-1086/congrats-screen

# How

- ModalScreen adds padding/margin when in BottomTabScreen via `useModalScreenViewStyle` hooks.

# Test Plan

- check if the screen is covered when the `ModalScreen` in `BottomTabScreen`.


axeldelafosse (approved):
LFG ✨

PR #1376, opened 2022-08-08
fix: mobile web hide feature is not accessible

intergalacticspacehighway:
# Why
Hide is not accessible on mobile web browser as we redirect to NFT detail screen to show the dropdown

# How
- use query parameter to pass the tab info to next screen

# Test Plan
Hide/Unhide should be accessible on mobile web


alantoa (approved):
LGTM! ✨

axeldelafosse (approved):
LFG ✨

PR #1377, opened 2022-08-08
Feature/show2 1089 notifications list scroll & height issue

alantoa:
# Why

- notifications list sometimes can't scroll.
- list of the bottom item is obscured.

# How

- set the list height to quickly fix it on the mobile web.

# Test Plan

check if the notification list looks good.


alantoa:
Just a quick fix for an urgent bug just found, will merge and deploy it!

PR #1379, opened 2022-08-09
Handle errors when reaching OZ quota

axeldelafosse:
# Why

We need to handle errors when we reach our OZ quota

# How

- Updated copy for error 420
- Added an `Alert` on error 500 when drop or claim failed

# Test Plan

Wasn't able to test this change, we need to simulate getting an error 500 from the backend


intergalacticspacehighway (approved):
LGTM ❇️

PR #1381, opened 2022-08-09
Feature/show2 1084 check the local time issue

alantoa:
# Why

- localized time is not used in many places. [SHOW2-1084](https://linear.app/showtime/issue/SHOW2-1084/check-the-local-time-issue#comment-a39e987b)

# How

- add `convertUTCDateToLocalDate` method to use convert local time. 

# Test Plan

check if the Comment/NFT's time is correct.


axeldelafosse (approved):
LFG ✨

PR #1386, opened 2022-08-10
refactor: kill the FlatList

alantoa:
# Why

- `FlatList` has bad performance, let's switch all FlatList to new `InfiniteScrollList`.
-  ready to release the `Tab/List` component. but the Tab's list type is not good.

currently, only some unused/design system components remain, and I'm not going to change them.

image

# How
use InfiniteScrollList to replace all FlatList:
- on the Native, use [FlashList](https://shopify.github.io/flash-list/). 
- on the Web, use [`react-virtuoso`'s](https://virtuoso.dev/) list. 
- improve the Tab list's type, make it like native's `ScrollView/FlatList/SectionList` type.

# Blockers 🛑

I want to use `FlashList` to replace `TabRecyclerList`, but I have encountered some obstacles, it has some issues with `Reanimated`'s method, I fought it for a long time but still didn't work.

so I don't want it to block too long, It doesn't affect much at the moment, I will try it again later.

# Test Plan

- check out the new list performance.
- check if the new list looks good on the web & native.


axeldelafosse (approved):
LFG ✨

intergalacticspacehighway (approved):
LGTM 🔥 🔥 🔥

PR #1388, opened 2022-08-12
fix: claiming with magic

intergalacticspacehighway:
# Why
- Sometimes claiming with magic shows the wallet connect popup

# How
- Initialise magic address early instead of fetching in the useWallet hook. This fetch can take a while and if user presses claim before that, it was leading to the above bug

# Test Plan
- Test claiming with magic


packages/app/providers/web3-provider.tsx
```diff
@@ -57,6 +61,12 @@ export function Web3Provider({
       if (magic.rpcProvider && isLoggedIn) {
         //@ts-ignore
         const provider = new EthersWeb3Provider(magic.rpcProvider);
+        provider
+          .getSigner()
+          .getAddress()
+          .then((address) => {
+            setMagicWalletAddress(address);
+          });
```
axeldelafosse, line 69: 💯

axeldelafosse (approved):
LFG ✨

PR #1389, opened 2022-08-12
feat: claim misc

alantoa:
# Why
fixed & done: 
- [SHOW2-1108](https://linear.app/showtime/issue/SHOW2-1103/make-claim-button-full-width-above-nav-bar-like-swm-design)
- [SHOW2-1103](https://linear.app/showtime/issue/SHOW2-1103/make-claim-button-full-width-above-nav-bar-like-swm-design)
- [SHOW2-1102](https://linear.app/showtime/issue/SHOW2-1102/show-description-in-nft-card-instead-of-details-dropdown)
- [SHOW2-1183](https://linear.app/showtime/issue/SHOW2-1083/add-claim-button-to-nft-card)

due we need to move description to feed bottom, will look very bad if the `More` text button is alone in the third row, so we need to move `More` to the end of the second line。

# How

- just adjust some style and layout.
- implement a `MultiClampText` component to support clamping the text:
 on the native, use `onTextLayout` props to calculate the last line.
 on the web, use `binarySearch` to find the last word if in the specified line and do some logic.

this component will be released with List and Tabs in the next PR. 

# Test Plan

check if the Feed/Card UI looks good.

### DIFF
--- a/packages/app/components/card/card.tsx
+++ b/packages/app/components/card/card.tsx
@@ -13,12 +13,14 @@ import { Creator } from "app/components/card/rows/elements/creator";
 import { Owner } from "app/components/card/rows/owner";
 import { Title } from "app/components/card/rows/title";
 import { Social } from "app/components/card/social";
+import { ClaimButton } from "app/components/claim/claim-button";
 import { ErrorBoundary } from "app/components/error-boundary";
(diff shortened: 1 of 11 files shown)


axeldelafosse (approved):
LFG ✨

PR #1391, opened 2022-08-15
Feature/show2 1116 FlashList applying to the Tab

alantoa:
# Why

- we want to get rid of `recyclerlistview`'s API.
- the FlashList has a great performance.
- we want to use `InfiniteScrollList` everywhere.

but the key tab list is not used yet,  so this PR is all for the Tabs list.

# How

- wrap an `Animated.ScrollView` to use in FlashList, 
- encapsulation our `InfiniteScrollList` components to replace old `TabRecyclerList`. 
- fixed some types and bugs.

# Test Plan

- check if the profile/trending screen looks good.
- check their performance.
- check if the list has some bugs.


yarn.lock
```diff
@@ -7962,7 +7962,6 @@ __metadata:
     react-string-replace: ^0.4.4
     react-test-renderer: 18.2.0
     readable-stream: 3.6.0
-    recyclerlistview: ^4.1.3
```
axeldelafosse, line 7965: 🙏

package.json
```diff
@@ -77,7 +77,8 @@
     "@storybook/react-native": "next",
     "@react-navigation/native": "6.0.11",
     "react-native-safe-area-context": "4.3.1",
-    "expo-blur": "~11.2.0"
+    "expo-blur": "~11.2.0",
+    "recyclerlistview": "^4.1.3"
```
axeldelafosse, line 81: Do we need this resolution? Looks like FlashList is using `4.1.2`, did you wanted to upgrade to `4.1.3`?

axeldelafosse (approved):
LFG ✨

package.json
alantoa: yeah, it is needed.
because the `recyclerlistview ` dependency of `Flashlist` is internal, will move it once [this PR](https://github.com/Shopify/flash-list/pull/564) is merged.

intergalacticspacehighway (approved):
LGTM ❇️ Legend 🐐🐐🐐

PR #1393, opened 2022-08-16
chore: upgrade dependencies

intergalacticspacehighway:


alantoa (approved):
LGTM! ✨

apps/expo/package.json
```diff
@@ -132,7 +132,7 @@
     "domain-browser": "^4.22.0",
     "dotenv": "^10.0.0",
     "events": "^3.3.0",
-    "expo": "next",
+    "expo": "~46.0.7",
```
alantoa, line 135: yeah, patch files will have some issues when the lib version has changed if we use `next`. 

for example: `react-native+0.69.3.patch` file, the expo `next` version is used react-native `v0.69.4`, If we forget to update this file, this patch will not be used.

alantoa: so I'm not sure if `expo` `~46.0.7` uses` react-native v0.69.4`.

intergalacticspacehighway: good catch, will update the patch!

> so I'm not sure if expo ~46.0.7 uses react-native v0.69.4.

yeah, it's uses 0.69.4

axeldelafosse (approved):
LFG ✨

PR #1394, opened 2022-08-16
remove expo cli from dev deps

intergalacticspacehighway:
# Why

we added it in dev dependency to keep the version consistent but expo 46 ships with a local CLI, so we don't need it anymore 
https://blog.expo.dev/the-new-expo-cli-f4250d8e3421


axeldelafosse (approved):
LFG ✨

PR #1395, opened 2022-08-16
Update push notifications preferences

axeldelafosse:
# Why

We are about to enable push notifications so we want users to be able to update their preferences. This is the worst design but we don't have any design for it, and we don't have any copywriting. This PR is unblocking the release, let's revisit this screen with a designer

# How

- Added a tab in Settings
- Added some switches to enable / disable notifications

# Test Plan

While being logged in, go to your settings, check the "Push Notifications" tab and play with the switches

### DIFF
--- a/packages/app/components/settings/index.tsx
+++ b/packages/app/components/settings/index.tsx
@@ -5,13 +5,16 @@ import Constants from "expo-constants";
 
 import { useIsDarkMode } from "@showtime-xyz/universal.hooks";
 import { useRouter } from "@showtime-xyz/universal.router";
+import { Switch } from "@showtime-xyz/universal.switch";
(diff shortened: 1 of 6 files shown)


intergalacticspacehighway (approved):
LGTM ❇️

alantoa:
waiting for my PRs, I'll do some improvements to UI.

alantoa:
just found the list `item` and the `TabBar` has some bug, so fixed it.
now I think we need some copywriting, we can push it.
 
Before: 
image

After:  

https://user-images.githubusercontent.com/37520667/184908226-f862e07f-cd27-467c-954f-d46e1b0165cf.mov

alantoa (approved):
LGTM! ✨

PR #1398, opened 2022-08-16
fix: cancel expo-update limit

alantoa:
# Why

just noticed that many App(Telegram), It's also updated multiple times a day, and we iterate fast, if now we limit the number of updates:

- some fixed issues unable to confirm.
- some urgent bugs could not be resolved in time, this will be more annoying than the update prompt.

# How

so I remove some `expo-update` logic and check updates every time, only auto-update when the local time is between 3 am and 5 am.

# Test Plan

check if the expo-update logic looks good.

### DIFF
--- a/packages/app/hooks/use-expo-update.tsx
+++ b/packages/app/hooks/use-expo-update.tsx
@@ -1,11 +1,10 @@
-import { useEffect } from "react";
+import { useEffect, useCallback } from "react";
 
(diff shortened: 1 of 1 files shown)


axeldelafosse (approved):
LFG ✨

intergalacticspacehighway (approved):
LGTM ❇️

PR #1400, opened 2022-08-17
fix: magic test login

intergalacticspacehighway:
# Why
appstore approval. Reviewer not able to test with magic login

# How
Looks like babel env plugin is not able to parse `process.env as any` and convert to the value . Also it has to do with the order of typescript transpile and babel plugin. 

# Test Plan
- Test sign in with test email works

### DIFF
--- a/packages/app/types/environment.d.ts
+++ b/packages/app/types/environment.d.ts
@@ -1,0 +1,9 @@
+export {};
+
+declare global {
+  namespace NodeJS {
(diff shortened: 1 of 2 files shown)


alantoa (approved):
LGTM! ✨
this type warning is very annoying, so I push a `app/types/environment.d.ts` file to your PR, it will fix this warning.

axeldelafosse (approved):
LFG ✨

PR #1402, opened 2022-08-17
fix: claim/drop modal heights on small screen web

intergalacticspacehighway:
# Why
fixes - https://linear.app/showtime/issue/SHOW2-1123/fix-mobile-web-claim-modal-height-also-verify-other-modals-its-too

claim button is invisible in claim explanation screen in small screen devices. It can be a bit confusing.

# How
Increase max height in modal content

# Test Plan
Test drop/claim and other modals


axeldelafosse (approved):
LFG ✨

PR #1403, opened 2022-08-17
feat: imporve feed layout

alantoa:
# Why

- the feed media position is not good, it always leaves a lot of extra space, so I improve it.
- [reported](https://showtime-rq88331.slack.com/archives/C02QD2VDLFK/p1660750694897019?thread_ts=1660683216.803799&cid=C02QD2VDLFK) by Alex.

# How

- now will find the most suitable positioning according to the height of the media, make sure it's centered and reduce white space.
- will remove some character(e.g. \n) when text is collapsed, make sure it fits 2 lines!
- move the media to the header position when text expands and the default visible area is not filled.
- remove the Tap to hide feed's “UI disappearing”, and add support long press to hide it.

# Test Plan

check if the feed looks good.

# Video 

### iOS 

https://user-images.githubusercontent.com/37520667/185206556-097e62d6-5fb8-4954-8bc1-0ffb1c4b53ec.MP4

### Android 
 

https://user-images.githubusercontent.com/37520667/185206835-196e84fd-ce15-439b-bbde-23f3638d16bc.mov

### Web

didn't change.


axeldelafosse (approved):
LFG ✨

PR #1404, opened 2022-08-18
fix(desktop): card style

alantoa:
# Why

image

profile list footer skeleton spacing is wrong.

# How

just adjust the spacing.


intergalacticspacehighway (approved):
LGTM ❇️

PR #1405, opened 2022-08-18
fix: expo update error/crash

alantoa:
# Why

just saw the [Sentry crash record](https://sentry.io/organizations/showtime-l3/issues/3510858122/?project=5860034&query=is%3Aunresolved&sort=date&statsPeriod=14d), I noticed that we not waiting for `fetchUpdateAsync` to download and execute the `reloadAsync`, which I think is a reason for the crash.

# How

waiting for the `Updates.fetchUpdateAsync` is download and execute the update logics.

# Test Plan

check if it crashes when executing the expo-update.


intergalacticspacehighway (approved):
LFG 🙌

axeldelafosse (approved):
LFG ✨

PR #1406, opened 2022-08-18
Feature/show2 1128 search list is breaking

alantoa:
# Why

the search list is breaking.

# How

just adjust the search list height.


axeldelafosse (approved):
LFG ✨

PR #1407, opened 2022-08-18
global mute state (native only)

intergalacticspacehighway:
# Why
- If user unmutes a video, all the other videos should be played unmuted.
- Only implemented on native. On web, I'll pick it after working on [better visibility tracking](https://github.com/showtime-xyz/showtime-frontend/tree/video-feed-improvements). Also there are some more priority items so will cover them first. Web is also a bit tricky as it allows to play more than one video at a time.

# How
- Use context state for mute on native

# Test Plan
- Test unmuting a video, all other videos should play unmuted.


alantoa (approved):
LGTM! ✨

axeldelafosse (approved):
LFG ✨

PR #1412, opened 2022-08-18
Render profiles and NFTs with data from server

axeldelafosse:
# Why

We can use the data from the server to pre-render the profiles and NFTs pages

# How

Pass `fallback` data to `getServerSideProps`: https://swr.vercel.app/docs/with-nextjs#pre-rendering-with-default-data

# Test Plan

On web, check if profiles and NFTs page are server-side rendered


intergalacticspacehighway (approved):
LGTM ❇️

PR #1414, opened 2022-08-19
fix: image not rendering on nft screen

intergalacticspacehighway:
# Why
- Image not rendering on nft screen

# Test Plan
- Test image renders on NFT screen


intergalacticspacehighway:
Merging this now. We need to export height and width prop for each media type. Will work on it.

axeldelafosse (approved):
LFG ✨

PR #1416, opened 2022-08-19
fix: refresh token

intergalacticspacehighway:
# Why
If refresh token is not called once in [3 days](https://showtime-rq88331.slack.com/archives/C02PXGK3V8D/p1660755626385409?thread_ts=1660683361.762629&cid=C02PXGK3V8D), token could get expired. It can happen if user is not killing the app
https://linear.app/showtime/issue/SHOW2-1127/call-refresh-token-once-in-a-day

# How
- Refetch refresh token in six hours if app comes from background

# Test Plan
- Test by reducing the interval and verify if it gets called.


packages/app/providers/auth-provider.tsx
```diff
@@ -26,13 +33,22 @@ interface AuthProviderProps {
...
 }: AuthProviderProps) {
+  const initialRefreshTokenRequestSent = useRef(false);
+  const lastRefreshTokenSuccessTimestamp = useRef<number | null>(null);
+  const appState = useRef(AppState.currentState);
+
   //#region state
   const [authenticationStatus, setAuthenticationStatus] =
-    useState<AuthenticationStatus>("IDLE");
+    useState<AuthenticationStatus>(() =>
+      accessTokenStorage.getAccessToken() ? "AUTHENTICATED" : "IDLE"
```
intergalacticspacehighway, line 50: default state we assume to be optimistic if the token is present

packages/app/providers/auth-provider.tsx
```diff
@@ -137,10 +155,34 @@ export function AuthProvider({
 
   //#region effects
   useEffect(() => {
-    if (authenticationStatus === "IDLE") {
+    if (!initialRefreshTokenRequestSent.current) {
       doRefreshToken();
```
intergalacticspacehighway, line 158: here we don't keep 6-hour check so not storing it in a persistent storage. Thought it would be early optimization.

intergalacticspacehighway:
we can increase the 6 hours interval to a day or two. I think a day should be fine.

axeldelafosse (approved):
LFG ✨

alantoa (approved):
LGTM! ✨

PR #1419, opened 2022-08-22
fix: mute button location

intergalacticspacehighway:
# Why
- Mute button hidden bug reappeared. 

# How
- Moved it to the feed item so it's always placed on top of detail section.

# Test Plan
- Mute button should not be hidden on native

# Aside
- I think we need to unify feed item NFT across web, android and iOS. Make it same as iOS. Will work on it after deciding the priority. This will make it much easier to modify stuff and improve viewing experience on web and android

### DIFF
--- a/apps/expo/app.config.js
+++ b/apps/expo/app.config.js
@@ -108,7 +108,10 @@ export default {
   extra: {
     STAGE: STAGE,
     eas: {
-      projectId: "45cbf5d5-24fe-4aa6-9580-acf540651abd",
+      projectId:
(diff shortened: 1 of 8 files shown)


alantoa (approved):
LGTM! 
oh my bad, I noticed mute button before, but forget to fix it.

intergalacticspacehighway:
> oh my bad, I noticed mute button before, but forget to fix it.

yeah, no worries. It was fragile. We'll have a better solution once we have the same UI on all platforms.

PR #1420, opened 2022-08-22
fix: setting tab failing to load

alantoa:
# Why
[SHOW2-1142 ](https://linear.app/showtime/issue/SHOW2-1142/on-staging-settings-tabs-failing-to-load-tests-aff)
Settings tabs failing to load.

# How

- downgrade reanimated to prev version, because `Switch` component's `Extrapolate` not working, will dig later.
- fix TabView some hooks.
- adjust setting tabs style.

# Test Plan

check if setting tabs works.


intergalacticspacehighway (approved):
LGTM ❇️

PR #1422, opened 2022-08-23
Feature/show2 1144 storybook some components are breaking

alantoa:
# Why

 storybook some components are breaking.

# How

fix components and adjust the storybook layout.

# Test Plan

check if the storybook project is working!


intergalacticspacehighway (approved):
LGTM ❇️

PR #1427, opened 2022-08-24
feat: new drop/claim endpoints. a.k.a 🧈 drop and claim

intergalacticspacehighway:
# Why
- Addresses - https://linear.app/showtime/issue/SHOW2-1211/%5Bbackend-gating%5D-flow-to-select-primary-wallet
- Introduces signatureless drop/claim experience
- Refactor `useWallet` 🤯. Make the behaviour consistent on web and native.
- Now we can do something like below on native and web! (also works with rainbowkit)

```js
await connector.connect();
// no need to do sign in useEffect when `connected` is true and all the hack.
await connector.signMessage();
```
- Refactor `use-add-wallet`. Use same hook for web and native. It was possible because of `useWallet` refactor. 
- Refactor `login with wallet` on native. Removes all the effect logic, makes it much simpler.
- Adds select primary wallet modal during drop and claim.

# How
Implements new backend endpoints for drop/claim

# Test Plan
- Test drop/claim. Also try claiming old drops
- Test add wallet, set primary wallet. Test on native and web.

Screenshot 2022-09-14 at 3 21 41 PM

Screenshot 2022-09-14 at 12 32 02 PM


axeldelafosse (approved):
LFG ✨

PR #1431, opened 2022-08-24
Feature/show2 1041 pull refresh controller

alantoa:
# Why

- try to add the bounce effect on iOS.
- currently, refresh control is not good, we need to use our design-system style.

# How

- use `Spinner` to refresh control.

# Test Plan

check if the Tab looks good.


intergalacticspacehighway (approved):
LGTM ❇️

axeldelafosse (approved):
LFG ✨

Should we remove `react-native-skia`?

alantoa:
@axeldelafosse yeah, I'll remove it.

PR #1432, opened 2022-08-25
fix: tab list has blank area when scrolling up

alantoa:
# Why

Tab list (FlashList) will get blank area when swiping up scrollview, the reason due to it doesn't take contentContainerStyle 's padding into the calculation.

https://user-images.githubusercontent.com/37520667/186619702-9be9beba-7c5f-4e3b-8be6-21742f39af3f.MP4

# How 

### FlashList
FlashList is already handled when having padding or margin cases, so add the `useExternalScrollView` prop to `TabScene`.
if use the `externalScrollView` component to FlashList will close paddingTop to Scene.

### RecyclerListView 
recyclerlistview hasn't handled this case, so I create a [PR](https://github.com/Flipkart/recyclerlistview/pull/741) for recyclerlistview, current we didn't use `RecyclerListView`, so this patch is not required.

# After 

https://user-images.githubusercontent.com/37520667/186625020-a9e390cb-be75-4552-af37-ea187a5093ec.MP4

# Test Plan

check if the profile list is a blank area when scrolling.


axeldelafosse (approved):
LFG ✨

PR #1433, opened 2022-08-25
upgrade some dependencies

intergalacticspacehighway:
# Why
- upgrade infinite scroll list to support better visible item tracking on web.
- Not sure how so many dependencies went out of sync with bit (they all don't seem to have dependency on scroll list), upgraded them too.

# How

# Test Plan


alantoa:
@intergalacticspacehighway  can I merge your branch? I need to move the `InfiniteScrollList` file to `index.ts`.

alantoa:
>Not sure how so many dependencies went out of sync with bit (they all don't seem to have dependency on scroll list), upgraded them too.

due to your `@babel/runtime` version has changed.

image

PR #1435, opened 2022-08-25
Support for hyperlinks and mentions

uneeb123:
# Why

# How

# Test Plan


packages/app/components/messages/message-row.tsx
```diff
@@ -12,12 +13,62 @@ import {
...
+  // next replace all the hyperlinks
+  const result = contentWithMention.map(token => {
+    if (typeof token === 'string') {
+      return (token as string).split(" ").map((word) => {
+        if (!hyperlinkRegex.test(word)) {
+          return word + " ";
+        }
+        const match = [...word.match(hyperlinkRegex)!];
+        return (
+          <Text style={{color: 'blue'}} onPress={() => Linking.openURL(match[0])}>
```
intergalacticspacehighway: here we can use tw="text-blue-500" https://tailwindcss.com/docs/text-color

packages/app/components/messages/message-row.tsx
```diff
@@ -12,13 +13,65 @@ import {
...
 import { colors } from "@showtime-xyz/universal.tailwind";
-import { Text, linkify } from "@showtime-xyz/universal.text";
+import { Text } from "@showtime-xyz/universal.text";
 import { VerificationBadge } from "@showtime-xyz/universal.verification-badge";
 import { View } from "@showtime-xyz/universal.view";
 
 import { Link } from "app/navigation/link";
 import { convertUTCDateToLocalDate, formatAddressShort } from "app/utilities";
 
+import { useRouter } from "design-system/router";
```
alantoa: for unity, need to use `import { useRouter } from "@showtime-xyz/universal.router";` replace it.

intergalacticspacehighway (approved):
LGTM ❇️

PR #1437, opened 2022-08-26
fix: card layout shift 

intergalacticspacehighway:
# Why
fix card layout shift due to claim button

# How
- Add static height if NFT is a drop

# Test Plan
Test NFT card in native and web feed


axeldelafosse (approved):
LFG ✨

alantoa (approved):
LGTM ✨

PR #1439, opened 2022-08-26
Feature/show2 1161 use radix UI select replace old

alantoa:
# Why

old Select is very annoying, it's always covered in places.

# How

- use Radix-UI Select to replace it, and use `Portal` to content, so now will be no more `zIndex` or `covered` issues.

- move some common hooks to `@showtime-xyz/universal.hooks`.
- add `@showtime-xyz/universal.utils` package.

https://user-images.githubusercontent.com/37520667/186896120-1578ab85-aa2d-442f-9dfc-1d5e2e5e8bdb.mov

# Test Plan 

check if the Select is covered.


packages/app/components/profile/profile-tab-filter.tsx
```diff
@@ -1,6 +1,6 @@
 import { useCallback, useContext } from "react";
 
-import { Select } from "@showtime-xyz/universal.select";
+import { Select } from "design-system/select";
```
axeldelafosse, line 3: `import { Select } from "@showtime-xyz/universal.select";`

axeldelafosse (approved):
LFG ✨

intergalacticspacehighway (approved):
LGTM ❇️ drop UI looking slick!

PR #1440, opened 2022-08-26
Lazy load Magic

axeldelafosse:
# Why

Magic is slowing down the web app

# How

- Lazy loaded Magic through `next/script` and CDN
- Removed `web3modal` 
- Update "Add Wallet" flow on web to use `rainbow-kit`

# Test Plan

Run the web app, sign in with Magic and do some web3 actions


intergalacticspacehighway (approved):
LGTM 🔥

PR #1441, opened 2022-08-26
feat: drop ux purge

alantoa:
# Why 

- [the-drop-card-ux-purge](https://linear.app/showtime/issue/SHOW2-1146/the-drop-card-[ux-purge]).
- `Accordion` component is not smooth on native and does not work in some cases.

# How 
- use `Moti` and `Reanimated` to fix the auto-height of the accordion.  
# Test Plan

- check if the drop UX looks good.


intergalacticspacehighway (approved):
LGTM ❇️

packages/design-system/scroll-view/scroll-view.tsx
```diff
@@ -32,13 +32,17 @@ class ScrollViewClass extends Component<ScrollViewProps & { innerRef?: any }> {
...
-    {...props}
-  />
-));
+const ScrollView = forwardRef((props: ScrollViewProps, ref: any) => {
+  const { colorScheme } = useColorScheme();
+  return (
+    <ScrollViewClass
+      //@ts-ignore
+      innerRef={ref}
+      indicatorStyle={colorScheme === "light" ? "black" : "white"}
```
intergalacticspacehighway, line 41: 💯

PR #1442, opened 2022-08-26
fix: magic claim not working first time

intergalacticspacehighway:
# Why

register magic load listener and force update hook

# How
pubsub


alantoa (approved):
LGTM ✨

intergalacticspacehighway:
we can move it to context later

PR #1443, opened 2022-08-26
fix: magic not initialising sometimes and not resetting address between accounts

intergalacticspacehighway:
some leftovers from https://github.com/showtime-xyz/showtime-frontend/pull/1442

### DIFF
--- a/apps/next/src/pages/_app.tsx
+++ b/apps/next/src/pages/_app.tsx
@@ -115,7 +115,7 @@ export default function App({ Component, pageProps, router }: AppProps) {
       </Head>
       <Script
         src="https://cdn.jsdelivr.net/npm/magic-sdk/dist/magic.js"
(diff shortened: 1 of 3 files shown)


apps/next/src/pages/_app.tsx
```diff
@@ -115,7 +115,7 @@ export default function App({ Component, pageProps, router }: AppProps) {
       </Head>
       <Script
         src="https://cdn.jsdelivr.net/npm/magic-sdk/dist/magic.js"
-        strategy="lazyOnload"
+        strategy="afterInteractive"
```
intergalacticspacehighway, line 118: keeping it `afterInteractive` for now. the `isLoggedIn` call from magic is a bit slow and we want it to happen soon after page load. We can switch after more testing!

PR #1445, opened 2022-08-29
Use yarn resolution to force upgrade `bn.js` to 5.2.1

axeldelafosse:
# Why

`bn.js` was imported 4 different times and it's quite heavy. Using yarn resolution to force upgrade to the latest version reduces the bundle size by ~0.20 MB

# How

Use yarn resolution to force upgrade `bn.js` to 5.2.1

# Test Plan

Test the web3 actions on the web app. The native app shouldn't be impacted because we use `react-native-bignumber`

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,7 +79,10 @@
     "react-native-safe-area-context": "4.3.1",
     "expo-blur": "~11.2.0",
     "metro-react-native-babel-preset": "0.71.3",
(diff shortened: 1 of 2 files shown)


intergalacticspacehighway (approved):
LGTM 🔥. Tested sign in, drop, and claim!

alantoa:
LGTM 🔥

alantoa (approved):

PR #1446, opened 2022-08-29
feat: use NativeWind shadow

alantoa:
# Why

`inline-style`'s shadow is old hacks code, now switch to `NativeWind` and need to replace it. 

# How

use `class` CSS to replace it.

# Test Plan

check if the shadow works on web.


axeldelafosse (approved):
LFG ✨

intergalacticspacehighway (approved):
LGTM ❇️

PR #1453, opened 2022-08-30
Show alert when drop or claim limit reached

axeldelafosse:
# Why

We need to show an alert when users reach the drop or claim limit

# How

- Added an alert when claim limit reached
- Updated the copy for the alert when drop limit reached

# Test Plan

Reach the drop and claim limit


intergalacticspacehighway (approved):
LGTM ❇️ 

cc - @uneeb123 this PR adds the limit alerts

alantoa (approved):
LGTM 🔥

PR #1454, opened 2022-08-30
feat: claimed more NFTs

alantoa:
# Why

- NFTs claimed “ more” should be in bold and just redirect to the person’s profile “Owned” section to display the drops

# How

- redirect to the profile Owned tab when clicking the ` more`.

# Test Plan
- check if the profile screen redirect works.


axeldelafosse (approved):
LFG ✨

intergalacticspacehighway (approved):
nice one 💯 LGTM ❇️

PR #1455, opened 2022-08-30
fix: activity section is not scrolling in NFT screen

intergalacticspacehighway:
# Why
fix: activity section is not scrolling in NFT screen
Also did some cleanup for magic loading. Last fix was did a bit hastily.

# Test Plan
- Test magic loads on web and activity section is scrollable


packages/app/providers/magic-provider.web.tsx
```diff
@@ -32,27 +37,28 @@ function useMagic() {
...
+      value={{
+        magic,
+        Magic: isServer
+          ? null
+          : (window as Window & typeof globalThis & { Magic: any })?.Magic,
+      }}
+    >
+      <Script
+        src="https://cdn.jsdelivr.net/npm/magic-sdk/dist/magic.js"
+        strategy="afterInteractive"
```
intergalacticspacehighway: we can change it to `lazyOnLoad`

axeldelafosse (approved):
LFG ✨

PR #1456, opened 2022-08-31
Added modal for one drop per day

uneeb123:
# Why

# How

Checking if allowed to drop and calculating time before next drop. Added a new modal.

Screen Shot 2022-08-30 at 6 56 48 PM
Screen Shot 2022-08-30 at 6 56 45 PM

# Test Plan

### DIFF
--- a/packages/app/components/drop/drop-form.tsx
+++ b/packages/app/components/drop/drop-form.tsx
@@ -172,6 +172,18 @@ export const DropForm = () => {
     );
   }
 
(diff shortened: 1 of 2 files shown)


intergalacticspacehighway (approved):
LGTM ❇️ 
IMO we should not indicate that it's an error, just inform the user about the limit. We can later improve the message with an illustration but this works for now 👍

alantoa (approved):
LGTM ✨

PR #1458, opened 2022-08-31
fix: crash when bad network

alantoa:
# Why

Using the app with a very bad network resulted in a forced logout.

# How

- add a patch file for web3-provider-engine  lib to avoid crashes when bad network.
- if no network connection or auth fails will not log out and use cache data on App.
- use snackbar to show `No internet connection` hints when device does not network.

# Test Plan

- check if it crashes when a bad network is on App.


packages/app/hooks/use-network-connection.tsx
```diff
@@ -0,0 +1,28 @@
...
+  const { bottom } = useSafeAreaInsets();
+  const isForeground = useIsForeground();
+  const snackbar = useSnackbar();
+
+  useEffect(() => {
+    const getNetwork = async () => {
+      const { isConnected } = await Network.getNetworkStateAsync();
+      if (isConnected) return;
+      snackbar?.show({
+        text: `No internet connection`,
```
axeldelafosse, line 20: 💯

patches/eth-block-tracker+4.4.3.patch
```diff
@@ -0,0 +1,24 @@
...
+-          this.emit('error', newErr)
+-        } catch (emitErr) {
+-          console.error(newErr)
+-        }
++        // try {
++        //   this.emit('error', newErr)
++        // } catch (emitErr) {
++        //   console.error(newErr)
++        // }
++        // avoid this.emit('error', newErr) to cause crash on native
```
axeldelafosse, line 19: Nice catch!

axeldelafosse (approved):
LFG ✨

intergalacticspacehighway (approved):
LGTM ❇️

PR #1459, opened 2022-08-31
Lazy loading

axeldelafosse:
# Why

The `First Load JS shared by all` is still pretty high (1 MB) so we need to reduce it. After this PR we are around 800 KB. Still a lot of work to do -- started a discussion here: https://github.com/rainbow-me/rainbowkit/discussions/699. 

# How

- Lazy load WalletConnect provider
- Lazy load Ethers
- Remove unused code

# Test Plan

Run the apps and test the web3 actions


intergalacticspacehighway (approved):
LGTM 🔥 🔥

PR #1460, opened 2022-08-31
fix: rudderstack

intergalacticspacehighway:
# Why
https://showtime-rq88331.slack.com/archives/C02PXGK3V8D/p1661957035679629

# Test Plan
- rudderstack and logout is working.


axeldelafosse (approved):
LFG ✨

packages/app/providers/auth-provider.tsx
```diff
@@ -94,8 +94,8 @@ export function AuthProvider({
     async function logout() {
       const wasUserLoggedIn = loginStorage.getLogin();
       if (wasUserLoggedIn && wasUserLoggedIn.length > 0) {
-        rudder.track("User Logged Out");
-        await rudder?.reset();
+        rudder?.track("User Logged Out");
+        rudder?.reset();
```
intergalacticspacehighway, line 98: removed await here. I think this should be fine.

PR #1462, opened 2022-09-01
fix: rudder stack is null

intergalacticspacehighway:
# Why
Will debug why it is null after merging. rudder stack is null in Karma's browser. https://showtime-rq88331.slack.com/archives/C02PXGK3V8D/p1662046873891879 but it works for me


alantoa (commented):
LGTM ✨

PR #1464, opened 2022-09-02
Lazy load LogRocket

axeldelafosse:
# Why

Continuing our work on lazy loading

# How

`const LogRocket = (await import("app/lib/logrocket")).default;`

# Test Plan

Run the apps and check if LogRocket is loading


intergalacticspacehighway (approved):
LGTM ❇️

PR #1465, opened 2022-09-02
fix: use netinfo replace to expo-network

alantoa:
# Why

`expo-network` does not work, so just use `@react-native-community/netinfo` to replace `expo-network`

# How

- add `useIsOnline` hooks, and show snackbar when `isOnline` is `false`.

# Test Plan

- check if the `useIsOnline` hooks is works.

### DIFF
--- a/packages/app/providers/swr-provider.tsx
+++ b/packages/app/providers/swr-provider.tsx
@@ -72,7 +72,6 @@ export const SWRProvider = ({
           ) {
             return;
           }
-          console.log(error.status, "status");
 
           if (error.status === 404) {
             return;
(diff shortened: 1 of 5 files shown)


packages/app/providers/swr-provider.tsx
```diff
@@ -84,6 +72,7 @@ export const SWRProvider = ({
           ) {
             return;
           }
+          console.log(error.status, "status");
```
intergalacticspacehighway: nit: log

packages/app/providers/swr-provider.tsx
```diff
@@ -99,7 +88,7 @@ export const SWRProvider = ({
           return AppState.currentState === "active";
         },
         isOnline: () => {
-          return isOnline.current;
+          return isOnline;
```
intergalacticspacehighway, line 90: 💯

intergalacticspacehighway (approved):
LGTM ❇️

PR #1466, opened 2022-09-02
fix: accordion auto height

alantoa:
# Why

just quickly fixed the accordion component auto height not working.


axeldelafosse (approved):
LFG ✨

PR #1467, opened 2022-09-02
Use tailwind css spin animation

alantoa:
# Why

CSS animation is better than js animation.

# How

use the tw class replace to the old js animation. 

# Test Plan 

check if the `Spinner` component looks good.


intergalacticspacehighway (approved):
LGTM ❇️

axeldelafosse (approved):
LFG ✨

PR #1469, opened 2022-09-05
Use `nativewind` on native

axeldelafosse:
# Why

We recently started to use `nativewind` on web. This PR adds support for native too

# How

- Fixed the `tailwind.config.js` file for the Expo app
- Stopped using `twrnc` where possible
- Started to replace `tw.style`
- Updated `Pressable` and `PressableScale`
- Added `ColorSchemeProvider` in Universal UI

# Test Plan

Run the apps and check if everything is rendering correctly


packages/design-system/fieldset/index.tsx
```diff
@@ -59,7 +59,7 @@ export function Fieldset(props: FieldsetProps) {
     ? "flex-1 flex-row items-center justify-between"
     : "";
   return (
-    <View tw={`rounded-4 p-4 ${style} ${twProp} ${switchTw}`}>
+    <View tw={`rounded-2xl p-4 ${style} ${twProp} ${switchTw}`}>
```
axeldelafosse, line 62: Oh btw we should prefer using an array here instead of string interpolation

PR #1470, opened 2022-09-05
fix: timezone issue

alantoa:
# Why

we use UTC+0 time zone by default, so I think just adding `Z` for the time string will fix this issue.


axeldelafosse (approved):
LFG ✨

PR #1473, opened 2022-09-06
upgrade next and reanimated swc plugin

intergalacticspacehighway:
# Why
Upgrade NextJS

# Test Plan
- Test webapp is working

# TODO
We'll have to republish some universal components as they have `next` as direct dependencies. Right now solved it with yarn resolutions. We can do it after the nativewind merge


axeldelafosse (approved):
LFG ✨

alantoa (approved):
LGTM! ✨

PR #1474, opened 2022-09-06
fix: scroll restoration on web back/forward press

intergalacticspacehighway:
# Why
- Scroll position is not preserved when user passes back/forward button in browser.

# How
- Uses NextJS's experimental scrollRestoration flag that restores scroll on back/forward press. It requires the page height to be synchronously calculated.
- I added `minHeight` to InfiniteQueryList on web that uses `estimatedItemSize`. We're already using `estimatedItemSize` as it is required by `FlashList`. I think this could be good enough solution for now until it is resolved at Virtuoso.

# Test Plan
- Scroll position is restored on back/forward browser buttons.

https://user-images.githubusercontent.com/23293248/188635476-be5e52a7-e70b-4453-b45b-1ad17fcfc540.mov

https://user-images.githubusercontent.com/23293248/188636482-45b9c879-b707-46a7-860d-e3dbadc80017.mov


intergalacticspacehighway:
will need to release InfiniteQueryList

alantoa:
I just tested it, seems to not work, or is it need to wait for the new List release?

https://user-images.githubusercontent.com/37520667/188635426-24425ad0-f44a-422a-8ccb-d3ae43aa33d0.mov

intergalacticspacehighway:
oh yes, we need to release new list. bit watch should work though!

alantoa (approved):
LGTM! 🔥

axeldelafosse (approved):
LFG ✨

PR #1475, opened 2022-09-06
fix: share drop on twitter crash

intergalacticspacehighway:
# Why

https://showtime-rq88331.slack.com/archives/C02PXGK3V8D/p1662167414652279

# Test Plan
- Tapping sharing on twitter after drop creation shouldn't crash the app


alantoa (approved):
LGTM ✨

PR #1477, opened 2022-09-07
fix: social buttons not clickable on native app

intergalacticspacehighway:
# Why
Snackbar stays mounted, I think something is acting weird with `AnimatePresence`. Will need to repro in isolation. This PR adds a quick fix for now.
https://showtime-rq88331.slack.com/archives/C02PXGK3V8D/p1662427058514019

# How
- Unmount snackbar if not visible

# Test Plan
- Test social buttons are clickable on native.

@alantoa let me know wyt. I can do a new snackbar release later.


axeldelafosse (approved):
LFG ✨

PR #1478, opened 2022-09-07
Feature/show2 1166 fold/expand bio

alantoa:
# Why

Some users' bios are too long, we should fold them if we haven't limited them on the backend. 

image

# How

 
the current web implementation is a bit hacks because the content includes a lot of `\n`, `tags`, and due to I can't format them, so if a `ReactNode` content  I'll use [mask-image](https://developer.mozilla.org/zh-CN/docs/Web/CSS/mask-image) CSS covered overflow part. 

# Test Plan

check if the clamp text works.

# Video

### Native 

https://user-images.githubusercontent.com/37520667/188839840-3ddc6cfa-eb48-4185-9920-544f94fb8d1d.MP4

### Web 

https://user-images.githubusercontent.com/37520667/188842200-5f9a5957-2e16-4136-bdce-9784c227a77b.mov


alantoa:
need to merge the [use-native-wind](https://github.com/showtime-xyz/showtime-frontend/pulls) branch then release the component.

axeldelafosse (approved):
LFG ✨

PR #1479, opened 2022-09-07
fix:  some tw not effect

alantoa:
# Why
some `node_modules` packages component tw class is not working. 

# How

add `@showtime-xyz` packages to `tailwind.config.js`.

# Test Plan

check if the expo app works.


packages/design-system/accordion/index.tsx
```diff
@@ -57,7 +57,7 @@ const Trigger = ({ tw = "", ...props }: ViewProps) => {
   return (
     <RNAccordion.Trigger>
       <View
-        tw={"w-full flex-row items-center justify-between px-4 py-4 " + tw}
+        tw={"w-full flex-row items-center justify-between px-4 py-5 " + tw}
```
axeldelafosse: Nit: prefer using an array here instead of string concatenation

axeldelafosse (approved):
LFG ✨

packages/design-system/accordion/index.tsx
alantoa: yeah, also like array style!

PR #1476, opened 2022-09-06
feat: add renderabsolutebackgroundcontent

Daavidaviid:
# Why

Add the possibility to have a banner component (ex: an image) with zoom effect when doing pull to refresh
Cf this discussion :
https://github.com/showtime-xyz/showtime-frontend/discussions/1471

# How

# Test Plan

PR #1481, opened 2022-09-07
deps: release/upgrade tabs

alantoa:
# Why

just merge @Daavidaviid's [PR](https://github.com/showtime-xyz/showtime-frontend/pull/1476) and release the `collapsible-tab-view` component.


packages/design-system/collapsible-tab-view/types.tsx
```diff
@@ -26,7 +26,7 @@ export enum RefreshTypeEnum {
 export type CollapsibleHeaderProps<T extends Route> = {
   initHeaderHeight?: number;
   renderScrollHeader: (
-    translateYValue: Animated.SharedValue<number>
+    translateYValue?: Animated.SharedValue<number>
```
alantoa, line 29: hey @Daavidaviid!
since the value is not passed on the web, I changed its type to optional to avoid type error.
maybe you need to make a judgment, thanks!

PR #1484, opened 2022-09-08
fix: coinbase wallet login

intergalacticspacehighway:
# Why
coinbase wallet login is failing on prod app.

# How
Not sure about the changes, did console logging and found the response didn't need JSON.parse and was throwing error. Maybe it's needed in different version of coinbase wallet? Will confirm it with @vishnumad 

# Test Plan
- Test sign in with coinbase wallet is working

### DIFF
--- a/packages/app/providers/wallet-mobile-sdk-provider.tsx
+++ b/packages/app/providers/wallet-mobile-sdk-provider.tsx
@@ -87,7 +87,7 @@ export function WalletMobileSDKProvider({
       };
       const [res] = await makeRequest([action]);
       if (res.result) {
-        return JSON.parse(res.result);
+        return res.result;
       } else {
         throw new Error(res.errorMessage ?? "Personal sign failed");
       }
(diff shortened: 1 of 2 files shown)


axeldelafosse (approved):
LFG ✨

alantoa (approved):
LGTM✨

vishnumad:
This is caused by a regression in Coinbase Wallet version 28.7.0. This will be resolved with version 28.8.0 that will roll out next week (September 12). I've added @intergalacticspacehighway to Testflight to verify that the 28.8 update resolves this. Might want to hold off on this fix or wrap the `JSON.parse` in a try catch and fallback to just returning the result, since this change will break once CB Wallet is updated.

packages/app/providers/wallet-mobile-sdk-provider.tsx
```diff
@@ -87,7 +87,7 @@ export function WalletMobileSDKProvider({
       };
       const [res] = await makeRequest([action]);
       if (res.result) {
-        return JSON.parse(res.result);
+        return res.result;
```
vishnumad, line 90: Maybe change to this?
```ts
try {
  return JSON.parse(res.result);
} catch (e) {
  return res.result;
}
```

intergalacticspacehighway:
Confirmed, it's working with 28.8.8 coinbase wallet. Closing. We can do another release on 12!

PR #1486, opened 2022-09-08
Feature/show2 1199 fix some rounded and spacing

alantoa:
# Why

we switch to `nativewind` so broke some style.

# How 

- just fixed some broke style.  

# Test Plan

check if the card/feed/modal style looks good.

# Before 

https://user-images.githubusercontent.com/37520667/189291145-379f0162-5745-4600-9672-01ec4ea5dd13.mov

# Aflter  

https://user-images.githubusercontent.com/37520667/189291362-e3eac2d4-4ad9-42ac-8399-6c88b001d750.mov


packages/app/components/card/rows/elements/ownership.tsx
```diff
@@ -36,30 +34,29 @@ function OwnershipContainer({
...
-              height={14}
-              colorMode={isDarkMode ? "dark" : "light"}
-              radius="round"
-              show={true}
-            />
-          ))}
-      </OwnershipContainer>
-    );
-  }
+  // if (loading) {
```
alantoa, line 41: drop card no longer has this data, if added this will every time adjust the card height.
so if has `Owner` we should show it directly.

alantoa:
It needs to release the `Button` component, so will wait for `nativewind` PR to have merged. https://github.com/showtime-xyz/showtime-frontend/pull/1485

axeldelafosse (approved):
LFG ✨

PR #1487, opened 2022-09-09
feat: override shareanimatedvalue and translateyvalue

Daavidaviid:
# Why

I realized this pr : #1476 was not great 

With this pr it's possible to override : 
- `shareAnimatedValue` with `overridenShareAnimatedValue?: SharedValue<number>` prop
- `translateYValue` with `overridenTranslateYValue?: SharedValue<number>` prop

That way it's easier to integrate this library in a page with a scroll animated header and a banner with zoom effect when doing pull to refresh (cf: example)

Example: 
https://github.com/Daavidaviid/showtime-scrollview-with-zoom-pull-to-refresh

https://user-images.githubusercontent.com/32457364/189254665-e227fb29-d45d-4590-8eae-d33e3d72591c.mov

# Test Plan

### DIFF
--- a/packages/design-system/collapsible-tab-view/hooks/use-refresh-value.tsx
+++ b/packages/design-system/collapsible-tab-view/hooks/use-refresh-value.tsx
@@ -3,19 +3,22 @@ import Animated, {
   useDerivedValue,
 } from "react-native-reanimated";
 
-export const useRefreshDerivedValue = ({
-  refreshHeight,
-  overflowPull,
-  animatedValue,
-  pullExtendedCoefficient,
-}: {
-  refreshHeight: number;
-  overflowPull: number;
-  animatedValue: Animated.SharedValue<number>;
-  pullExtendedCoefficient: number;
-}) => {
+export const useRefreshDerivedValue = (
+  translateYValue: Animated.SharedValue<number>,
+  {
+    refreshHeight,
+    overflowPull,
+    animatedValue,
+    pullExtendedCoefficient,
+  }: {
+    refreshHeight: number;
+    overflowPull: number;
+    animatedValue: Animated.SharedValue<number>;
+    pullExtendedCoefficient: number;
+  }
+) => {
   return useDerivedValue(() => {
-    return interpolate(
+    translateYValue.value = interpolate(
       animatedValue.value,
       [0, refreshHeight + overflowPull, refreshHeight + overflowPull + 1],
       [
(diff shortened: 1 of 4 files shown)


packages/design-system/collapsible-tab-view/hooks/use-refresh-value.tsx
```diff
@@ -3,19 +3,22 @@ import Animated, {
...
-  animatedValue,
-  pullExtendedCoefficient,
-}: {
-  refreshHeight: number;
-  overflowPull: number;
-  animatedValue: Animated.SharedValue<number>;
-  pullExtendedCoefficient: number;
-}) => {
+export const useRefreshDerivedValue = (
+  translateYValue: Animated.SharedValue<number>,
```
alantoa, line 7: I think is better! 💯

alantoa:
this is what I mentioned before, the native bounce effect is not delivered. 😣
but I think it's not a big deal, just not perfect compared to the native implementation. 

https://user-images.githubusercontent.com/37520667/189483325-236c878a-9fd1-4850-be7e-2a0b073a1f78.MP4

alantoa (approved):
@Daavidaviid  thanks for PR, I think these improvements are great, I'll make an example of this later. 
LGTM 🔥🔥🔥

alantoa:
hey @Daavidaviid, just merged your PR and released `v0.0.38`.

Daavidaviid:
> this is what I mentioned before, the native bounce effect is not delivered. 😣 but I think it's not a big deal, just not perfect compared to the native implementation.
> 
>  RPReplay_Final1662812658.MP4

Oh I see.
I have an issue with iPhone simulators on M1 device, the scroll don't have momentum so I didn't realize that
I updated the example for this : 
https://github.com/Daavidaviid/showtime-scrollview-with-zoom-pull-to-refresh/blob/main/src/pages/ScrollWithZoom/components/BelowBanner.tsx

It's not perfect but it's more ok.

https://user-images.githubusercontent.com/32457364/189492921-c7c08b39-4a81-49cd-bbcb-5ecdc3bd03e1.mov


```typescript
  const containerStyle = useAnimatedStyle(() => {
    if (scrollY.value < 0 && translateYValue.value === 0) {
      return {
        transform: [
          {
            scale: interpolate(
              -scrollY.value,
              [0, 300],
              [1, 3],
              Extrapolate.CLAMP
            ),
          },
        ],
      };
    }
    return {
      transform: [
        {
          scale: interpolate(
            translateYValue.value,
            [0, 300],
            [1, 3],
            Extrapolate.CLAMP
          ),
        },
        {
          translateY: interpolate(
            scrollY.value,
            [0, 300],
            [0, -300],
            Extrapolate.EXTEND
          ),
        },
      ],
    };
  });
```

PR #1488, opened 2022-09-09
fix: coinbase wallet warning

intergalacticspacehighway:
# Why
We'll have to add universal link to prevent a warning at coinbase wallet.

IMG_E9EE4E5828F2-1

# Test Plan
- Test login and web3 actions are working with coinbase. This will require app reinstall


axeldelafosse (approved):
LFG ✨

PR #1492, opened 2022-09-12
fix: tab selected state in edit profile

intergalacticspacehighway:
# Why
fix controlled state in edit-profile form

# Test Plan
- Test tab is changed to select website link if link is not filled by the user


axeldelafosse (approved):
LFG ✨

alantoa (approved):
LGTM ✨
@intergalacticspacehighway thanks for fixed it!

PR #1493, opened 2022-09-12
Fix images

axeldelafosse:
# Why

We need to set the `width` and `height` as props

# How

Pass `width` and `height`

# Test Plan

Check the wallet connect screen on mobile for example:

Screenshot 2022-09-12 at 15 30 55


alantoa (approved):
LGTM ✨

PR #1494, opened 2022-09-12
feat: set primary wallet and add wallet prompt

intergalacticspacehighway:
# Why

Addresses - https://linear.app/showtime/issue/SHOW2-1211/%5Bbackend-gating%5D-flow-to-select-primary-wallet

- Refactor `useWallet` 🤯. Make the behaviour consistent on web and native.
- Now we can do something like below on native and web! (also works with rainbowkit)
```js
await connector.connect();
// no need to do sign in useEffect when `connected` is true and all the hack.
await connector.signMessage();
```
- Refactor `use-add-wallet`. Use same hook for web and native. It was possible because of `useWallet` refactor. 
- Refactor `login with wallet` on native. Removes all the effect logic, makes it much simpler.
- Adds select primary wallet modal during drop and claim.

# Test Plan
- Test add wallet, set primary wallet. Test on native and web.

Screenshot 2022-09-14 at 3 21 41 PM

Screenshot 2022-09-14 at 12 32 02 PM

# Todo
- Add E2E tests for the complete flow.


packages/app/components/add-wallet-or-set-primary/add-wallet-or-set-primary.tsx
```diff
@@ -0,0 +1,32 @@
+import * as React from "react";
```
axeldelafosse: Can remove this line

packages/app/components/add-wallet-or-set-primary/add-wallet-or-set-primary.tsx
```diff
@@ -0,0 +1,32 @@
...
+      <View tw="mt-8 mb-10">
+        <Text tw="text-center text-base text-black dark:text-white">
+          {description}
+        </Text>
+      </View>
+      <TextLink
+        href="/settings"
+        tw="rounded-3xl bg-gray-900 py-3 px-4 text-sm font-bold text-white dark:bg-gray-200 dark:text-black"
+      >
+        Select Primary wallet
```
axeldelafosse: Nit: full uppercase `Select Primary Wallet` or sentence case `Select primary wallet`

packages/app/components/settings/settings-wallet-slot.tsx
```diff
@@ -19,39 +16,33 @@ import { SettingSubTitle } from "./settings-subtitle";
...
-      </Text>
-      <Button variant="primary" size="small" onPress={addWallet}>
-        {walletCTA}
-      </Button>
+      <View tw="flex-1">
+        <Text tw="text-xl font-bold text-gray-900 dark:text-white">
+          Your Wallets
+        </Text>
+        <Text tw="pt-4 text-sm text-gray-900 dark:text-white">
+          Your Primary wallet will be the Polygon address that automatically
```
axeldelafosse: Same as last comment: either `Primary Wallet` or `primary wallet`

packages/app/hooks/use-add-wallet.ts
```diff
@@ -1,155 +1,65 @@
...
-          await disconnectMagic();
+          mutate(MY_INFO_ENDPOINT);
 
-          toastMessage = "Address added and will soon appear on your profile";
-        } else {
-          toastMessage = "Address already connected to your profile";
+          // automatically set the primary wallet on add wallet if user doesn't have one
+          if (hasNoPrimaryWallet) {
+            setPrimaryWallet(addedWallet);
+          }
```
axeldelafosse, line 50: 💯

axeldelafosse (approved):
LFG ✨

packages/app/hooks/use-add-wallet.ts
intergalacticspacehighway: Jorge is adding it on [backend](https://github.com/showtime-xyz/stbackend/pull/495) so I guess we can remove this!

alantoa (approved):
LGTM! 🔥🔥🔥

packages/app/hooks/use-add-wallet.ts
jorgeavaldez: i'm adding the primary wallet if they are a magic user or if they only have the one wallet added! we should probably still prompt them to add one otherwise, i'm going to have the web3 endpoints return an error code for this!

intergalacticspacehighway: > we should probably still prompt them to add one otherwise

yeah definitely! if `primary_wallet` is `null`, they'd see the prompt regardless!

> i'm going to have the web3 endpoints return an error code for this!

Perfect, we'd not let them perform the action if `primary_wallet` is `null`. But good to have backend error as well!

intergalacticspacehighway:
Closing in favor of https://github.com/showtime-xyz/showtime-frontend/pull/1427

PR #1495, opened 2022-09-12
Update `Card`

axeldelafosse:
# Why

We don't show the `Owner` anymore because everything is a drop now

# How

Remove `Owner` in `Card` and adjust padding or margin

# Test Plan

Run the web app on desktop and check the design


alantoa (approved):
LGTM 🔥

PR #1498, opened 2022-09-13
Add `PressableHover`

axeldelafosse:
# Why

In a previous PR, I added the hover effect on the `Pressable` directly. This is not ideal because we don't want to have an effect on all the `Pressable`. This PR adds a `PressableHover` component. We need to think about where we want the hover effects and use the new component there

# How

- Created `PressableHover` based on old `Pressable`
- Removed hover effect from `Pressable`

# Test Plan

Run the web app on desktop and check the hover effects


packages/design-system/pressable-hover/index.tsx
```diff
@@ -0,0 +1,24 @@
...
+import { styled } from "@showtime-xyz/universal.tailwind";
+import type { TW } from "@showtime-xyz/universal.tailwind";
+
+export type Props = ComponentProps<typeof ReactNativePressable> & {
+  tw?: TW;
+};
+
+const StyledPressable = styled(ReactNativePressable);
+
+export function Pressable({ tw, ...props }: Props) {
```
intergalacticspacehighway: should we name it `PressableHover`?

intergalacticspacehighway (approved):
LGTM ❇️

packages/design-system/pressable-hover/index.tsx
axeldelafosse: Ah yes, good catch haha

PR #1499, opened 2022-09-13
perf: notification badge use style to hide

alantoa:
# Why

I just dig [badge still exist even reading notifications on native](https://linear.app/showtime/issue/SHOW2-1158/badge-still-exist-even-reading-notifications-on-native) issue, but always can't repro, I think hide/show notification badge use style will be better.


intergalacticspacehighway (approved):
LGTM ❇️

axeldelafosse (approved):
LFG ✨

PR #1500, opened 2022-09-13
fix: media dimension in nft screen

intergalacticspacehighway:
https://showtime-rq88331.slack.com/archives/C02PXGK3V8D/p1663072878660519

### DIFF
--- a/packages/app/components/media/index.tsx
+++ b/packages/app/components/media/index.tsx
@@ -80,8 +80,8 @@ function Media({
             style={sizeStyle}
             data-test-id={Platform.select({ web: "nft-card-media" })}
             blurhash={item?.blurhash}
-            width={size}
-            height={size}
+            width={sizeStyle?.width ?? size}
+            height={sizeStyle?.height ?? size}
             resizeMode={resizeMode}
           />
         </PinchToZoom>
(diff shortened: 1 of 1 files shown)


packages/app/components/media/index.tsx
```diff
@@ -80,8 +80,8 @@ function Media({
             style={sizeStyle}
             data-test-id={Platform.select({ web: "nft-card-media" })}
             blurhash={item?.blurhash}
-            width={size}
-            height={size}
+            width={sizeStyle?.width ?? size}
```
alantoa, line 83: I have added this to 57 lines, is it necessary here? (maybe you missed it.) 😄
image

alantoa: maybe my bad, it seems to miss height, but not sure it if is height cause of cropped media.  🤔

alantoa (approved):
LGTM ✨

axeldelafosse (approved):
LFG ✨

packages/app/components/media/index.tsx
intergalacticspacehighway: yeah, height was causing it. I think on some NFTs it has different height and width!

PR #1501, opened 2022-09-13
Handle error if failed to upload on IPFS

axeldelafosse:
# Why

We don't handle the error properly and the user is left without knowing what's happening

# How

- Catched the error
- Displayed an error message asking to the user to retry

# Test Plan

Throw an error in `uploadMedia` at the `https://api.pinata.cloud/pinning/pinFileToIPFS` call, run the app and create a new drop. You should see the error message


alantoa (approved):
LGTM ✨

PR #1504, opened 2022-09-14
fix: video won't play unless muted

intergalacticspacehighway:
# Why
- Video doesn't autoplay when it is not muted on mobile browsers

> By default, autoplay executes only if the video doesn’t contain an audio track, or if the video element includes the muted attribute.

https://developer.apple.com/documentation/webkit/delivering_video_content_for_safari?language=objc

iOS Safari doesn't autoplay videos if it is not muted. Same can happen on [Android](https://stackoverflow.com/questions/52261078/autoplay-is-not-working-in-chrome-browser-on-android-device). So, we reset the global mute setting on route change for mobile browsers.

# Test Plan
- Test video autoplay works on mobile browsers even if user had unmuted it once.

### DIFF
--- a/packages/app/providers/mute-provider.tsx
+++ b/packages/app/providers/mute-provider.tsx
@@ -4,13 +4,33 @@ import {
   useState,
   Dispatch,
   SetStateAction,
+  useEffect,
 } from "react";
+
+import { useRouter } from "@showtime-xyz/universal.router";
+
+import { usePreviousValue } from "app/hooks/use-previous-value";
+import { isMobileWeb } from "app/utilities";
 
 export const MuteContext = createContext([true, () => {}] as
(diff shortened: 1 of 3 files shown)


packages/app/providers/mute-provider.tsx
```diff
@@ -4,13 +4,33 @@ import {
...
+
+  const prevRouter = usePreviousValue(router);
+
+  useEffect(() => {
+    if (
+      isMobileWeb() &&
+      (prevRouter?.pathname !== router.pathname ||
+        prevRouter?.query !== router.query)
+    ) {
+      setValue(true);
```
alantoa, line 31: I got the trivia! 💯

alantoa (approved):
LGTM 💯

PR #1505, opened 2022-09-14
fix: video not loading in grid on native

intergalacticspacehighway:
We'll have to clean up dimension logic in all media. https://github.com/showtime-xyz/showtime-frontend/blob/staging/packages/design-system/image/image.tsx#L40 This sets default height, width to 0 so it was not rendering the image unless height and width were specified (So StyleSheet.absolute broke). I think we should avoid such defaults.

We should probably unify Image and Video on all platforms and make them as dumb as possible, also make dimension mandatory. I'll experiment after the wallet flow.


axeldelafosse (approved):
LFG ✨

alantoa (approved):
LGTM ✨

PR #1508, opened 2022-09-15
fix: bio does not wrap

alantoa:
# Why
just fix bio doesn't warp line.

# Before 

image 

# After 

image


intergalacticspacehighway (approved):
LGTM ❇️

PR #1509, opened 2022-09-15
Verify phone number

axeldelafosse:
# Why

We need to let users verify their phone numbers

# How

- Added a modal to verify a phone number
- Added a screen in the Settings to manage the phone number
- Added a check when claiming failed to verify the phone number to unlock more claims

# Test Plan

Run the apps and check if you can verify your phone number

Screenshot 2022-09-15 at 12 58 33
Screenshot 2022-09-15 at 12 58 31
Screenshot 2022-09-15 at 12 54 43
Screenshot 2022-09-15 at 12 54 38
Screenshot 2022-09-15 at 12 50 38
Screenshot 2022-09-15 at 12 50 33

### DIFF
--- a/packages/app/components/login/phone-number-picker.tsx
+++ b/packages/app/components/login/phone-number-picker.tsx
@@ -56,6 +56,55 @@ export const PhoneNumberPicker = (props: PhoneNumberPickerProp) => {
       : data;
   }, [search]);
 
+  const leftElement = useMemo(() => {
+    if (Platform.OS === "web") return null;
+
+    return (
+      <PressableScale
+        onPress={() => {
+          setSearch("");
+          setModalVisible(true);
+        }}
+        hitSlop={{ top: 10, bottom: 10, left: 10, right: 10 }}
+        style={{
+          marginTop: Platform.select({
+            android: 7,
+            default: 0,
+          }),
+          height: 28,
+          flexDirection: "row",
+          alignItems: "center",
+          justifyContent: "center",
+        }}
+      >
+        <Text
+          style={{
+            marginTop: Platform.select({
+              android: -4,
+              default: 0,
+            }),
+            marginRight: 1,
+          }}
+        >
+          {selectedCountry?.emoji}
+        </Text>
+        <Text tw="text-base font-semibold text-gray-600 dark:text-gray-400">
+          {selectedCountry?.dial_code}{" "}
+        </Text>
+      </PressableScale>
+    );
+  }, [selectedCountry]);
+
+  const onSubmit = useCallback(
+    (phoneNumber: string) =>
+      props.handleSubmitPhoneNumber(
+        Platform.OS === "web"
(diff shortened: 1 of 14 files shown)


packages/app/components/login/phone-number-picker.tsx
```diff
@@ -79,53 +128,16 @@ export const PhoneNumberPicker = (props: PhoneNumberPickerProp) => {
...
         />
       </Modal>
+
       <LoginInputField
         key="login-phone-number-field"
         validationSchema={phoneNumberValidationSchema}
         label="Phone number"
         placeholder="Enter your phone number"
-        keyboardType="numeric"
+        keyboardType="phone-pad"
```
intergalacticspacehighway, line 137: 💯

packages/app/components/settings/add-email.tsx
```diff
@@ -27,25 +28,26 @@ export const AddEmailModal = () => {
 
   const submitEmail = useCallback(
     async (email: string) => {
+      rudder?.track("Button Clicked", {
+        name: "Login with email",
```
intergalacticspacehighway, line 32: should this be named `Login with email`?

packages/app/components/settings/add-email.tsx
```diff
@@ -27,25 +28,26 @@ export const AddEmailModal = () => {
...
+
       try {
         const did = await magic.auth.loginWithMagicLink({ email });
         if (did) {
           await addEmail(email, did);
           router.pop();
         }
       } catch (error) {
-        console.log("Error:", error);
+        console.error(error);
```
intergalacticspacehighway, line 42: nit: we can use Logger.error(). Right now it doesn't do anything but in future we can avoid logging on prod and send it to sentry.

intergalacticspacehighway (approved):
LGTM ❇️

packages/app/components/settings/add-email.tsx
alantoa: yeah agreed! use  `Logger.error` will be better!

intergalacticspacehighway:
nit: design gap between button and text

IMG_F72FC0B31B37-1

intergalacticspacehighway:
would be cool to initialise phone country [using this](https://docs.expo.dev/versions/latest/sdk/localization/#localizationregion)! (can be done is separate PR)
https://docs.expo.dev/versions/latest/sdk/localization/

packages/app/components/settings/add-email.tsx
axeldelafosse: Nope, want to keep it consistent with the tracking for the login (the idea being to track someone using an email via Magic)

packages/app/components/settings/add-email.tsx
axeldelafosse: OK good idea, let's switch all `console.error` to `Logger.error` in a separate PR

PR #1511, opened 2022-09-15
fix: pressing tab refreshes the page

intergalacticspacehighway:
# Why
- Page is reloading on bottom tab press.

# How
- Remove `PressableHover` from BottomTabs. Hover won't work for mobile devices so we can use a View instead.

# Test Plan
- Tapping Bottom tabs shouldn't reload the page


alantoa (approved):
LGTM ✨

PR #1517, opened 2022-09-19
Feature/show2 1220 deprecate activity on all nft cards

alantoa:
# Why

- `Activity` needs to die.
- we need a `View on OpenSea` menu on the card dropdown.

# How

- remove the `Activity` drop-down from all Card.
- replace the `Claimers` tab with the `Activity` tab on the NFT detail screen.
- added the `View on OpenSea` menu to the NFT card dropdown.
# Test Plan

Check the drop-down and the tab of the NFT detail screen looks good!


packages/app/components/nft-dropdown.tsx
```diff
@@ -96,39 +97,18 @@ function NFTDropdown({
...
   const isShareAPIAvailable = Platform.select({
     default: true,
     web: typeof window !== "undefined" && !!navigator.share,
   });
 
+  const viewOnOpenSea = () => {
+    // Todo: maybe need to use token_id from backend.
+    const token_id = "1";
+    const link = `https://opensea.io/assets/matic/${nft?.contract_address}/${token_id}`;
+    Linking.openURL(link);
```
intergalacticspacehighway, line 109: does it work without `token_id` from backend?

intergalacticspacehighway (approved):
not sure about the opensea link, everything else LGTM ❇️

packages/app/components/nft-dropdown.tsx
alantoa: yeah, the backend's `token_id` always is 0, but OpenSea requires a valid(it will be greater than 0).
maybe no one claim when a user has dropped, so view the first claim NFT on OpenSea I think it works at least.

axeldelafosse (approved):
LFG ✨

Nit: when logged out, you can see the "sign in" button in the Claimers tab 
Screenshot 2022-09-19 at 15 33 11

PR #1518, opened 2022-09-19
fix: follow list not re-rendering

intergalacticspacehighway:
# Why
Fixes - https://linear.app/showtime/issue/SHOW2-1221/debug-follow-button-on-lists

# How
- Follow button was not re-rendering as state was calculated in `renderItem`. Moved state calculation in Follow button

# Test Plan

- Test pressing follow button updates the state to following on native


alantoa (approved):
LGTM ✨

PR #1519, opened 2022-09-19
Fix/select on web

alantoa:
# Why

just noticed that select is breaking on the web, just fix it!

# Before 

https://user-images.githubusercontent.com/37520667/191102162-a4e4b20c-6e9b-4a55-952b-94b99d2faaa6.mov

# After 
https://user-images.githubusercontent.com/37520667/191102174-f7daa253-f6c3-421d-8472-d0fe0353d652.mov


intergalacticspacehighway (approved):
LGTM ❇️

PR #1521, opened 2022-09-20
fix: screens are broken on Android

alantoa:
# Why

- due to the header's layout position being `absolute`, so many screens are broken on Android.
- the `refreshControl` progress view offset is not good on the notification screen.
- `refreshControl` color is weird on light mode.

# How

- handle headers like iOS.
- and add some props to `RefreshControl` on the notification list.

# Before 

https://user-images.githubusercontent.com/37520667/191218840-42f2b66f-b057-48c7-9f17-ddb6a50c31f7.mov

# After 

https://user-images.githubusercontent.com/37520667/191220095-6926a7ec-645f-4c99-a6ab-a702f6187bc3.mov

# Test Plan

check if all screen looks good on Android.

### DIFF
--- a/packages/app/components/drop/drop-form.tsx
+++ b/packages/app/components/drop/drop-form.tsx
@@ -261,7 +261,7 @@ export const DropForm = () => {
 
   return (
     <BottomSheetModalProvider>
-      {Platform.OS === "ios" ? <View style={{ height: headerHeight }} /> : null}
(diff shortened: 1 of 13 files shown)


axeldelafosse (approved):
LFG ✨

intergalacticspacehighway (approved):
LGTM ❇️

PR #1522, opened 2022-09-20
fix: claim limit increase phone verification alert

intergalacticspacehighway:
# Why
- Alert can contain old state buttons, it was partially leading to bug where user might be shown verify phone button. 
- When user is verified (tickmark) they have drop limit to 10, showing increase claim limit alert to these users doesn't make sense. 

# How
- Show claim limit increase alert for users that are not verified and that haven't verified phone number

# Test Plan
- Try a new account and try claiming more than 5 NFTs. Verify adding number increases it. Right now hard to test due to https://showtime-rq88331.slack.com/archives/C02QD3J7DJM/p1663668503893109

### DIFF
--- a/.bitmap
+++ b/.bitmap
@@ -23,7 +23,7 @@
     },
     "alert": {
         "scope": "showtime.universal",
-        "version": "0.0.68",
+        "version": "0.0.69",
         "mainFile": "index.tsx",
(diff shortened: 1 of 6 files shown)


alantoa (approved):
LGTM ✨

axeldelafosse (approved):
LFG ✨

PR #1523, opened 2022-09-20
fix: sometimes header incorrect height on Android

alantoa:
# Why
just noticed that sometimes the header is the incorrect height on the Android home tab screen, looks like is `HomeNavigator`'s `headerRight` prop of reason.

# How

adjust header height and use `null` to replace with `<View/>` when not logged in on home tab `HeaderRight`.

# Before 

https://user-images.githubusercontent.com/37520667/191244819-4e20f338-3152-497d-8d14-aa501617d21e.mov

# After

https://user-images.githubusercontent.com/37520667/191244886-cf1d914d-1683-425f-8d9a-26d9a1ca5544.mov


axeldelafosse (approved):
LFG ✨

PR #1524, opened 2022-09-20
Feature/show2 1207 pressing tab key takes to the next tab

alantoa:
# Why

- [pressing `tab key` takes to the next Tab content](https://linear.app/showtime/issue/SHOW2-1207/pressing-tab-key-takes-to-the-next-tab-content) 
- links input field is not aligned.
image

# How

- set `display: none;` when the tab is not visible on the web. 
- adjust input left element style.

# Test Plan

check if the `edit-profile` screen looks good on Android and Web!

### DIFF
--- a/packages/app/components/edit-profile.tsx
+++ b/packages/app/components/edit-profile.tsx
@@ -4,7 +4,7 @@ import { Platform, useWindowDimensions } from "react-native";
 import { BottomSheetModalProvider } from "@gorhom/bottom-sheet";
 import { yupResolver } from "@hookform/resolvers/yup";
 import { Controller, useForm } from "react-hook-form";
-import { KeyboardAwareScrollViewProps } from "react-native-keyboard-aware-scroll-view";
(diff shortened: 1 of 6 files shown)


packages/app/components/edit-profile.tsx
```diff
@@ -1,10 +1,10 @@
 import React, { useEffect, useMemo, useState, useCallback } from "react";
-import { Platform, useWindowDimensions, ScrollView } from "react-native";
+import { Platform, useWindowDimensions } from "react-native";
 
 import { BottomSheetModalProvider } from "@gorhom/bottom-sheet";
 import { yupResolver } from "@hookform/resolvers/yup";
 import { Controller, useForm } from "react-hook-form";
-import { KeyboardAwareScrollView } from "react-native-keyboard-aware-scroll-view";
+import { KeyboardAwareScrollViewProps } from "react-native-keyboard-aware-scroll-view";
```
axeldelafosse: Nit: `import type`

axeldelafosse (approved):
LFG ✨

PR #1526, opened 2022-09-21
Fix/remove phone number

alantoa:
# Why

remove the phone is not working, need to move new API.

# How

use `v2/wallet/${address}/remove` endpoint replace to old. 
# Test Plan

check if the removed phone number works.


intergalacticspacehighway (approved):
LGTM ❇️

PR #1527, opened 2022-09-21
new drop/claim UX improvements

intergalacticspacehighway:
# Why
- Go back to the claim page after set primary from the claim modal https://linear.app/showtime/issue/SHOW2-1224/go-back-after-set-primary-wallet-flow
- Fix claim happens with magic wallet after phone verification https://showtime-rq88331.slack.com/archives/C02PXGK3V8D/p1663705719814059
- Add export private key for magic wallet

### DIFF
--- a/apps/expo/package.json
+++ b/apps/expo/package.json
@@ -49,7 +49,7 @@
   "dependencies": {
     "@bacons/link-assets": "^1.1.0",
     "@capsizecss/core": "^3.0.0",
-    "@coinbase/wallet-mobile-sdk": "^0.3.1",
+    "@coinbase/wallet-mobile-sdk": "0.3.1",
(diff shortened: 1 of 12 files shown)


apps/expo/package.json
```diff
@@ -49,7 +49,7 @@
   "dependencies": {
     "@bacons/link-assets": "^1.1.0",
     "@capsizecss/core": "^3.0.0",
-    "@coinbase/wallet-mobile-sdk": "^0.3.1",
+    "@coinbase/wallet-mobile-sdk": "0.3.1",
```
intergalacticspacehighway, line 52: pod install is failing on eas, so trying this. I am not sure if this will work though!

alantoa (approved):
LGTM ✨

PR #1532, opened 2022-09-21
fix: get user address if disconnected user tries to claim old drops

intergalacticspacehighway:
We can take better approach than https://github.com/showtime-xyz/showtime-frontend/pull/1530 due to useWallet refactor


alantoa (approved):
LGTM ✨

PR #1535, opened 2022-09-21
redirect to claim or drop page after setting primary

intergalacticspacehighway:
# Test Plan
- Test setting primary wallet from drop or claim modal


alantoa (approved):
LGTM ✨

PR #1537, opened 2022-09-22
Pass `nsfw` flag to `create-gated-edition` endpoint

axeldelafosse:
# Why

We can now send the `nsfw` flag to the new `create-gated-edition` endpoint

# How

Moved the flag from `poll-edition` to `create-gated-edition`

# Test Plan

Create a drop with the `nsfw` flag and check if it is flagged correctly


intergalacticspacehighway (approved):
LGTM ❇️

alantoa (approved):
LGTM ✨

PR #1541, opened 2022-09-22
fix: switching screen shake on Android

alantoa:
# Why

https://showtime-rq88331.slack.com/archives/C02PXGK3V8D/p1663755863469259

# How

`LayoutAnimation` is not expected behavior, so using `enableLayoutAnimations(false);` to disable it can be fixed.
also, now the header occupies the screen on Android, so need to handle it.

# Test Plan

check if all screen looks good on Android.
# Video 

https://user-images.githubusercontent.com/37520667/191785732-d7f7d449-e3a4-422d-a40e-cbe05e298ee0.mov


axeldelafosse (approved):
LFG ✨

PR #1542, opened 2022-09-23
fix: pressing "more" in notifications should open profile's "owned" tab

intergalacticspacehighway:
# Test Plan
Tapping `and 9 more` in the first notification below should open `owned` tab of that profile
IMG_C412FF7F0423-1


alantoa (approved):
LGTM 🔥

PR #1544, opened 2022-09-23
fix: video doesn't stop on unmounting

intergalacticspacehighway:
# Why
Videos were not unloading due to a bug in reanimated layout animations. Thanks to @hirbod https://github.com/expo/expo/issues/17539#issuecomment-1133980754 

# How
- Remove layout animations for now.

# Test Plan
- Test if videos unload on unmounting


alantoa (approved):
LGTM 🔥

hirbod:
FYI: LayoutAnimation is currently being rewritten by SWM, hopefully it will be released soon. (Already 5 months+). AFAIK it should ship together with native shared navigation elements.

oldstud:
@hirbod Hello. And what is the solution? Need to change the video player library?

hirbod:
@oldstud no, don't use LayoutAnimations until it has been rewritten. It does interfere with every library, SVGs, Video, Audio, Camera etc

ScreamZ:
@hirbod Experiencing the issue

having the following code:
```tsx
 <Animated.View style={tw("flex-1")} entering={FadeIn.duration(300)}>
      <Video
        source={{
          uri: callerVideoURL,
        }}
        style={tw("absolute inset-0")}
        shouldPlay
        onPlaybackStatusUpdate={(status) => {
          if (status.isLoaded && status.didJustFinish) onHangup();
        }}
        resizeMode={ResizeMode.COVER}
      />
```

Is there any thread for a reanimated update about this?
Any workaround to pause the video before navigating or something like ?

hirbod:
You need to remove LayoutAnimations for now.
https://github.com/software-mansion/react-native-reanimated/issues/3124

Rewrite
https://github.com/software-mansion/react-native-reanimated/pull/3332

It should be ready soon, but for now, don't use it

ScreamZ:
@hirbod Experienced the issue with camera and video components on Expo with the native stack navigator.

Current workaround I used is

```ts
export function __WORKAROUND__useMediaUnmount() {
  const [displayMedia, setDisplayMedia] = useState(true);
  const navigation = useNavigation();

  useEffect(
    () =>
      navigation.addListener("beforeRemove", (e) => {
        e.preventDefault();

        // Before unmounting
        setDisplayMedia(false);

        setImmediate(() => {
          navigation.dispatch(e.data.action);
        });
      }),
    [navigation],
  );

  return displayMedia;
```

I'll investigate further to see if it's related to the issue here

hirbod:
@ScreamZ while this might be a working workaround for you, you will quickly notice that the beforeRemove "hack", which I also used, is not reliable. It will still happen (timing thing) and memory will still keep growing until your app will eventually crash. When it crash depends on the amount of videos (or even usage of react-native-svg, which is also highly affected by that bug).

PR #1545, opened 2022-09-23
fix: nft desc too height on desktop

alantoa:
# Why

https://linear.app/showtime/issue/SHOW2-1058/nft-description 

# How

- just set `overflow-auto` to fix it.

# Test 

https://user-images.githubusercontent.com/37520667/191978932-f1267bc3-ed45-4af7-925b-8cd371ba339a.mov


intergalacticspacehighway (approved):
LGTM ❇️

PR #1550, opened 2022-09-26
fix: image selection in edit-profile is weird

intergalacticspacehighway:
# Why
Image shows up when tried twice. 

https://user-images.githubusercontent.com/23293248/192231434-2768575c-5105-42b4-bb9f-dcbe3ebffc1f.mov

# How
- Not really sure why but changing the order of set state calls fixes it 😕 

# Test Plan
Test changing image in edit-profile form


axeldelafosse (approved):
LFG ✨

alantoa (approved):
LGTM ✨

PR #1552, opened 2022-09-26
Feature/show2 1237 profile card

alantoa:
# Why

card border is covered and Skeleton is not aligned.

# How 

 adjust styles.

# Before 

image

# After 

image


axeldelafosse (approved):
LFG ✨

PR #1553, opened 2022-09-26
Lock body scroll

axeldelafosse:
# Why

The modal and bottom sheet are not locking the body scroll, which makes a pretty bad experience on mobile web: the user can move the body while messing around with the modal / bottom sheet.

# How

- Updated the React Native for Web CSS reset to be in the global `.css` file
- Created a hook `useLockBodyScroll`
- Upgraded the design system

# Test Plan

Run the mobile web app on iPhone or Android, open a modal and check if you can mess with the body scroll


intergalacticspacehighway (approved):
LGTM 🐐 🐐

PR #1557, opened 2022-09-27
refactor: clamp text on web

alantoa:
# Why

- clamp text some conditions is not working on the web.
- the hooks impletion code is a bit confusing on the web.

# How

- use `Math.ceil` to fix calculation error.
- use the `Class` method to refactor clamp text hooks on web, and the code will be cleaner and easier to understand

 

# Test Plan

check if the clamp text is working in all conditions.

# Before 
 
image 

image

# After  

https://user-images.githubusercontent.com/37520667/192480776-6ed2e1bb-2f9d-4e83-bc10-06c75abe1b69.mov


intergalacticspacehighway (approved):
LGTM 🔥

axeldelafosse (approved):
LFG ✨

PR #1559, opened 2022-09-27
fix: keyboard in forms 

intergalacticspacehighway:
# Why
- Fix keyboard handling in forms on native
- Fix crash on android during claim (cronet crashes when data is undefined during a POST request)

# How
- Adds `react-native-avoid-softinput`. This is working much better than `useKeyboard` hook implementation and requires no custom code!
- Hopefully, we will never have to worry about keyboards anymore (unless something custom e.g. comments). So I have removed a bunch of custom keyboard handling code.

# Test Plan
- Test all forms on all platforms
- This requires a new native build


packages/app/components/login/login-input-field.tsx
```diff
@@ -102,10 +90,10 @@ export function LoginInputField({
         onPress={handleSubmit(handleSubmitData)}
         variant="primary"
         size="regular"
-        tw="mt-6"
+        tw={`mt-6 ${!inputValue ? "opacity-50" : null}`}
```
axeldelafosse, line 93: Nit: when composing styles conditionally, prefer to use an array of styles instead of
using string interpolation

axeldelafosse (approved):
LFG ✨

alantoa (approved):
LFG 🔥 🔥 🔥

PR #1562, opened 2022-09-28
PWA

axeldelafosse:
# Why

Most of our traffic is on web at the moment, so we need to let users install the web app as a PWA and use a service worker to improve the performances.

# How

- Installed `next-pwa`. Used it in the past, it's pretty good! The default runtime cache works for us: https://github.com/shadowwalker/next-pwa/blob/master/cache.js
- Added a redirection to the stores at `/app-store` and `/google-play`
- Added preloading and DNS prefetching for our API and CDN

# Test Plan

Check if the service worker is working correctly with your browser dev tools


apps/next/src/pages/_app.tsx
```diff
@@ -94,13 +94,33 @@ export default function App({ Component, pageProps, router }: AppProps) {
   return (
     <>
       <Head>
+        {process.env.NODE_ENV !== "development" ? (
+          <>
+            <link rel="preconnect" href="//showtimenft.wl.r.appspot.com" />
+            <link rel="dns-prefect" href="//showtimenft.wl.r.appspot.com" />
+            <link rel="preconnect" href="//lh3.googleusercontent.com" />
+            <link rel="dns-prefect" href="//lh3.googleusercontent.com" />
+          </>
```
intergalacticspacehighway, line 103: 💯

alantoa (approved):
just checked it, working fine for me! 
also maybe we can share to [here](https://github.com/shadowwalker/next-pwa/discussions/206) later. 
🔥 🔥 🔥

apps/next/workbox-cache.js
```diff
@@ -0,0 +1,183 @@
...
+      },
+      networkTimeoutSeconds: 10, // fall back to cache if api does not response within 10 seconds
+    },
+  },
+  {
+    urlPattern: ({ url }) => {
+      const isSameOrigin = self.origin === url.origin;
+      if (!isSameOrigin) return false;
+      const pathname = url.pathname;
+      if (pathname.startsWith("/api/")) return false;
```
intergalacticspacehighway, line 155: nit: nice 💯 later we can remove the swr cache on web

intergalacticspacehighway (approved):
LGTM ❇️

PR #1565, opened 2022-09-28
Use expo-application Application ID to register for push notifications between environments

jorgeavaldez:
# Why

Token ID's are specific to environment. When running locally or in staging the bundle identifier needs to match what is configured by the environment in order to register for the appropriate push notifications.

# How

- Use value from `expo-application` since it's configured via env vars appropriately

# Test Plan

- Using local build and local backend


intergalacticspacehighway (approved):
Good catch. LGTM ❇️ 🐐

alantoa (approved):
LGTM 🔥 🔥 🔥

PR #1566, opened 2022-09-29
fix: crash on android when setting primary wallet

intergalacticspacehighway:
# Why
- cronet crashes when data is undefined on PATCH. 

# How
- Pass empty object 

# Test Plan
- Test setting primary works


alantoa (approved):
LGTM ✨

PR #1567, opened 2022-09-29
Stop saying NFTs

axeldelafosse:
# Why

Users shouldn't have to understand what a NFT is. We should try to make it as simple as possible and avoid web3 jargon and technical stuff. When Web 2.0 started, the value proposition wasn't focused on the underlying tech. Web3 shouldn't advertise the tech -- it should focus on what people can do with it (and what's new and exciting about it).

# How

Renamed a bunch of "Free NFTs" copy

# Test Plan

Run the apps

### DIFF
--- a/apps/expo/app.config.js
+++ b/apps/expo/app.config.js
@@ -118,7 +118,7 @@ export default {
       "react-native-vision-camera",
       {
         cameraPermissionText:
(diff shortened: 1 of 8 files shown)


alantoa (approved):
LGTM 🔥

PR #1568, opened 2022-09-29
Feature/show2 943 swipe list

alantoa:
# Why 
 
- `SwipeList` web support, so need to change a lot of files!
- improve feed item on the desktop, now it supports toggle fullscreen.

# How 

on desktop, it's a simple implement, we haven't changed many things.
but on mobile web, we must disable `windowScroll` on the web. 

so:  
- set `body` to `overflow: hidden` on the mobile web.
- refactor many screen layouts on the web, especially the profile page.

next, I want to get rid of `react-native-tab-view` on the web(Setting and Trending are left, also Profile still needs to improve), Its performance is very poor, especially on the dev, I think to do this, the web will be faster! 

about `Tabs` animations I want to use CSS to replace them.

# Test Plan

- check if all screens look good on the web.

# Video 

https://user-images.githubusercontent.com/37520667/193306600-2eefcebe-6f22-4846-84ad-986768d8e45c.mov

https://user-images.githubusercontent.com/37520667/193307450-569196b9-73e1-45c5-b7b4-2916e8fb71c9.MP4


packages/app/components/card/card-skeleton.tsx
```diff
@@ -11,25 +11,28 @@ type CardSkeletonProps = {
...
+  }) => {
     const { colorScheme } = useColorScheme();
     const { width } = useWindowDimensions();
     const isMdWidth = width >= breakpoints["md"];
     if (isMdWidth) {
       return (
         <View
           tw={[
-            "dark:shadow-dark shadow-light mb-4 overflow-hidden rounded-2xl",
+            "dark:shadow-dark shadow-light flexe-1 mb-4 overflow-hidden rounded-2xl",
```
intergalacticspacehighway: `flexe-1` 😂

alantoa: oh damn, 🙀

packages/app/components/feed-item/feed-item.md.tsx
```diff
@@ -83,20 +95,15 @@ export const FeedItemMD = memo<FeedItemProps>(function FeedItemMD({
   const [index, setIndex] = useState(0);
 
   const { width: windowWidth } = useWindowDimensions();
+  const contentWidth = useContentWidth(-240);
```
intergalacticspacehighway: nit: we can assign names to such (-240) constants. I have also done it in many places. Now it feels hard to find why this value

axeldelafosse (approved):
LFG ✨

PR #1569, opened 2022-09-29
chore: remove vision camera

intergalacticspacehighway:
# Why
We have been seeing some crash report. It points to `installFrameProcessorBindings` step. I am not really sure what's causing it. So we remove it for now as we're not using the camera.

Screenshot 2022-09-29 at 5 39 17 PM

# How
- Remove vision camera. We can add it later if needed with some debugging

# Test Plan
Test app works


axeldelafosse (approved):
LFG ✨

PR #1570, opened 2022-09-29
feat: auto initialise country code

intergalacticspacehighway:
# Why
Auto initialise country code in phone number picker

# How
Use expo localization region

# Test Plan
Test native app. Phone picker is currently not on web. cc - @alantoa worked on the web one. expo-localization is not very accurate on web. It shows NZ to me. We could add an API in future that uses IP to get the region 🤷‍♂️ doesn't seem high prio right now.


alantoa (approved):
LGTM 🔥

axeldelafosse (approved):
LFG ✨

intergalacticspacehighway:
it uses navigator.locale to get the region on web. I changed the language settings to English India and it detected proper region on web. This looks like a good enough solution for now on web. cc - @alantoa 

Screenshot 2022-09-29 at 6 49 29 PM

PR #1575, opened 2022-10-03
fix: initialIndex on profile list

alantoa:
# Why

just fix initialIndex on the profile list.


intergalacticspacehighway (approved):
LGTM ❇️

PR #1579, opened 2022-10-03
fix: blocked & tabbar initial & profile router

alantoa:
# Why

 profile router has some issues:
- triggering route deep changes when switching tabs. - [profile-route ](https://linear.app/showtime/issue/SHOW2-874/profile-route)
- [tab bar initial always is first.](https://showtime-rq88331.slack.com/archives/C02PXGK3V8D/p1664811936029879)

# How

- special handling for profile tab bar router change.
- fix tab bar initial index.

# Test Plan

check if the profile router looks pretty good.


axeldelafosse (approved):
LFG ✨

PR #1585, opened 2022-10-06
feat: claims left & various changes

alantoa:
# Why

tasks on Linear:
- [Claims Left on Profile](https://linear.app/showtime/issue/SHOW2-1238/display-of-claims-left-on-profile-tell-people-how-many-claims-they) 
- [Various Profile Changes](https://linear.app/showtime/issue/SHOW2-1246/message-from-alex)

# How

- remove the profile header and add a custom header to the profile, so need to move the profile screen to the new Stack Group, and this Group will be without a default header on the screen.
- add a new screen to display the claim limit description. but I'm running into some obstacles:

1. if I use the default modal screen, I will not be able to change the screen height on iOS. 
2. if I add `BottomSheet` to the component instead of the screen, will have `z-index` issues. the `Modal` content will be covered by the list on the web, and also will be covered by `BottomTab` on native.  

- add a [Chip](https://github.com/showtime-xyz/showtime-frontend/blob/e10240b781459435a38f5028ad2038af31ee462e/packages/design-system/chip/index.tsx) component to design-system, It is used for prompts like `Follows You` copy, If have another about design, we can improve this.

so to solve these issues,  I have added a new Modal Group, this group on iOS it will be the same as on Android, you just need to add `useNativeModal: false` prop to `withModalScreen` HOC.  

# Test Plan 

check if the profile looks good!


packages/app/navigation/root-stack-navigator.tsx
```diff
@@ -102,8 +106,21 @@ export function RootStackNavigator() {
         />
         <Stack.Screen name="drop" component={DropScreen} />
         <Stack.Screen name="claim" component={ClaimScreen} />
+
         <Stack.Screen name="claimers" component={ClaimersScreen} />
       </Stack.Group>
+      <Stack.Group
```
alantoa, line 112: I have added a new Modal Group, see "How 2" for reasons. cc @axeldelafosse  @intergalacticspacehighway

packages/app/screens/claim-limit-explanation.tsx
```diff
@@ -0,0 +1,14 @@
...
+import { ClaimLimitExplanationModal } from "app/components/claim/claim-limit-explanation";
+
+export const ClaimeLimitExplanationScreen = withModalScreen(
+  ClaimLimitExplanationModal,
+  {
+    title: "Claim Limit",
+    matchingPathname: "/claim/claim-limit-explanation",
+    matchingQueryParam: "claimLimitExplanation",
+    snapPoints: [240],
+    useNativeModal: false,
```
alantoa, line 12: set this props the iOS modal will same as Android.

packages/app/components/claim/claim-limit-explanation.tsx
```diff
@@ -0,0 +1,36 @@
...
+  const dailyClaimLimit = userProfile?.data?.daily_claim_limit ?? 5;
+  return (
+    <BottomSheetModalProvider>
+      <View tw="mb-6 justify-center px-4 pt-4">
+        <Text tw="text-sm text-black dark:text-white">
+          {`- Your claims are precious — use them wisely!`}
+        </Text>
+        <View tw="h-4" />
+        <Text tw="text-sm text-black dark:text-white">
+          {hasVerifiedPhoneNumber ? (
```
alantoa, line 22: this copy maybe still need to change, Rize has some comments, waiting for Alex to reply.

packages/app/navigation/root-stack-navigator.tsx
```diff
@@ -45,14 +46,17 @@ export function RootStackNavigator() {
         component={BottomTabNavigator}
         options={{ headerShown: false }}
       />
-
-      {/* Screens accessible in most of the navigators */}
-      <Stack.Group screenOptions={screenOptions({ safeAreaTop, isDark })}>
+      {/* Screens without default header */}
```
alantoa, line 49: if some screens need to without default header, add to this places.
alos next I need to change some screen to this places, I have tasks on Linear.

apps/next/src/pages/_app.tsx
```diff
@@ -19,6 +19,7 @@ import { renderEmptyAnalyticsSnippet } from "app/lib/rudderstack/script";
 import { Sentry } from "app/lib/sentry";
 import { AppProviders } from "app/providers/app-providers";
 import { ClaimScreen } from "app/screens/claim";
+import { ClaimeLimitExplanationScreen } from "app/screens/claim-limit-explanation";
```
axeldelafosse: `ClaimeLimitExplanationScreen` -> `ClaimLimitExplanationScreen`

axeldelafosse (approved):
LFG ✨

PR #1587, opened 2022-10-06
Show Snackbar to display claiming status

axeldelafosse:
# Why

We want to create a seamless UX when claiming drops

# How

- Created `ClaimContext` to use everywhere in the app (TODO: add drop in profile while creating?)
- Added a `Snackbar` to display the claiming status

# Test Plan

Run the apps and claim a drop


packages/app/providers/claim-provider.tsx
```diff
@@ -0,0 +1,119 @@
...
+      }
+
+      await delay(intervalMs);
+    }
+
+    dispatch({ type: "error", error: "polling timed out" });
+  };
+
+  useEffect(() => {
+    if (state.status === "success") {
```
intergalacticspacehighway, line 95: i think instead of using effect we can show it where the status becomes success. Scared of putting stuff in effects

intergalacticspacehighway (commented):
we might need to reset the context state, right?

I think we can reset the context state when someone presses the `claim for free` button and also show alert if a claim is in progress.

[This place](https://github.com/showtime-xyz/showtime-frontend/blob/staging/packages/app/hooks/use-redirect-to-claim-drop.ts#L12) looks good to show that alert

intergalacticspacehighway (approved):
LGTM ❇️

alantoa (approved):
LGTM ✨ 
created a [PR](https://github.com/showtime-xyz/showtime-frontend/pull/1590) fpr fixed Snackbar and Toast compoents.

PR #1590, opened 2022-10-07
Feature/show2 1287 use fullwindowoverlay instead of portal

alantoa:
# Why

this PR is for fix https://github.com/showtime-xyz/showtime-frontend/pull/1587 

# How

use `FullWindowOverlay` to replace Portal way.

# Test Plan

check if the `snackbar` and `toast` components work fine!


packages/design-system/snackbar/snackbar-container.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+export const SnackbarContainer = ({
+  show,
+  snackbar,
+  ...rest
+}: SnackbarProps) => {
+  return (
+    <AnimatePresence>
+      {show && (
+        <FullWindowOverlay
```
intergalacticspacehighway: FullWindowOverlay only works on iOS. On android, it's a normal [View](https://github.com/software-mansion/react-native-screens/blob/3b4d2641b1b757e25a6dabccf78d238d5f31f661/src/index.native.tsx#L465). 🙁

intergalacticspacehighway: We can use `Modal` on android but I think it'll block background View's touches. I am not very sure though, will need to test

alantoa: oh my bad, I don't know `FullWindowOverlay ` only works on iOS, so I will try `Modal` on android.
but at the moment on Android, I think using `View` may be enough, because `ModalScreen` is just a normal JS screen, right? 
anyway, will try once.

intergalacticspacehighway: > I think using View may be enough

yeah, I forgot that it's mounting in the provider which is at the top. Still, it won't show up on top of Modal for e.g. But I think it's fine for now. We don't have that usecase yet.

intergalacticspacehighway (approved):
LGTM ❇️

alantoa:
https://user-images.githubusercontent.com/37520667/194823669-77bc40e6-232c-4bdf-a6d9-faaac4a6ae5e.mov 

at the moment it's working on Android.

packages/app/hooks/use-claim-nft.ts
```diff
@@ -106,59 +109,17 @@ const reducer = (state: State, action: Action): State => {
...
-
-        return;
-      }
-
-      await delay(intervalMs);
-    }
-
-    dispatch({ type: "error", error: "polling timed out" });
-  };
+  const snackbar = useSnackbar();
```
alantoa, line 122: btw I think should get rid of the hooks code, it always warns in`useEffect` hooks depends.
I will use `createRef` method to call it but improve it needs to the next PR.

axeldelafosse (approved):
LFG ✨

PR #1591, opened 2022-10-07
Update Like button position on desktop

axeldelafosse:
# Why

The position is not ideal

# How

Updated the position

# Test Plan

Check if the button is aligned with the title


alantoa (approved):
LGTM ✨

PR #1593, opened 2022-10-10
Ignore unknown notifications

axeldelafosse:
# Why

We want to avoid displaying weird notifications 

# How

Don't display the notification if the frontend don't know the type

# Test Plan

Ask Jorge to send a new notification


alantoa (approved):
LGTM ✨

PR #1594, opened 2022-10-10
Link to the app stores

axeldelafosse:
# Why

We need to tell the web app users that they can install the native app

# How

- On desktop: show a link to the app stores in the home feed below the suggested users
- On mobile: show a download button in the home feed for the first item only

# Test Plan

Run the apps and check if the links are working

### DIFF
--- a/packages/app/components/feed-item/feed-item.web.tsx
+++ b/packages/app/components/feed-item/feed-item.web.tsx
@@ -1,4 +1,4 @@
-import { memo, useState, useMemo } from "react";
+import { memo, useState, useMemo, useContext } from "react";
 import {
   StatusBar,
   StyleProp,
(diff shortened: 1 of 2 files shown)


intergalacticspacehighway (approved):
LGTM ❇️

alantoa (approved):
LGTM ✨ 
but now has two header UI looks are not good, I think will be improved when I finished [SHOW2-1259](https://linear.app/showtime/issue/SHOW2-1259/drop-card-claimed-by-following-and-n-others)

also, can we remove this default banner on Safari? 
otherwise, there will be three areas similar to the banner/header when a user first enters, and removing it will be fixed this [issue](https://showtime-rq88331.slack.com/archives/C02PXGK3V8D/p1665360703822059).



IMG_9262

axeldelafosse:
Yeah +1 for getting rid of the Safari banner

PR #1596, opened 2022-10-10
fix: claim snackbar

alantoa:
# Why

- pressing the `Skip for now` button hasn't reset states when claimed.
- pressing the claim button no feedback when having a claim in progress.

# How

- reset all states and disappear snackbar immediately when pressing `Skip for now`.
- claim button text change to `Claim in progress` when claiming.

# Test Plan

check if the claim + snackbar flow looks good!

# Claim + snackbar flow 

https://user-images.githubusercontent.com/37520667/194920343-b54a8a7b-2dad-45ad-93fa-e3dffd58f5d0.mov

### DIFF
--- a/packages/app/components/claim/claim-form.tsx
+++ b/packages/app/components/claim/claim-form.tsx
@@ -48,9 +48,9 @@ import {
 
 export const ClaimForm = ({ edition }: { edition: CreatorEditionResponse }) => {
   const { rudder } = useRudder();
-  const { state } = useContext(ClaimContext);
+  const { state, dispatch } = useContext(ClaimContext);
 
-  const { claimNFT, onReconnectWallet } = useClaimNFT(
+  const { claimNFT, onReconnectWallet, hideSnackbar } = useClaimNFT(
     edition.creator_airdrop_edition
   );
   const share = useShare();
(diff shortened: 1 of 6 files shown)


packages/app/components/claim/claim-form.tsx
```diff
@@ -124,11 +125,11 @@ export const ClaimForm = ({ edition }: { edition: CreatorEditionResponse }) => {
     );
   }
 
-  if (state.status === "loading" && state.signaturePrompt === false) {
-    router.pop();
+  // if (state.status === "loading" && state.signaturePrompt === false) {
```
alantoa: I'm not sure if need this, if I open it will not be able to enter this screen. 
or should disable the `Claim for Free` button instead of this? any thought? 
cc @axeldelafosse @intergalacticspacehighway

intergalacticspacehighway: yeah, we should disable it for now and show a text maybe `claim in progress`?

alantoa: looks good, will do it.

axeldelafosse (approved):
LFG ✨

PR #1597, opened 2022-10-10
fix: snackbar doesn't update status

intergalacticspacehighway:
# Why
snackbar doesn't update to `claimed` sometimes after success.

# How
we need to call update if snackbar is already [appearing](https://github.com/showtime-xyz/showtime-frontend/blob/staging/packages/design-system/snackbar/index.tsx#L88) or else it doesn't update the snackbar

# Test Plan
Test snackbar updates on success or failure

### DIFF
--- a/packages/app/hooks/use-claim-nft.ts
+++ b/packages/app/hooks/use-claim-nft.ts
@@ -180,16 +180,15 @@ export const useClaimNFT = (edition: IEdition) => {
     try {
       if (edition?.minter_address) {
         dispatch({ type: "loading" });
+        snackbar?.show({
+          text: "Claiming...",
+          iconStatus: "waiting",
+          bottom,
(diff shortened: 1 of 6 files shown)


packages/app/hooks/use-claim-nft.ts
```diff
@@ -253,15 +253,15 @@ export const useClaimNFT = (edition: IEdition) => {
           "Oops. An error occured.",
           "We are currently experiencing a lot of usage. Please try again in one hour!"
         );
-      } else {
```
intergalacticspacehighway, line 256: moving it outside as we'll update the snackbar in any case after an error

alantoa (approved):
LGTM 🙏

alantoa (approved):
LGTM 🙏

PR #1602, opened 2022-10-12
feat: good enough solution for NFT URLs in web SwipeList

intergalacticspacehighway:
# Why
We need to show NFT URL in address bar on web swipelist.

# How
use replaceState from History. Seems it's supported in all the browsers https://developer.mozilla.org/en-US/docs/Web/API/History/replaceState

# Test Plan
Test opening NFT in swipelist on web

https://user-images.githubusercontent.com/23293248/195362611-1518b80a-73ec-4012-9b0a-cb83f2b65def.mov

### DIFF
--- a/packages/app/components/swipe-list.tsx
+++ b/packages/app/components/swipe-list.tsx
@@ -23,6 +23,7 @@ type Props = {
   refresh?: () => void;
   initialScrollIndex?: number;
   bottomPadding?: number;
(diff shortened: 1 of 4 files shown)


axeldelafosse (approved):
LFG ✨

alantoa (approved):
LGTM 🔥

PR #1604, opened 2022-10-12
feat: view likers list & remove liked by

alantoa:
# Why

- card to crowded, so need to remove liked by.
- support to view the liker list.   

# How

- remove liked by from the card.
- add the likers list modal screen. 

# Test Plan

check if the UI looks good and the likers list works.


axeldelafosse (approved):
LFG ✨

PR #1607, opened 2022-10-13
fix: comments UI

alantoa:
# Why

- comments empty component is not vertical centering.
- claimed by avatar border color is not good on dark mode.

# How

adjust the style.

# Before 

image

# After  

image


intergalacticspacehighway (approved):
LGTM ❇️

PR #1608, opened 2022-10-14
feat: update profile tab

alantoa:
# Why 

[SHOW2-1300](https://linear.app/showtime/issue/SHOW2-1300/update-profile-tabs) 

# How

- removed the filter and ITEM, and move the item total to the tab bar. 
- removed page setting on profile setting.

# Test Plan

check if the profile page looks good.

### DIFF
--- a/.bitmap
+++ b/.bitmap
@@ -263,7 +263,7 @@
     },
     "tab-view": {
         "scope": "showtime.universal",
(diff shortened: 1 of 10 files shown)


intergalacticspacehighway (approved):
LGTM ❇️

PR #1610, opened 2022-10-17
Fix/claim left issues

alantoa:
# Why

image 

fix: 
- the claimed by number is incorrect.
- avoid claim left copy invisible when number 0


intergalacticspacehighway (approved):
LGTM ❇️

PR #1611, opened 2022-10-17
fix: avoid Twitter share username appear @@

alantoa:
# Why

 Twitter Share: If the Twitter bio link starts with “@“, share drop/claim shows “@@“ 

# How

via regex to remove username's `@`

# Test Plan 
change username make sure it includes double `@` then share on Twitter.


intergalacticspacehighway (approved):
LGTM ❇️

PR #1612, opened 2022-10-17
fix: profile nickname error

alantoa:
# Why

it's haven't a nickname, name, or primary_wallet when a new wallet is, so will get an error.  

image

# How

Using the `getProfileName` method to get the nickname will be safe. 

# Test Plan

check if use the new wallet login and works.


axeldelafosse (approved):
LFG ✨

PR #1613, opened 2022-10-17
fix: prevent feed refresh after claim

intergalacticspacehighway:
# Why
Noticed the feed refreshes if someone opens the claim modal. This could change the scroll position.

# How
set `revalidateOnFocus` false 

# Test Plan
- Try claiming multiple items, feed should not be reloaded.

# Aside
- We can call the API in the background and let the user know to scroll up for new items if there are x new items.


axeldelafosse (approved):
LFG ✨

alantoa (approved):
LGTM ✨

PR #1618, opened 2022-10-18
fix: console error warning

alantoa:
# Why

now we have a lot of errors or warnings from the console which makes it difficult to debug. we need to fix some, have to make it less.  

https://user-images.githubusercontent.com/37520667/196414885-fb642171-40da-48de-82a0-6a87dd7a3d6b.mov

# How

fix some components warning.

# Test Plan

check if now the console error or warning less.


packages/app/components/feed/feed.md.tsx
```diff
@@ -264,18 +264,18 @@ const SuggestedUsers = () => {
           Get the app
         </Text>
         <View tw="flex flex-row items-center justify-between p-4">
-          <Link
-            tw="font-space-bold dark:text-white"
+          <TextLink
```
alantoa, line 267: if only has single text, we should use the `TextLink` component to avoid a warning from `next.js`.

packages/app/components/header/index.tsx
```diff
@@ -78,9 +77,7 @@ export const HeaderCenter = ({
 }) => {
   return (
     <View tw="flex flex-row">
-      <Link href="/">
-        <ShowtimeTabBarIcon color={isDark ? "black" : "white"} tw="mr-4" />
-      </Link>
+      <ShowtimeTabBarIcon color={isDark ? "black" : "white"} tw="mr-4" />
```
alantoa, line 80: we need to avoid using inside `Link`, otherwise will appear a warning. 
for example: 
```
<Link> <Link>...</Link> </Link>
```

intergalacticspacehighway (approved):
LGTM ❇️

packages/app/components/profile/profile.web.tsx
```diff
@@ -44,6 +43,7 @@ import { NFT } from "app/types";
 import { formatProfileRoutes } from "app/utilities";
 
 import { Spinner } from "design-system/spinner";
+import { Route, TabBarSingle } from "design-system/tab-view";
```
axeldelafosse: Should we use `@showtime-xyz/universal.tab-view` instead?

axeldelafosse (approved):
LFG ✨

PR #1619, opened 2022-10-19
Add check to go back home on native

axeldelafosse:
# Why

If you deep link to a drop or a user, you can get stuck if the back button is not displayed on native

# How

Added a check for the native header ([similar to the check on web](https://github.com/showtime-xyz/showtime-frontend/blob/2f599f3111df35da19b67cf224424fdfa716f644/apps/next/src/pages/_app.tsx#L149))

# Test Plan

Deep link to a drop screen on native and check if you can go back home


alantoa (approved):
LGTM ✨

PR #1620, opened 2022-10-19
Back button for swipe list on web

axeldelafosse:
# Why

We should be able to go back if viewing swipe list on web

# How

Check if `router.pathname === "/list"`

# Test Plan

On the mobile web app, go to Trending or a user profile and press on a drop to display it inside the swipe list. You should be able to go back


alantoa (approved):
LGTM ✨

intergalacticspacehighway (approved):
LGTM ❇️

PR #1621, opened 2022-10-19
Feature/show2 1286 wallets auto nickname user wallets by

alantoa:
# Why

- we wanna auto-nickname user wallets by wallet APP. 
- the setting screen code is not clear, and performance is very slow, especially on the dev web. 
- the header was sorted out before, causing some circular dependencies, need to fix it. 
- has some type error on login logic.
 
# How 

- add the `getWalletDefalutNickname` method to get the wallet default `nickname`, then add this field to the `addWalletToBackend` API when the user adds a new wallet. 
- refactor the setting page. because the setting page haven't a lot of list data, so I  replace all tabs `FlatList` with `ScrollView, and the performance will be better. 
- Separate `HeaderLeft`/`HeaderCenter`/`HeaderRight` components, so will fix import warning.

# Test Plan

### auto-nickname notes:
- on web:`rainbow` wallet depends on `WalletConnect`, so the wallet nickname still is `WalletConnect`.
- on native: everything is normal.  

# Web 

https://user-images.githubusercontent.com/37520667/196670328-72c411cc-48ff-4cb2-be02-d41edd990781.mov

# Native 

https://user-images.githubusercontent.com/37520667/196670410-6967c50c-8a72-4b5b-967f-0507b1adfff2.MP4


intergalacticspacehighway (approved):
LGTM 🔥🔥

axeldelafosse (approved):
LFG ✨

PR #1625, opened 2022-10-20
fix: remove mov file support in create drop

intergalacticspacehighway:
# Why
Issue - https://linear.app/showtime/issue/SHOW2-1317/remove-mov-type-because-opensea-doesnt-support-it
- Remove .mov upload support from frontend

# How?
- Show error instead of using input `accept` prop. This is tricky with cross platform as native image picker doesn't seem to have a similar prop.

# Test Plan
- Try uploading .mov file on drop form

### DIFF
--- a/packages/app/components/drop/drop-form.tsx
+++ b/packages/app/components/drop/drop-form.tsx
@@ -20,7 +20,6 @@ import { useIsDarkMode } from "@showtime-xyz/universal.hooks";
 import { FlipIcon, Image as ImageIcon } from "@showtime-xyz/universal.icon";
 import { Pressable } from "@showtime-xyz/universal.pressable";
 import { useRouter } from "@showtime-xyz/universal.router";
-import { useSafeAreaInsets } from "@showtime-xyz/universal.safe-area";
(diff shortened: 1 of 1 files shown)


axeldelafosse (approved):
LFG ✨

alantoa (approved):
LGTM ✨

packages/app/components/drop/drop-form.tsx
```diff
@@ -281,6 +282,25 @@ export const DropForm = () => {
     );
   }
 
+  const handleFileChange = (file: string | File) => {
+    let extension;
+    // On Native file is a string uri
+    if (typeof file === "string") {
+      extension = file.split(".").pop();
```
alantoa, line 289: if using MIME types to get it will be better because some files haven't the extension.

intergalacticspacehighway: yeah, we use MIME type on web. On native, I don't think there's a way yet. https://docs.expo.dev/versions/latest/sdk/imagepicker/#imageinfo

PR #1626, opened 2022-10-20
Feature/show2 1261 improve trending on web

alantoa:
# Why

- trending performance is not good on the mobile web.   
- trending card missed `1px` spacing on the native.  

# How

on the web: get rid of `react-native-tab-view`, we just simply use a single page to implement. 
on the native: just like the profile screen list to handle it.

# Test Plan 

check if the trending screen is good. 

Web: 

https://user-images.githubusercontent.com/37520667/196973689-3c336312-63e2-44ab-8453-91ac28834b0c.mov 

Native: 

https://user-images.githubusercontent.com/37520667/196973922-bedd3b58-5e5f-4719-8084-b2219e15d763.mov


intergalacticspacehighway (approved):
LGTM 🔥

PR #1627, opened 2022-10-20
fix: linkify URLs and mentions in drop description

intergalacticspacehighway:
# Why
We need to highlight links in drop descriptions

# How
Create a reusable linkify function. Use it in profile bio, comments and description

# Test Plan
Test creating drop with links and mentions


patches/solito+1.0.12.patch
```diff
@@ -0,0 +1,99 @@
+diff --git a/node_modules/solito/build/link/core.js b/node_modules/solito/build/link/core.js
+index 18ef1d9..71c138e 100644
+--- a/node_modules/solito/build/link/core.js
++++ b/node_modules/solito/build/link/core.js
```
intergalacticspacehighway, line 4: PR in solito https://github.com/nandorojo/solito/pull/194

alantoa (approved):
LGTM 🔥 🔥 🔥

PR #1629, opened 2022-10-21
fix: claim fail alert copy

alantoa:
# Why 

alert copy next refill time incorrect when claim failed.


intergalacticspacehighway (approved):
LGTM ❇️

PR #1630, opened 2022-10-21
fix: data pill color in drop form

intergalacticspacehighway:
# Why
We need datapill text to be white. 

# How
Add textTw in datapill

# Test Plan
Test drop form


alantoa (approved):
LGTM ✨

PR #1631, opened 2022-10-21
fix: initial url set in swipelist

intergalacticspacehighway:
# Why
Bug - Swipelist on open sets first drop URL instead of `initialScrollIndex`

# Test Plan
- Tapping drop in feed should set valid drop URL in swipelist screen


alantoa (approved):
LGTM ✨

PR #1638, opened 2022-10-25
fix: select text on swipelist

alantoa:
# Why

we need to make text that can be selected on the SwipeList web.

# How

use `noSwiping` and `noSwipingClass` props to implement it.

# Test Plan

https://user-images.githubusercontent.com/37520667/197824225-07c81c2f-266b-41ab-be74-09a912bf9fda.mov


axeldelafosse (approved):
LFG ✨

intergalacticspacehighway (approved):
LGTM ❇️

PR #1640, opened 2022-10-25
Feature/show2 1096 handle detail not found

alantoa:
# Why

prompt the user to return to the home page when detail is not found.

https://user-images.githubusercontent.com/37520667/197854831-efdea3d2-2b47-41c1-a263-34731cbd7999.mov


intergalacticspacehighway (approved):
LGTM ❇️

PR #1643, opened 2022-10-25
fix: DataPill text color on native

alantoa:
# Why

DataPill component text color does not work on native dark mode.  

# How

use `useIsDarkMode` hooks to fix it first.

# Test Plan 

WX20221026-033533


intergalacticspacehighway (approved):
LGTM ❇️

PR #1645, opened 2022-10-26
fix: alchemy provider endpoint

intergalacticspacehighway:
# Why
We were using alchemy public provider as environment variable was not set and looks like the public provider got killed recently

# How
Add alchemy provider with our own API keys

# Test Plan
We shouldn't see alchemy error in console

# Aside
- We don't need alchemy or any provider now as we don't poll transaction or query blockchain, keeping it here as we might use it to fetch ens names or some other features in the future.


alantoa (approved):
LGTM ✨

PR #1647, opened 2022-10-26
Fix/social styles

alantoa:
# Why

image

social style is not aligned, and sometimes it behaves weirdly because of CSS order on dev and prod. 
context: https://showtime-rq88331.slack.com/archives/C02Q0E4PBD0/p1666777447949999

# How

- avoid props with the same type className, so I use the `Pressable` component to create a new social button to support this.
- add a background color hover effect to the social button.
- adjust just `DropdownMenuTrigger` style.  

# Test Plan 

check if the new social block style looks good on the card.

image
image

### DIFF
--- a/patches/zeego+1.0.0-alpha.20.patch
+++ b/patches/zeego+1.0.0-alpha.20.patch
@@ -1,5 +1,5 @@
 diff --git a/node_modules/zeego/lib/commonjs/menu/web-primitives/index.js b/node_modules/zeego/lib/commonjs/menu/web-primitives/index.js
-index 0928e39..43336de 100644
+index ebdee10..4365982 100644
 --- a/node_modules/zeego/lib/commonjs/menu/web-primitives/index.js
 +++ b/node_modules/zeego/lib/commonjs/menu/web-primitives/index.js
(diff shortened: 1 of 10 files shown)


patches/zeego+1.0.0-alpha.20.patch
```diff
@@ -46,42 +46,29 @@ index f0ae44c..795f9b9 100644
...
    let title = <></>
-@@ -21,11 +21,12 @@ const ItemPrimitive = ({ children, style }: MenuItemProps) => {
+@@ -21,11 +21,11 @@ const ItemPrimitive = ({ children, style }: MenuItemProps) => {
    }
  
    return (
 -    <View style={style}>
 -      {title}
-+    // @ts-ignore
 +    <View style={style} className={className}>
```
intergalacticspacehighway, line 72: is this patch required? I am not sure if passing `className` here would work

intergalacticspacehighway (approved):
LGTM ❇️

patches/zeego+1.0.0-alpha.20.patch
alantoa: this is the previous file, I just update the code line and version, and did not do any changes, so it works. 

image

PR #1648, opened 2022-10-26
feat: spotify play url in drop cards

intergalacticspacehighway:
# Why
Show Play on Spotify button in cards

# Test Plan
- Test `Play on Spotify` appears in Music drops


axeldelafosse (approved):
LFG ✨

PR #1650, opened 2022-10-27
fix: android auth spotify

intergalacticspacehighway:
# Why
Right now auth with android doesn't work very well. Noticed this weird issue on android with spotify auth - https://github.com/spotify/android-sdk/issues/325

# How
Use Webview on android instead of SDK, the below value of userAgent works on google/fb auth. We can write the native module later.  

# Test Plan
Test Spotify auth


alantoa (approved):
LGTM ✨

PR #1651, opened 2022-10-27
nit: fix empty drop screen

intergalacticspacehighway:
# Why
Show empty screen when drop is not present and it's not loading


alantoa (approved):
LGTM ✨ 
btw I already fixed it in my branch, thank you! 🙏

PR #1652, opened 2022-10-27
fix: mention matching regex

intergalacticspacehighway:
# Why
fix mention matching regex (until it breaks again)
https://showtime-rq88331.slack.com/archives/C02PXGK3V8D/p1666742290541129

# How
detect mention only when it starts with `@` 

# Test Plan
- Test drop description, bio and comments.


alantoa (approved):
LGTM ✨

PR #1653, opened 2022-10-27
feat: new swipe list layout on web

alantoa:
# Why

- the feed item UI looks cluttered on the web, we need to improve it. 
- if the content is [Music Drop](https://linear.app/showtime/my-issues/assigned), we need to change the claim button copy. (“Claim for free” into “Save to Collect" )

# How 

on mobile web: 
- reduce Header/BottomBar height, and change the button style on mobile web.
- reduce Header height to media `paddingTop`, make maximize the show of the media. 
- remove the `Download the app` banner, we need to find a better way of adding it later. 
- bring back the `showtime.xyz` deep link, and redirect to `/foryou` URL when swiping once on the home screen. (profile page doesn't support it yet, because I haven't come up with a better URL for the profile, let's try this solution on home page first) 

# Test Plan

check if the new swipe list looks good on mobile web! 

https://user-images.githubusercontent.com/37520667/198255606-76585749-702d-4519-84eb-ee76e6064e37.MP4

https://user-images.githubusercontent.com/37520667/198253588-ce2129e0-7f28-4578-a655-facb348625bf.MP4


axeldelafosse (approved):
LFG ✨

intergalacticspacehighway (approved):
LGTM. Great finding! 🔥

PR #1656, opened 2022-10-27
Feature/show2 1341 dont display the bottom tab bar on nft

alantoa:
# Why

- the NFT detail without the bottom tab bar looks good on native, we should use the same design on mobile web!  

image 

- move the Trending/Notifications title to the header center position, which can save more space, reference to [Figma](https://www.figma.com/file/720aUothRCUZMtMEVHfJV0/Showtime?node-id=1977%3A51355). 

image


axeldelafosse (approved):
LFG ✨

PR #1657, opened 2022-10-28
fix: deeplink initial route

intergalacticspacehighway:
# Why
We need to mount initial route along with deeplinking so back button works as expected in the deeplink screen route

# How
After trying bunch of weird solutions, I found that react navigation has a prop for this exact usecase. `initalRouteName`. This will make sure that home is mounted before deeplink redirection

# Test Plan
- Test deeplinks


axeldelafosse (approved):
LFG ✨

alantoa (approved):
LGTM ✨

PR #1658, opened 2022-10-28
fix: attempt unable to open url

intergalacticspacehighway:
# Why
Not sure what's causing [this](https://showtime-rq88331.slack.com/archives/C02PXGK3V8D/p1666963333401349). Looks like it could [sometimes fail](https://reactnative.dev/docs/linking#canopenurl) on Android. Removing the check as of now, the openUrl also returns a promise which acts mostly similar. 

# Test Plan
Try opening social link URLs


alantoa (approved):
LGTM ✨

PR #1659, opened 2022-10-28
fix: profile inital router

alantoa:
# Why
 
https://linear.app/showtime/issue/SHOW2-1335/profile-router
- profile page is not loaded as expected from the URL `query` when back to the profile screen.

# How

if has URL has a query tab avoid using the default tab from the backend.
 
# Test Plan

• Go to your profile
• Go to the Collected tab
• Click on a drop
• Close the drop
• I would expect to come back to Collected, but instead I am on Drops  

check if fix above repro steps flow.


intergalacticspacehighway (approved):
LGTM ❇️

PR #1664, opened 2022-11-01
fix: layout shift in drop screen

intergalacticspacehighway:
# Why
Layout shift happens when media is loading in drop detail screen. We can add a placeholder skeleton later.

https://user-images.githubusercontent.com/23293248/199182480-6c13c5f9-df75-44a1-8269-03eb5634a949.mov

# How
Add media height/width to the container

# Test Plan
Test drop screen


alantoa (approved):
LGTM ✨

PR #1670, opened 2022-11-02
feat: disconnect spotify

intergalacticspacehighway:
# Why
Add a switch to connect/disconnect Spotify.

Screenshot 2022-11-02 at 9 04 07 PM

# Test Plan
- Test connect/Disconnect works

### DIFF
--- a/packages/app/components/settings/tabs/account.tsx
+++ b/packages/app/components/settings/tabs/account.tsx
@@ -2,6 +2,15 @@ import { useMemo } from "react";
 import { Platform, ScrollView } from "react-native";
 
 import { TabScrollView } from "@showtime-xyz/universal.collapsible-tab-view";
+import { useIsDarkMode } from "@showtime-xyz/universal.hooks";
+import { Spotify } from "@showtime-xyz/universal.icon";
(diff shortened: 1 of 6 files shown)


packages/app/components/settings/tabs/account.tsx
```diff
@@ -29,13 +38,35 @@ export const AccountTab = ({ index = 0 }: AccountTabProps) => {
...
+  const isDark = useIsDarkMode();
+
   return (
     <SettingScrollComponent index={index}>
       <SettingAccountSlotHeader />
       {accountSettings?.length > 0 &&
         accountSettings.map((item) => (
           <AccountSettingItem {...item} key={item.id} />
         ))}
+      <View tw="mb-4 flex-row items-center px-4">
```
alantoa: if make this style is the same as `PushNotificationsTab` will be more consistent, wdyt? 
image

alantoa (approved):
LGTM ✨

packages/app/components/settings/tabs/account.tsx
intergalacticspacehighway: good catch, pushed it. Although not keeping all UPPERCASE 😏

PR #1674, opened 2022-11-03
stop persistent caching on all requests

intergalacticspacehighway:
# Why
- We currently persist all requests! drops, editions, profiles and even comments.
- This PR adds a feature where we can specify what to cache in persistent storage.
- Some redundancy clean up between web and native

# Test
- It should only persist `trending` and `myinfo` API


alantoa (approved):
LGTM ⚡

axeldelafosse (approved):
LFG ✨

PR #1682, opened 2022-11-09
feat: drops left color

alantoa:
# Why 
context: https://showtime-rq88331.slack.com/archives/C02PXGK3V8D/p1667259452594649
color will change when there are <=10 drops left 

# Test Plan

image


intergalacticspacehighway (approved):
LGTM ❇️

PR #1683, opened 2022-11-09
delete profile

intergalacticspacehighway:
https://user-images.githubusercontent.com/23293248/200788154-92f60764-37d7-4a88-8dae-f2e7da8fd9fc.MP4

# Test Plan
- Try deleting account from setting screen. Make sure you don't test with your main account 😆

### DIFF
--- a/packages/app/components/settings/settings-account-slot.tsx
+++ b/packages/app/components/settings/settings-account-slot.tsx
@@ -92,12 +92,14 @@ export const SettingAccountSlotFooter = () => {
               be undone
             </Text>
             <Text tw="py-4 text-base text-gray-900 dark:text-gray-100">
-              Type your <Text tw="font-bold">username</Text> to confirm.
+              Please type{" "}
+              <Text tw="font-bold">{user.user?.data.profile.username}</Text> to
+              confirm.
             </Text>
           </View>
           <Fieldset
             onChangeText={(t) => (username.current = t)}
-            placeholder="@username"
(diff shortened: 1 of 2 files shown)


packages/app/components/settings/settings-account-slot.tsx
```diff
@@ -56,15 +65,55 @@ export const SettingAccountSlotFooter = () => {
...
+              Are you sure you want to delete your account? This action cannot
+              be undone
+            </Text>
+            <Text tw="py-4 text-base text-gray-900 dark:text-gray-100">
+              Type your <Text tw="font-bold">username</Text> to confirm.
+            </Text>
+          </View>
+          <Fieldset
+            onChangeText={(t) => (username.current = t)}
+            placeholder="@username"
```
alantoa: if the placeholder is real `username` will be better, just like GitHub remove the repo, wdyt?

alantoa (approved):
LGTM 🔥

packages/app/components/settings/settings-account-slot.tsx
intergalacticspacehighway: i kept it like that to make it harder. But we can do it the Github way too

alantoa: 🤣 that makes sense!

intergalacticspacehighway: updated

PR #1685, opened 2022-11-09
capture error on sentry if pinata/ipfs upload fails

intergalacticspacehighway:
## Why?
- Some users reported error while uploading files. I am not able to reproduce the issue locally or in prod, but it's possible that pinata could be down
https://showtime-rq88331.slack.com/archives/C02PXGK3V8D/p1667946192395429


alantoa (approved):
LGTM ✨

PR #1686, opened 2022-11-09
feat: dark mode support following system

alantoa:
# Why 

this PR is for: 

- dark mode support following the system. 
- update QR code icons. 

# How

- use the `Appearance` API to implement it. 

# Test Plan 

todo: I need to release some packages in social login PR and then update it, otherwise has some file conflicts. 

- check the DarkMode system and the QR code icon looks good.   
- check if App switching dark mode works when changing the system to dark mode on Web and Native.  

# Video 
https://user-images.githubusercontent.com/37520667/200836203-81ddd111-ca1b-43d2-a67d-fa4282a2a4ab.mov


axeldelafosse (approved):
LFG ✨

PR #1688, opened 2022-11-09
fix: image size incorrectly on web

alantoa:
# Why

fix the image size incorrectly on the web. 

image


intergalacticspacehighway (approved):
LGTM ❇️

PR #1687, opened 2022-11-09
Upgrade to Expo SDK 47

axeldelafosse:
# Why

https://blog.expo.dev/expo-sdk-47-a0f6f5c038af

# How

- Upgraded the packages
- Updated `react-native-quick-crypto` for RN >=0.70 support https://github.com/margelo/react-native-quick-crypto/pull/108
- Updated `@coinbase/wallet-sdk` for RN >=0.70 support  https://github.com/coinbase/wallet-mobile-sdk/pull/119

# Test Plan

Run the native apps

PR #1690, opened 2022-11-10
fix: build error on iOS

alantoa:
# Why 

building has some error on iOS because the latest `expo-av` version rely on `expo-core-module`, also related to `expo` `v47`, it should be upgraded in https://github.com/showtime-xyz/showtime-frontend/pull/1687 PR

# How

- downgrade `expo-av` to `12.0.4`.


axeldelafosse (changes requested):
I think that the lock file is wrong

intergalacticspacehighway (approved):
LGTM ❇️

PR #1693, opened 2022-11-11
feat: dropdown icon support native

alantoa:
# Why

currently, the Dropdown menu icon only supports the web, if make it support native will be better. 

# How 

find many awesome icons that apply it from [Android](https://developer.android.com/reference/kotlin/android/R.drawable)  and [iOS](https://developer.apple.com/sf-symbols/) system icons and then use `zeego/dropdown`'s `androidIconName` and `iosIconName` props support it. 

# Test Plan

# iOS 

 

# Android 

I try to use an Android system drawable to add to the dropdown icon but the UI it's really ugly, so I haven't to continue adding it.


intergalacticspacehighway (approved):
LGTM ❇️

PR #1695, opened 2022-11-14
fix: notification list height on desktop

alantoa:
# Why 
[SHOW2-1347](https://linear.app/showtime/issue/SHOW2-1347/notifications-list-height-incorrect-on-the-web)

fixed the notifications list height incorrectly on the desktop. 

# How

set height `100%` when the `useWindowScroll` prop is disabled. 

# Test Plan

https://user-images.githubusercontent.com/37520667/201590947-72827692-b042-445f-9ddd-1db33af9d3d0.mov


intergalacticspacehighway (approved):
LGTM ❇️

PR #1698, opened 2022-11-15
Feature/show2 1361 profile pictures not displaying

alantoa:
# Why

profile pictures not displaying, because preset the `width` and `height` to the image

# How

remove `width` and `height` props on image styles props. 

# Test Plan  

check if the profile picture is displayed on the profile page. 

image


intergalacticspacehighway (approved):
LGTM ❇️

PR #1699, opened 2022-11-15
feat: add password in QRCode URL for drop owners

intergalacticspacehighway:
# Why
We need to add password in QRCode URL if it's generated by drop owner.

# How
- Append `password` if it's coming in the edition API 

# Test Plan
- Test whether password in appended in URL


alantoa (approved):
LGTM ✨

axeldelafosse (approved):
LFG ✨

PR #1700, opened 2022-11-15
fix: double clicking tolike on web

alantoa:
# Why

always trigger double click to like content when swiping. 

# How

move double-clicking to like logic to `onEnd` from `onFinalize`.

# Before 

https://user-images.githubusercontent.com/37520667/201911961-19615c2d-f06f-44fe-840d-83c58883255e.mov

# After 

https://user-images.githubusercontent.com/37520667/201911975-0b50d133-ec5b-4972-9931-e4de8604b467.mov


axeldelafosse (approved):
LFG ✨

PR #1701, opened 2022-11-15
feat: hide radius when location null

alantoa:
# Why  

for drop form clear and simple: 
- hide `Radius` options when `Location` is null.
- hide `Location` and `Password` data when them is `null`.

# Test Plan

check if the drop form looks good! 

https://user-images.githubusercontent.com/37520667/201939629-87c080e5-0b34-42e5-a365-681676ca633b.mov


axeldelafosse (approved):
LFG ✨

PR #1705, opened 2022-11-16
perf: optimise fetches in grid items

intergalacticspacehighway:
# Why
- App makes too many requests https://showtime-rq88331.slack.com/archives/C02PTPFCZAA/p1668598439477099 (less than 1200 though)
- Right now grid items in mobile screen fetch NFT details and edition API for each item. This makes sense for large screen as we show gift count from edition API and multiple_owner_list from detail API (it can also be optimised, will do in next PR)/

# How
Don't fetch detail/edition for mobile screen grid items

# Test Plan
- Test mobile screen doesn't fetch detail/edition for each item.

### DIFF
--- a/packages/app/components/card/card.tsx
+++ b/packages/app/components/card/card.tsx
@@ -65,27 +65,18 @@ type Props = {
   style?: StyleProp<ViewStyle>;
 };
 
-function Card({
-  nft,
-  numColumns = 1,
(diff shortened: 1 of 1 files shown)


alantoa (approved):
LGTM ✨

axeldelafosse (approved):
LFG ✨

PR #1708, opened 2022-11-17
fix: remove recylyerlistview patch 

alantoa:
# Why  

FlashList has [supported this patch](https://github.com/Shopify/flash-list/pull/626) file function now, not we can remove the patch file! 

# Test Plan

check the Profile/Trending list if has blank when scrolling on Native.


axeldelafosse (approved):
LFG ✨

PR #1709, opened 2022-11-17
Fix/remove react native menu patch

alantoa:
# Why

- on iOS, we are using [react-native-ios-context-menu](https://github.com/dominicstop/react-native-ios-context-menu), so this patch file no longer need. 
- I have created a https://github.com/react-native-menu/menu/pull/474 and has been released, so upgrade the version.


axeldelafosse (approved):
LFG ✨

PR #1710, opened 2022-11-17
fix: removeTags method regex

alantoa:
# Why

report from https://showtime-rq88331.slack.com/archives/C02PXGK3V8D/p1668708931682819, the `removeTags` method regex will remove `_` symbol, just fix this.


axeldelafosse (approved):
LFG ✨

PR #1713, opened 2022-11-18
fix: regex & clean code

alantoa:
# Why

- now we have two `utilities.ts` files, which is a little redundant. 
- `remove-markdown` packages regex some rules don't suit us very well. 
- `linkifyDescription` method @mentions regex is not correct. 

# How

- remove `app/lib/utilities.ts` and move some methods to `app/utilities.ts`. 
- copy the `remove-markdown` file to our lib, and remove some unless rules. 
- fix `@mentions` regex.

# Test Plan

check if the `linkifyDescription` regex all looks good!


intergalacticspacehighway (approved):
LGTM 🔥

PR #1715, opened 2022-11-21
fix: dark mode change

alantoa:
# Why

- dark mode got changed when switching the app 

https://user-images.githubusercontent.com/37520667/203341526-31c283cc-303b-4108-aa02-a5282628a96e.mov

# How

- update `color-scheme` hooks. 
- upgrade `nativewind` version. 

# Test Plan

- change the color theme then switch the app to check if the theme has changed for a while. 

https://user-images.githubusercontent.com/37520667/203342041-59c50d74-1d8e-48af-a6a4-f7f61c9add9b.mov


packages/design-system/extensions/react-native-web/react-native-web.main.runtime.ts
```diff
@@ -155,6 +155,7 @@ export class ReactNativeWebMain {
           "@babel/core": "^7.18.9",
           "@babel/runtime": "^7.18.9",
           "babel-preset-expo": "^9.1.0",
+          "@storybook/react": "6.5.9",
```
alantoa, line 158: I noticed that this package should be moved to `devDependencies`.

axeldelafosse (approved):
LFG ✨

PR #1721, opened 2022-11-24
feat: file upload max size

alantoa:
# Why 

- improve video error when the picked file size is too large.
- clean some old code. 

# How

- when selecting a file that is too large, an error is shown on the form. 

# Test Plan

select a file size of more than 50 MB then upload it.  

https://user-images.githubusercontent.com/37520667/203830055-97bed13e-6708-4216-9265-5747dbada164.mov


packages/app/components/camera/camera-buttons.tsx
```diff
@@ -5,8 +5,8 @@ import { Flip, Close, Check } from "@showtime-xyz/universal.icon";
 import { PressableScale } from "@showtime-xyz/universal.pressable-scale";
 import { View } from "@showtime-xyz/universal.view";
 
-import { FilePickerResolveValue } from "design-system/file-picker";
-import { ImagePickerButton } from "design-system/image-picker";
+import { FilePickerResolveValue } from "app/lib/file-picker";
```
alantoa, line 8: move the `file-picker` and `image-picker` folders to `app/lib/*`, I think It should not be in the ·design-system·

packages/app/components/drop/drop-form.tsx
```diff
@@ -142,7 +144,7 @@ export const DropForm = () => {
   const windowWidth = useWindowDimensions().width;
 
   const [accordionValue, setAccordionValue] = useState("");
-  const { clearStorage } = usePersistDropForm(FORM_DATA_KEY, {
+  const { clearStorage } = usePersistForm(FORM_DATA_KEY, {
```
alantoa: rename `usePersistDropForm` to `usePersistForm`, because it also can use in other forms I think.

axeldelafosse: 💯

packages/app/components/drop/drop-form.tsx
```diff
@@ -315,13 +315,22 @@ export const DropForm = () => {
...
+    const { file, size } = fileObj;
     let extension;
     // On Native file is a string uri
     if (typeof file === "string") {
       extension = file.split(".").pop();
     }
+    if (size && size > MAX_FILE_SIZE) {
+      setError("file", {
+        type: "custom",
+        message: "File size more than 50MB",
```
axeldelafosse: "File size more than 50MB" -> "Oops, this file is too large (>50MB). Please upload a smaller file."

axeldelafosse (approved):
LFG ✨

packages/design-system/alert/index.tsx
```diff
@@ -32,6 +33,10 @@ export const AlertContext = createContext<AlertContextType>({
   alert: () => undefined,
   isMounted: false,
 });
+export let Alert: AlertContextType = {
```
alantoa: I think the `Alert`, `Toast`, and `Snackbar` components should get rid of `hooks` mode, otherwise always show warnings in `useEffect`, `useCallback`, etc deps. 
but need to do in the next PR.

packages/app/components/drop/drop-form.tsx
alantoa: thanks for this copy that's will better! (just not sure if is too long, because the preview image is too small on mobile screen)

PR #1722, opened 2022-11-24
fix: image clip path

alantoa:
# Why

just noticed that the image has a `0.5px` path bug on some browsers. 

image

# How

add a media query on `img[loading="lazy"]` CSS following docs: https://nextjs.org/docs/api-reference/next/image#known-browser-bugs

# After 

image


intergalacticspacehighway (approved):
Good catch 🙌

PR #1724, opened 2022-11-25
feat: profile suspension

alantoa:
# Why 

we need to show a message when the user account suspension. 
FYI: [context](https://showtime-rq88331.slack.com/archives/C02PTNZMEPQ/p1664899592826549).

# How 

- handle profile endpoint 403 error, and create a `ProfileError` component. 

# Test Plan 

access a suspension account and check the error message. 

example


intergalacticspacehighway (approved):
LGTM 🔥

packages/app/components/profile/profile-error.tsx
```diff
@@ -0,0 +1,139 @@
...
+        <LinkComponent href="mailto:help@showtime.xyz">
+          <Button
+            {...Platform.select({
+              web: {},
+              default: {
+                onPress: async () => {
+                  const isCanOpen = await Linking.canOpenURL(
+                    "mailto:help@showtime.xyz"
+                  );
+                  await Clipboard.setStringAsync("help@showtime.xyz");
```
intergalacticspacehighway: need to remove this from here? as we also add the copy in the `else`

alantoa: yeah, my bad for the test on local, will update it.

PR #1725, opened 2022-11-28
nit: account screen ui

intergalacticspacehighway:
# Why
UI changes based on feedback - https://showtime-rq88331.slack.com/archives/C02S5PQ1XJT/p1669590404295289

### DIFF
--- a/packages/app/components/settings/setting.web.tsx
+++ b/packages/app/components/settings/setting.web.tsx
@@ -1,12 +1,10 @@
-import { useEffect, useState } from "react";
+import { useState } from "react";
 
-import { useRouter } from "@showtime-xyz/universal.router";
 import { View } from "@showtime-xyz/universal.view";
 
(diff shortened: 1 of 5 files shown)


alantoa (approved):
LGTM ✨

packages/app/components/settings/setting.web.tsx
```diff
@@ -1,33 +1,22 @@
...
 
 const SettingsTabs = () => {
-  const { isAuthenticated } = useUser();
-  const router = useRouter();
   const bottomHeight = usePlatformBottomHeight();
   const [editingWallet, setEditingWallet] = useState<
     WalletAddressesV2 | undefined
   >(undefined);
 
-  useEffect(() => {
```
intergalacticspacehighway, line 24: we should avoid using effect for routing, leads to weird bugs sometimes (try connecting google from settings and then logging out). Route change can be done in the handler or use some special flag in useEffect if not possible in handler

PR #1727, opened 2022-11-28
fix: remove unfollow user in self-view

alantoa:
# Why 

removed the “Unfollow User” drop-down menu if NFT is on self-view.


axeldelafosse (approved):
LFG ✨

PR #1729, opened 2022-11-29
Fix/next image border on safari

alantoa:
# Why

Safari v16 is supported [resolution media query](https://developer.apple.com/documentation/safari-release-notes/safari-16-release-notes), so before `@media not all and (min-resolution: 0.001dpcm)` query is not working.

# How  

add specifically media query conditions for safari `v16`. 

# Test Plan

- check if the image border exists on Safari. 
image
- check if lose the 1px `path` on other browsers. 
image


axeldelafosse (approved):
LFG ✨

PR #1732, opened 2022-11-30
Feature/show2 1371 improve qr code drop

alantoa:
# Why

- QR code drop if including media and the creator will be better. 
- QR code images should be shared on Twitter and Instagram or other social Apps. 

# How

- on native: use `react-native-view-shot` and `react-native-share` to implement it. 
- on the web: use `dom-to-image` to save photos. 

# Test Plan 

I have tested `save to photo` and share to `Twitter` and `Instagram` is working on `Android`  and `iOS`. 

[Video on iOS](https://showtime-rq88331.slack.com/archives/C02Q0EY6RMH/p1669830844176689)


packages/app/lib/qr-code/index.tsx
```diff
@@ -0,0 +1,168 @@
...
+          preserveAspectRatio="xMidYMid slice"
+          href={logo}
+          clipPath="url(#clip-logo)"
+        />
+      </G>
+    </G>
+  );
+};
+
+export const ReactQRCode = ({
```
alantoa, line 79: it forks from [react-native-qrcode-svg](https://github.com/awesomejerry/react-native-qrcode-svg) lib because I wanna try to custom QR code and make it look more beautiful. (but need to next time try)
also, it uses group `path` on the web, which will reduce the number of `dom` nodes even more, so I think it performs better than `react-qr-code`.

intergalacticspacehighway (approved):
LGTM 🔥🔥🔥

axeldelafosse (approved):
LFG ✨

PR #1735, opened 2022-12-04
fix: correct IG + Twitter keys in QR code modal

scottarnold:
# Why

Quick little fix to address an issue of misaligned keys in the QR Code Modal twitter/ instagram app sniffing.


alantoa (approved):
hey, @scottarnold thank you very much for finding this issue!

PR #1739, opened 2022-12-06
fix: pwd&location drop exception flow

alantoa:
# Why 

- if take an incorrect password or location is still closed on the claim modal, and no prompt to the user, which will cause the user don't know what's happened. 
- the console has some annoying error logs that need to clear up. 

# How
- move the `closeModal` method to the `claimNFT` hooks, check if the gated passed, and then close it, avoid haven't any check to close the modal. 

# Test Plan 

taking some incorrect values for password/location Drops and the check if the prompt is coming. 

https://user-images.githubusercontent.com/37520667/205951527-45454f78-283c-4bcc-ab9f-b1b4bf9aa1ee.mov


packages/app/hooks/use-claim-nft.ts
```diff
@@ -270,28 +273,27 @@ export const useClaimNFT = (edition: IEdition) => {
...
         url:
           "/v1/creator-airdrops/mint-gated-edition/" + edition.contract_address,
         method: "POST",
         data: {
           password: password !== "" ? password : undefined,
-          location: {
-            latitude: location?.coords?.latitude,
-            longitude: location?.coords?.longitude,
-          },
+          location: location?.coords
```
alantoa, line 288: this is for avoiding the response `500` error on the backend, if set as `undefined`, will back the `400` error.

intergalacticspacehighway (approved):
LGTM ❇️

packages/app/components/claim/claim-form.tsx
```diff
@@ -401,16 +405,8 @@ export const ClaimForm = ({
...
-                (edition.gating_type === "multi" && !location)
-              }
-              tw={
-                state.status === "loading" ||
-                (edition.gating_type === "multi" && !location)
-                  ? "opacity-[0.45]"
-                  : ""
-              }
+              disabled={isDisableButton}
+              tw={isDisableButton ? "opacity-[0.45]" : ""}
```
alantoa, line 409: IMO reducing opacity should be an inner button feature when `disabled`  instead of this, I'm happy to do this.

PR #1740, opened 2022-12-07
fix: location permissions prompt

alantoa:
# Why 

If denying location permission accidentally then will no longer be able to continue it. 

# How

use the `Alert` component to prompt the user to open the location service when location permission is disabled.

# Test Plan 

https://user-images.githubusercontent.com/37520667/206108594-b3a50fab-ef41-46c5-b94c-cec92c8d098f.mov


intergalacticspacehighway (approved):
LGTM ❇️

PR #1741, opened 2022-12-07
fix: date time picker on Chromatic

alantoa:
# Why

just ignore the `DateTimePicker` component UI test on Chromatic, otherwise will UI test every PR. 

image


intergalacticspacehighway (approved):
LGTM ❇️. Thanks for fixing this! 🙏

PR #1743, opened 2022-12-08
perf: imporve keyboard on iOS

alantoa:
# Why

I found [this PR](https://github.com/facebook/react-native/pull/31402) has improved the keyboard on iOS, we can remove dirty lib to work around this. 
 at the same time maybe can fix [this issue](https://linear.app/showtime/issue/SHOW2-1320/screen-sometimes-freezes-cause-by-comments-box).

# How 

on iOS, use the `automaticallyAdjustKeyboardInsets` prop + `InputAccessoryView` to handle the keyboard, it's better and smooth now. 
on other platforms, not any changes. 
 
# Test Plan

check if the keyboard works well and smoothly now. 

https://user-images.githubusercontent.com/37520667/206409135-5f7b2d9a-395a-4a7d-8762-062a124d3812.mov


intergalacticspacehighway (approved):
LGTM ❇️

PR #1745, opened 2022-12-12
feat: custom QR code style

alantoa:
# Why 

old QR code It doesn't look pretty cool, want to improve it and make it looks beautiful. 

image

# How

handle QR code `matrix` and use `SVG` to piece together graphics. 

# Test Plan 

check the accurate recognition rate of the QR code if good. 
image


intergalacticspacehighway (approved):
This is so cool 😍 😍 😍

PR #1746, opened 2022-12-13
fix: text capitalize

alantoa:
# Why

just fix the text to capitalize only the first letter on the push notification setting screen. 

image


intergalacticspacehighway (approved):
LGTM ❇️

PR #1749, opened 2022-12-14
fix: replace modals with stack screens for user list components

intergalacticspacehighway:
# Why
Modal is not very suitable for UI where we have user lists as discussed here - https://showtime-rq88331.slack.com/archives/C02PXGK3V8D/p1670920369295469

# How
- Replace modal with regular screens on native

# Test Plan
Test followers, following, collectors and comments screens on all platforms.

### DIFF
--- a/packages/app/components/collectors-modal.tsx
+++ b/packages/app/components/collectors-modal.tsx
@@ -1,9 +1,7 @@
 import { Suspense } from "react";
-import { Platform } from "react-native";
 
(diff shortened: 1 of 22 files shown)


alantoa (approved):
LGTM 🔥 🔥 🔥 

just tested it only has a small issue on header title in dark mode, I just fixed and pushed code.  
CleanShot 2022-12-15 at 01 20 50

PR #1757, opened 2022-12-21
fix: @next/font bold

alantoa:
# Why

- font bold is not working when we switch to `@next/font`.  

# How

- set the `Space_Grotesk` font weight to `700`. 
 - remove `next-font` because we didn't use it for now. 

# Test Plan
- check if the font looks good on the website.


intergalacticspacehighway (approved):
LGTM ❇️

PR #1763, opened 2023-01-10
feat: switch to expo-image

alantoa:
# Why

- we want to switch to [expo-image](https://twitter.com/tsapeta/status/1605240759342739456), the benefit for us:
  - improve performance, especially since most of our content is images
  - get rid of blurhash hacks code 
- fix the tooltip component before not working on the card title. 

# How

- remove `react-native-fast-image` and install `expo-image` to our `design-system`, and clean some code. 

# Test Plan

- reinstall the app on the device and check if all image is works.
> ⚠️note: since we used `FastImage` before and it uses `SDWebImage` on iOS and `Glide` on Android, its version is different from `ExpoImage`, maybe you need to run `pod update` some packages, I think the best way is to remove the `ios` folder and reinstall it.


intergalacticspacehighway (approved):
LGTM 🔥🔥🔥

PR #1764, opened 2023-01-10
fix: pop drop select route on native

intergalacticspacehighway:
# Why
On Native, we need to pop the drop select route or else it stays mounted after drop success and also it's breaking some flow for new user primary wallet select


alantoa (approved):
LGTM✨ 
nice catch!

PR #1771, opened 2023-01-18
Feature/show2 1437 update search bar design

hirbod:
# Why

Search input should be moved to the header. 

# How

1) Moved Search-Screen to the navigation group with `headerShown` false
2) For Web and Android, used the provided hook to hide the header
3) Made a change to `Input` from universal in order to be able to overwrite paddings. I had to import from `design-system` directly after speaking to @alantoa. He will publish my changes and afterwards we should change the import once again. (no showstopper afaik)
4) Aligned layout to fit native experience (no back button but instead a Cancel text on the right side)
5) In order to prevent a jump and have both search icons (home) aligned at the exact position (on search screen), I had to drastically reduce the height and reduce paddings, which was fine according to @alantoa

# Test Plan

| Before  | After iOS | After Android |
| ------------- | ------------- | ------------- |
| CleanShot 2023-01-13 at 19 32 03  |   |Bildschirm­foto 2023-01-18 um 19 46 46|

Tested on iOS, Android and web.

### DIFF
--- a/packages/app/components/header/header-cancel.tsx
+++ b/packages/app/components/header/header-cancel.tsx
@@ -1,0 +1,29 @@
+import { PressableScale } from "@showtime-xyz/universal.pressable-scale";
+import { useRouter } from "@showtime-xyz/universal.router";
+import { Text } from "@showtime-xyz/universal.text";
+
+type HeaderCancelProps = {
+  canGoBack?: boolean;
+  withBackground?: boolean;
+};
+export const HeaderCancel = ({ canGoBack = true }: HeaderCancelProps) => {
+  const router = useRouter();
+  const canGoHome = router.pathname.split("/").length - 1 >= 2;
+
+  return (
+    <PressableScale
+      hitSlop={{ top: 10, bottom: 10, left: 10, right: 10 }}
+      onPress={() => {
+        if (canGoBack) {
+          router.pop();
+        } else if (canGoHome) {
+          router.push("/home");
+        } else {
+          router.push("/search");
+        }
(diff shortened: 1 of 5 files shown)


hirbod:
After speaking with Isha I've reverted the "Cancel" approach on the right side and restored the back button. She agreed to reduce paddings though, making the swap more linear and not jumping.

alantoa (approved):
LGTM! 🔥🔥🔥

intergalacticspacehighway (approved):
LGTM ❇️  Congrats on the first PR!

packages/app/components/header/header-cancel.tsx
```diff
@@ -0,0 +1,28 @@
...
+type HeaderCancelProps = {
+  canGoBack?: boolean;
+};
+export const HeaderCancel = ({ canGoBack = true }: HeaderCancelProps) => {
+  const router = useRouter();
+  const canGoHome = router.pathname.split("/").length - 1 >= 2;
+
+  return (
+    <PressableScale
+      hitSlop={{ top: 10, bottom: 10, left: 10, right: 10 }}
```
intergalacticspacehighway, line 14: 💯

intergalacticspacehighway:
will publish the updated Input on npm after merging

packages/app/components/header/header-cancel.tsx
hirbod: Actually, RN and RNGH Buttons now support `hitSlop={10}` as value. We could change that in the future.

PR #1772, opened 2023-01-19
fix: drop music misc feedback

alantoa:
# Why

just quickly fix and update much drop-music feedback on the slack channel.


packages/app/components/drop/drop-event.tsx
```diff
@@ -554,21 +554,17 @@ export const DropEvent = () => {
                     </View>
                     <ScrollView tw="flex-row" horizontal={true}>
                       <DataPill
-                        label={`Royalties ${watch("royalty")}%`}
+                        label={`${watch("editionSize")} Editions`}
```
hirbod: This does not take into account when there is 1 Edition, 2 Edition**s**
Singular/Plural

packages/app/components/drop/drop-free.tsx
```diff
@@ -502,17 +502,17 @@ export const DropFree = () => {
                     </View>
                     <ScrollView tw="flex-row" horizontal={true}>
                       <DataPill
-                        label={`Royalties ${watch("royalty")}%`}
+                        label={`${watch("editionSize")} Editions`}
```
hirbod: Singular/Plural (1 Edition vs 10 Editions)

packages/app/components/drop/drop-private.tsx
```diff
@@ -527,17 +527,18 @@ export const DropPrivate = () => {
                     </View>
                     <ScrollView tw="flex-row" horizontal={true}>
                       <DataPill
-                        label={`Royalties ${watch("royalty")}%`}
+                        label={`${watch("editionSize")} Editions`}
```
hirbod: singular/plural

hirbod (commented):
LGTM, just had one comment about singular and plural for editions.

packages/app/components/drop/drop-event.tsx
alantoa: 👍 yeah, I was though it too.

alantoa: just fixed and update it, thanks remind me.

PR #1773, opened 2023-01-19
Feature/show2 1438 turn i already have a wallet into

alantoa:
# Why

just updated the connet wallet button UI, making it UI same as the social login button. 

# Before  

CleanShot 2023-01-19 at 22 19 37@2x

# After 
CleanShot 2023-01-19 at 22 15 51@2x


intergalacticspacehighway (approved):
LGTM ❇️

PR #1774, opened 2023-01-19
fix: content type icon not showing

alantoa:
# Why

just fixed the drop content type icon not showing on video or other media type.


intergalacticspacehighway (approved):
LGTM ❇️

PR #1775, opened 2023-01-19
Add a rule to enforce a max drop title length of 255

jorgeavaldez:
# Why

We were receiving titles that were too long and this was breaking the indexer.

More context in the ticket: https://linear.app/showtime/issue/SHOW2-1445/validate-drop-name-is-less-than-255-characters

# How

Chain `max` at the end of the title validator

# Test Plan

In staging


alantoa (approved):
LGTM! 💯

intergalacticspacehighway (approved):
LGTM 🔥🔥🔥

PR #1776, opened 2023-01-20
fix: expo-image causing crash

alantoa:
# Why 

I found that if we use the `expo-image` to `FlashList` will be got is a very poor performance. 
sometimes got crashes(image out of the memory) or sometimes loads very slowly. 

# How 
so I have to bring it back to FastImage first, I just created an ExpoImage example to [here](https://github.com/alantoa/expo-image-issue), I'll continue to dig it.


intergalacticspacehighway (approved):
LGTM ❇️

PR #1777, opened 2023-01-20
feat: switch to expo-image

alantoa:
# Why

# How

# Test Plan


hirbod:
There is a new PR at expo which brings async rendering, which should improve more issues.
I think we should still wait and not merge, the library is currently receiving plenty of issue reports and we should wait until its RC.

https://github.com/expo/expo/pull/21086

apps/expo/package.json
```diff
@@ -154,6 +154,7 @@
     "expo-font": "~11.0.1",
     "expo-gl": "~12.0.0",
     "expo-haptics": "~12.0.1",
+    "expo-image": "^1.0.0-beta.4",
```
hirbod:
```suggestion
    "expo-image": "^1.0.0-beta.6",
```

apps/storybook-react-native/package.json
```diff
@@ -48,6 +48,7 @@
     "expo-font": "~11.0.1",
     "expo-gl": "~12.0.0",
     "expo-haptics": "~12.0.1",
+    "expo-image": "^1.0.0-beta.4",
```
hirbod:
```suggestion
    "expo-image": "^1.0.0-beta.6",
```

yarn.lock
```diff
@@ -9460,6 +9460,7 @@ __metadata:
     expo-font: ~11.0.1
     expo-gl: ~12.0.0
     expo-haptics: ~12.0.1
+    expo-image: ^1.0.0-beta.4
```
hirbod, line 9463:
```suggestion
    expo-image: ^1.0.0-beta.6
```

yarn.lock
```diff
@@ -9660,6 +9661,7 @@ __metadata:
...
```
hirbod:
```suggestion
    expo-image: ^1.0.0-beta.6
```

apps/expo/app.config.js
```diff
@@ -155,8 +155,6 @@ export default {
     "sentry-expo",
     "./plugins/react-native-cronet.js",
     "./plugins/with-animated-webp-support.js",
```
hirbod: I think this one can also be removed

packages/app/components/media/index.tsx
```diff
@@ -126,7 +126,7 @@ function Media({
             style={sizeStyle}
             blurhash={item?.blurhash}
             isMuted={numColumns > 1 ? true : isMuted}
-            resizeMode={resizeMode as ResizeMode}
+            resizeMode={resizeMode as any}
```
hirbod, line 129: contentFit instead resizeMode

packages/design-system/image/image.tsx
```diff
@@ -10,12 +12,12 @@ export type ImgProps = ExpoImageProps & {
...
 const StyledExpoImage = styled(Image);
 
-type ImageProps = Omit<ExpoImageProps, "resizeMode"> & {
+type ImageProps = Omit<ImgProps, "resizeMode"> & {
   tw?: string;
   alt?: string;
   blurhash?: string;
-  resizeMode?: any;
-} & ImgProps;
+  resizeMode?: ResizeMode;
```
hirbod, line 22: Don't you think we should be consistent and change it across the app? Maybe too much effort though

packages/design-system/image/image.tsx
```diff
@@ -30,10 +32,15 @@ function StyledImage({
 }: ImageProps) {
   return (
     <StyledExpoImage
-      source={source}
       style={[{ height, width, borderRadius }, style as any]}
       contentFit={contentFit ?? resizeMode}
       placeholder={{ blurhash, width, height }}
+      source={{
+        ...(typeof source === "object" && source),
```
hirbod: Are there any places where we pass a `require` instead a source? If yes, this might break, even though its a copy from my App

alantoa: yeah, good catch!

packages/design-system/image/image.tsx
alantoa: due to we are using `expo-av` now and it has a `resizeMode` prop, but i'm also not sure should we continue to use `resizeMode` or use new `contentFit` prop. 
did you prefer to use `contentFit`? happy to change it.

packages/design-system/image/image.tsx
```diff
@@ -30,17 +32,25 @@ function StyledImage({
...
+        ? {
+            ...source,
+            headers: {
+              Accept: "image/webp,*/*;q=0.8",
+            },
+          }
+        : source,
+    [source]
+  );
+
```
hirbod: 💯

hirbod (approved):

packages/design-system/image/image.tsx
hirbod: Let's keep it for a future task. Right now we have over 102 `resizeMode` across 22 files. We can revise this in a Spring cleaning

alantoa: >Spring cleaning 

🤣 🤣 🤣

intergalacticspacehighway (approved):
LGTM ❇️

packages/app/components/feed-item/feed-item.tsx
```diff
@@ -188,31 +187,19 @@ export const FeedItem = memo<FeedItemProps>(function FeedItem({
   useEffect(() => {
     setMomentumScrollCallback?.(showHeader);
   }, [setMomentumScrollCallback, showHeader]);
-
+  console.log(getMediaUrl({ nft, stillPreview: true }));
```
hirbod: nit: console.log

hirbod (approved):

PR #1778, opened 2023-01-20
hotfix: create screen throws bottomsheet context error

intergalacticspacehighway:
# Why
We're using native sheet on iOS so I think we'd need the `useNativeModal` true on iOS. 

# Test Plan
- Open create form on iOS


alantoa (approved):
LGTM! thanks for fount it! 🙏

PR #1784, opened 2023-01-24
Hotfix/missing icon and types

hirbod:
# Why

Hotfix for missing Icon music_presave. We agreed to quickly use Spotify but will most likely change that.
Also fixed a type at FeedItem

# How

# Test Plan


intergalacticspacehighway (approved):
LGTM ❇️

PR #1787, opened 2023-01-26
fix: comments recycling

hirbod:
# Why

The comments are powered by FlashList. The real benefit from FlashList vs FlatList is that we can take advantage of recycling. In order to fully leverage recycling, we need to take care of some rules.

# How

1) No `key` prop down the whole tree. Using `key` forces a re-render and breaks recycling (more blanks) An exception to this rule is when using `.map` like we do for the replies. While still not preferred, but unavoidable. But instead of tying the data to an id, we need to use the index as key.

https://shopify.github.io/flash-list/docs/fundamentals/performant-components#remove-key-prop

2) Removing `key` comes with a cost in terms of state-recycling. When children leverage `useState`, the values will be recycled. We need to store a ref and actually detect the change and reset it back to the initial value.

https://shopify.github.io/flash-list/docs/recycling/

Even though this change causes two `useState` per prop change, it's still a lot faster. The best would be not relying on state at all and instead consuming from a "live" global state. (or even combining the calls to one object or `useReducer`. But these are micro optimizations we could talk about at a later stage.

# Test Plan

| Before (many blanks)  | After (no blanks) |
|  <video src="https://user-images.githubusercontent.com/504909/214960296-a6b85f48-5518-4a03-a244-5268ac03f96b.mp4"></video> | <video src="https://user-images.githubusercontent.com/504909/214960370-82479c80-81ef-408c-8249-ec8b036d3500.mp4"></video> |

### DIFF
--- a/packages/app/components/comments/comment-row.tsx
+++ b/packages/app/components/comments/comment-row.tsx
@@ -67,9 +67,7 @@ function CommentRowComponent({
   const repliesCount = comment.replies?.length ?? 0;
   const replies = useMemo(
     () =>
-      repliesCount > 0
-        ? comment.replies!.slice(0, displayedRepliesCount).reverse()
-        : [],
+      repliesCount > 0 ? comment.replies!.slice(0, displayedRepliesCount) : [],
     [comment.replies, repliesCount, displayedRepliesCount]
   );
   const isMyComment = useMemo(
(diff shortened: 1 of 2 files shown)


intergalacticspacehighway (approved):
wow, good catch. LGTM ❇️

alantoa (approved):
LGTM! nice catch! ✨

PR #1788, opened 2023-01-27
fix: show spotify url on pre-save drop after the release

intergalacticspacehighway:
# Why
- show spotify url on pre-save drop after the release

# How
Show the Play button only after the release date of the song.

# Test Plan


intergalacticspacehighway:
Tiny change, so merging! 🙏

PR #1790, opened 2023-01-30
Feature/show2 1457 replace share screen

alantoa:
# Why

replace the QR Code screen with the share screen:
- when a Drop is created
- when a Drop is collected → `Share` button 

because the QR Code screen is more beautiful.  

# How 

- kill the old share screen and use the `QRCodeModal` component to replace it and reduce a lot of repeat code. 
- find out a better Twitter outline style icon to apply on the QR code share screen. 

# Test Plan
- created a Drop and waited for success then check the share screen. 
- selected any already collected drop and then press the `Share` button, and check the share screen.  

the share screen is just like this:


intergalacticspacehighway (approved):
LGTM 😍😍😍

hirbod:
LGTM. Looks very good 👍🏼

hirbod (approved):

PR #1791, opened 2023-01-31
fix: username is too long on Suggested

alantoa:
# Why

Before: 
CleanShot 2023-01-31 at 14 31 36

After: 

CleanShot 2023-01-31 at 14 29 23


intergalacticspacehighway (approved):
LGTM ❇️

PR #1795, opened 2023-01-31
fix: Notifications (List) perf

hirbod:
# Why

I discovered that the Notification list wasn't performing optimally, especially on Android.

# How

After examining and measuring the performance, I found that the function `getFormatDistanceStrictToWeek` was slow due to multiple calls to `new Date` and multiple passes through `date-fns`. This was adding extra milliseconds to the render time. I refactored the function to minimize calls to `new Date` (down to 2) and completely removed the usage of `date-fns` at this stage. This should now prevent significant slowdowns, although date operations are inherently slow. I tested the solution against a large list and saw improved results. Additionally, I memoized the Items and some of their return values, which I deemed expensive.

With the upcoming release of Expo SDK 48, we should reevaluate all of our time and date functions and consider migrating them to `Intl`. This will provide native support for iOS and Android. Additionally, Intl has support for relative time, which could be very beneficial for us.

# Test Plan

### DIFF
--- a/packages/app/utilities.ts
+++ b/packages/app/utilities.ts
@@ -1,7 +1,7 @@
-import * as React from "react";
+import React from "react";
 import { Platform } from "react-native";
 
-import { formatDistanceToNowStrict, formatDistanceStrict } from "date-fns";
+import { formatDistanceToNowStrict } from "date-fns";
 import { ResizeMode } from "expo-av";
 import * as FileSystem from "expo-file-system";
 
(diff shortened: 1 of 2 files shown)


packages/app/utilities.ts
```diff
@@ -637,24 +637,17 @@ export const findTokenChainName = (chainId?: string) => {
...
-  if (days === 0) {
-    const distanceHours = formatDistanceStrict(new Date(time), new Date(), {
-      unit: "hour",
-    });
-    const hours = Number(distanceHours.split(" ")[0]);
-    return `${hours}h`;
-  }
-  if (days >= 7) {
-    return `${Math.ceil(days / 7)}w`;
+  const currentDate = new Date();
```
hirbod, line 640: Actually, we could place `currentDate` outside, which would save one more `new Date` call. Just looks ugly

intergalacticspacehighway (approved):
> consider migrating them to Intl
+1

LGTM ⚡

alantoa (approved):
wow, nice catch! LGTM! ⚡

packages/app/utilities.ts
hirbod: Thumbs up = it's ugly don't do it? 

Or shall I do it?

hirbod: @alantoa @intergalacticspacehighway 
Thumbs up = it's ugly don't do it?

Or shall I do it?

alantoa: I prefer to keeping inner!

PR #1797, opened 2023-02-01
fix: don't show header gear to wrong users

hirbod:
# Why

The Settings gear was displayed even when viewing someone else's profile and without checking if the user was logged in, which could lead to spamming the username lookup route.

# How

Had to pass the profile-user down to `HeaderDropdown` to compare some data against the signed-in user, especially when the username is not set.

# Test Plan

https://user-images.githubusercontent.com/504909/216087035-9d37015c-bf7b-4912-a602-63c483c44956.mp4

### DIFF
--- a/packages/app/components/header-dropdown.tsx
+++ b/packages/app/components/header-dropdown.tsx
@@ -1,3 +1,4 @@
+import { useMemo } from "react";
 import { Platform, useWindowDimensions } from "react-native";
 
 import { Avatar } from "@showtime-xyz/universal.avatar";
@@ -27,6 +28,7 @@ import { Text } from "@showtime-xyz/universal.text";
 import { View } from "@showtime-xyz/universal.view";
 
 import { MenuItemIcon } from "app/components/dropdown/menu-item-icon";
+import { UserProfile } from "app/hooks/api-hooks";
 import { useAuth } from "app/hooks/auth/use-auth";
 import { useCurrentUserAddress } from "app/hooks/use-current-user-address";
 import { useUser } from "app/hooks/use-user";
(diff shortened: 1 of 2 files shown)


packages/app/components/header-dropdown.tsx
```diff
@@ -36,27 +38,51 @@ import { breakpoints } from "design-system/theme";
...
   const { colorScheme, setColorScheme } = useColorScheme();
-  const { user } = useUser();
+  const { user: currentUser, isAuthenticated } = useUser();
   const { userAddress } = useCurrentUserAddress();
   const { width } = useWindowDimensions();
   const isWeb = Platform.OS === "web";
   const isMdWidth = width >= breakpoints["md"];
   const isDark = colorScheme === "dark";
+  const walletAddresses = useMemo(
+    () => user?.profile?.wallet_addresses_v2,
```
intergalacticspacehighway: nit: we can avoid memo here

packages/app/components/profile/profile.tsx
```diff
@@ -232,7 +232,13 @@ const Profile = ({ username }: ProfileScreenProps) => {
                 <HeaderLeft canGoBack={true} withBackground />
               )
             }
-            headerRight={<HeaderDropdown type="settings" withBackground />}
+            headerRight={
+              <HeaderDropdown
+                type="settings"
+                withBackground
+                data={profileData?.data}
```
intergalacticspacehighway: nit: should we rename `data` to just `user`?

packages/app/components/header-dropdown.tsx
```diff
@@ -36,27 +38,51 @@ import { breakpoints } from "design-system/theme";
...
+  );
+
+  // If the user is not authenticated, don't show the dropdown
+  if (!isAuthenticated) return null;
+
+  // If the user is on a profile page, and the currentUser is not the same as the passed user,
+  // don't show the dropdown menu (only show it on your own profile page)
+  if (
+    currentUser?.data.profile.username !== user?.profile.username &&
+    !walletAddressMatch
```
intergalacticspacehighway, line 73: curious why we also need to match the address, is the username not sufficient? 🤔

intergalacticspacehighway: okay, got it, new users might not have it set. Good catch 😄

intergalacticspacehighway (approved):
LGTM ❇️

packages/app/components/header-dropdown.tsx
hirbod: You're totally right. I can just use the value directly

packages/app/components/profile/profile.tsx
hirbod: Yes. Honestly thought that but being sick made me lazy lol.

packages/app/components/header-dropdown.tsx
hirbod: Exactly. Because you can basically skip the profile as new user and without that check you would never be able to see the gear again.

alantoa (approved):
LGTM ✨

intergalacticspacehighway (approved):
LGTM ❇️

PR #1799, opened 2023-02-02
fix: missed profile avatar on web

alantoa:
# Why
I just found that we missed the profile avatar on the desktop because [this PR](https://github.com/showtime-xyz/showtime-frontend/pull/1797/files) missed passing the `user` prop to the `header-right.tsx` file. 
cc @intergalacticspacehighway @hirbod 
# How 

changed the type and passed the user prop to the `HeaderDropdown` component in the `header-right.tsx` file

# Test Plan

## Before 

CleanShot 2023-02-02 at 16 07 44

## After 
CleanShot 2023-02-02 at 16 16 24


alantoa:
because already affected the product so have to deploy it first!

hirbod:
Good catch.

PR #1801, opened 2023-02-02
fix: add-wallet button doesn't work

alantoa:
# Why

A few users reported that the add-wallet button doesn't work sometimes on the Wallet Setting page and no prompt to the user, but we already add `Alert` to prompt when connect got an error from the backend. 

So I guess and tried this because `openConnectModalRef` sometimes is `null`, with no reaction when the user clicks the `Add Wallet` button. 

CleanShot 2023-02-02 at 23 22 01 

# How
- Prompt the user to refresh the page when `openConnectModalRef` is `null`. 
- Upgrade the `wagmi` and `ethers` packages versions maybe can fix something. 

# Test Plan 

Check connect the wallet to sign-in flow work well with all wallets since I have updated the `wagmi` version, I have tested `Rainbow` and `MetaMask` and work well.

### DIFF
--- a/packages/app/hooks/use-wallet/use-wallet.web.ts
+++ b/packages/app/hooks/use-wallet/use-wallet.web.ts
@@ -9,6 +9,8 @@ import {
   useDisconnect,
 } from "wagmi";
 
+import { Alert } from "@showtime-xyz/universal.alert";
+
 import { useWeb3 } from "app/hooks/use-web3";
 
 import { useLatestValueRef } from "../use-latest-value-ref";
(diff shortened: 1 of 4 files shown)


packages/app/hooks/use-wallet/use-wallet.web.ts
```diff
@@ -67,7 +69,13 @@ const useWallet = (): UseWalletReturnType => {
     return {
       address,
       connect: async () => {
-        openConnectModalRef.current?.();
+        if (openConnectModalRef.current) {
+          openConnectModalRef.current();
+        } else {
+          Alert.alert(
+            "Oops, connect wallet failed, please refresh the page and then try again."
```
hirbod:
```suggestion
            "Oops, connection to wallet failed. Please refresh the page and try again."
```

hirbod (changes requested):

packages/app/hooks/use-wallet/use-wallet.web.ts
alantoa: thank you!

intergalacticspacehighway (approved):
LGTM ❇️

PR #1802, opened 2023-02-02
chore: update complete profile modal copy

alantoa:
# Why
we need to update the complete profile modal copy and add `<CompleteProfileModalContent />` to the `drop-select` screen.


intergalacticspacehighway (approved):
LGTM ❇️

PR #1804, opened 2023-02-04
fix: wrong comment count

hirbod:
# Why
The comment count of FeedItem was totally wrong

# How
The comment count on the FeedItem was calculated based on the length of the returned data from `useComments`.
1) This caused an API roundtrip for no reason and was a complete wrong number (just first page, no replies included)
2) The `nft` object on FeedItem already has the right value from the API (`nft.comment_count`).

The only change now is that the count does not instantly update when a new comment is added. I think that's okay, I'm not sure if this is a crucial feature @intergalacticspacehighway @alantoa. 

If we require an instant reflection of the amount, we might have to mutate nft every time a new comment is posted. Should we spend time on this or just use the real, cached value?

# Test Plan

### DIFF
--- a/packages/app/components/feed-item/details.tsx
+++ b/packages/app/components/feed-item/details.tsx
@@ -50,7 +50,7 @@ export const NFTDetails = ({ nft, edition, detail }: NFTDetailsProps) => {
         <View tw="flex-row">
           <Like nft={nft} />
           <View tw="w-6" />
-          <CommentButton nft={nft} count={detail?.comment_count} />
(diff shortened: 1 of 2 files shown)


intergalacticspacehighway (approved):
LGTM ❇️ 

> This caused an API roundtrip for no reason and was a complete wrong number 

😱  good catch!

> Should we spend time on this or just use the real, cached value?

i think it's fine for now, but yeah we can mutate the swr cache after coment success in next PR.

PR #1805, opened 2023-02-05
fix: avoid soft input breaking on iOS

hirbod:
# Why
Edit wallet nickname was breaking due to soft input enabled globally.

# How

Disabled for this particular screen.

# Test Plan


intergalacticspacehighway (approved):
LGTM ❇️

alantoa (approved):
thank you!

PR #1806, opened 2023-02-05
Feat/comments redesign

hirbod:
# Why

This is the first step of the comments redesign. Since reactions and context menu are not ready yet, we're starting with the new and condensed version, also without any usage of `Space Grotesk` anymore. 

# How

Followed re-design by Isha, just added the trash icon as a fallback for now (and no colored hearts yet) until we have reactions ready.

# Test Plan

Fully tested on Web, iOS and Android.

| Before  | After |
| Simulator Screen Shot - iPhone 14 Pro Max - 2023-02-05 at 18 08 58 | Simulator Screen Shot - iPhone 14 Pro Max - 2023-02-05 at 18 09 17 |


intergalacticspacehighway (approved):
LGTM 🔥🔥

we can add double tap to like comments in future!

alantoa (approved):
LGTM! just tested it! it's 🔥 🔥 🔥

hirbod:
> LGTM 🔥🔥
> 
> 
> 
> we can add double tap to like comments in future! 

In the future, we need to add reactions, guess no double tap needed. What I was missing is a longPress context menu. Looks like Zeego has it on its todo but is not implemented yet. There is a native library for this though which supports iOS and Android but no web

PR #1812, opened 2023-02-06
fix: recycling for FeedItem

hirbod:
# Why

FlashLists recycling was breaking because we used `key` on LikeContextProvider

# How

Added proper recycling back with a ref. Unfortunately, we're relying on `onLayout` (we should think how to get rid of this) which won't trigger again when recycled. So I need to re-measure the view. Even though we're doing two useState calls (and a measuring) here, we still take advantage of a lot of recycling which leads into better JS-FPS perf on fast scrolls.

# Test Plan


intergalacticspacehighway (approved):
LGTM ❇️

alantoa (approved):
LGTM 💯

PR #1813, opened 2023-02-07
nit: improve twitter intent

intergalacticspacehighway:
# Why
Improve Twitter intent message. Also reduced the size of media to prevent overflow scroll.

# Test Plan
Claim a drop and share on twitter


alantoa (approved):
LGTM!

PR #1815, opened 2023-02-07
fix: align drop ui info modal more with design

hirbod:
# Why

The drop ui modal was not in line with the layout

# How

Optimized the look to match Figma design better and also fixed the transition on mobile.

# Test Plan


intergalacticspacehighway (approved):
LGTM ❇️

PR #1816, opened 2023-02-07
Feature/show2 1485 lightdark mode system by default

alantoa:
# Why

- make sure dark mode is using the system/light by default
- fix dark mode got reset when refreshing the page. 

https://user-images.githubusercontent.com/37520667/217250890-72fb45d8-b109-4595-a548-e53cceaf6a27.mp4


packages/design-system/dropdown-menu/index.tsx
```diff
@@ -270,7 +270,7 @@ const DropdownMenuItemIcon = DropdownMenu.menuify(
 );
 
 const DropdownMenuItemNativeIcon =
-  Platform.OS === "web" ? Fragment : DropdownMenu.ItemIcon;
+  Platform.OS === "web" ? () => <Fragment /> : DropdownMenu.ItemIcon;
```
hirbod, line 274: Should we `useMemo` this?

hirbod (approved):
LGTM, just one comment

packages/design-system/dropdown-menu/index.tsx
alantoa: this is only temporary because the latest `zeego` (`v1.0.2 `) version fixes this, I'm going to update it next.

alantoa:
also, I just noticed that the dark mode has the same issue on native, will fix it!

intergalacticspacehighway (approved):
LGTM 🌔

PR #1817, opened 2023-02-07
Fix/comments padding

hirbod:
# Why

Addressed comments by Isha regarding padding for long comments.

# How

Addressed comments by Isha regarding padding for long comments.
Also added swipe-to-delete for native and moved the heart icon as asked. Promisified the Alert to auto close the swipeable component. Fixed some tiny issues

# Test Plan


alantoa (approved):
checked all thing and work well! 🔥

PR #1819, opened 2023-02-07
fix: comments navigate to profile

alantoa:
# Why

two bugs  in the comments screen, it’s also happened in the new comments UI before:
- the ⬅️ button didn’t show when pressing the user avatar to enter the profile screen.
- back to the feed screen directly instead of the comments screen when swiping to back the profile screen.

# How

use `router.push` to fix it.  

# Test Plan

### DIFF
--- a/packages/app/components/comments/comment-row.tsx
+++ b/packages/app/components/comments/comment-row.tsx
@@ -1,5 +1,4 @@
 import { memo, useCallback, useMemo, useState, useRef } from "react";
-import { Platform } from "react-native";
 
 import { useIsDarkMode } from "@showtime-xyz/universal.hooks";
 import { useRouter } from "@showtime-xyz/universal.router";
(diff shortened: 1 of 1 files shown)


packages/app/components/comments/comment-row.tsx
```diff
@@ -134,18 +131,9 @@ function CommentRowComponent({
...
-        StackActions.replace("profile", {
-          username: username,
-        })
-      );
-    }
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, []);
+  const handleOnUserPress = (username: string) => {
+    router.push(`/@${username}`);
+  };
```
hirbod, line 136: Please wrap this into useCallback, since its a list, we need to memoize everything

hirbod (changes requested):

packages/app/components/comments/comment-row.tsx
alantoa: oh sorry missed it, on it

hirbod: You merged it now without changing, right?

PR #1821, opened 2023-02-07
feat: allow drop creators to delete all comments

hirbod:
# Why

A drop creator should be able to delete all comments

# How

Added a check of the current drop is made by logged in user, if yes, allow swipe to delete on all of them

# Test Plan


alantoa (approved):
LGTM! but still wait backend depoly sucessed then test it.

hirbod:
Tests passed!

PR #1823, opened 2023-02-08
chore: add webcredentials in aasa

intergalacticspacehighway:
# Why
@hirbod found we're missing webcredentials, so adding it. This should enable Phone number auto suggestion, better UX!

# How
- add webcredentials in aasa


hirbod:
This also needs an entry in app.config.js, right next to `applinks:${url}` it needs `webcredentials:...`

alantoa (approved):
LGTM 💯

PR #1825, opened 2023-02-08
fix: update qr code modal

alantoa:
# Why

quick changes on the web success modal:
- remove “Scan to Collect” because it’s simpler and makes the image too big for the modal size in height
- add Copy Link between Twitter and Download


hirbod (approved):
LGTM 🔥

PR #1826, opened 2023-02-08
fix: cap free drop editions

alantoa:
# Why

- from the [Slack](https://showtime-rq88331.slack.com/archives/C02PTNZMEPQ/p1675880814486929): if creators non-verified we need to cap the at 50 editions. 
- fix free drop UI. 

# Test Plan

check if the free drop edition works well.

### DIFF
--- a/packages/app/components/drop/drop-free.tsx
+++ b/packages/app/components/drop/drop-free.tsx
@@ -120,7 +120,10 @@ export const DropFree = () => {
     resolver: yupResolver(dropValidationSchema),
     mode: "onBlur",
     reValidateMode: "onChange",
-    defaultValues: defaultValues,
+    defaultValues: {
+      ...defaultValues,
+      editionSize: isVrified ? defaultValues.editionSize : 50,
+    },
   });
 
   const gatingType = watch("gatingType");
--- a/packages/app/components/drop/drop-free.tsx
+++ b/packages/app/components/drop/drop-free.tsx
@@ -71,7 +71,7 @@ export const DropFree = () => {
   const { rudder } = useRudder();
   const { data: userProfile } = useMyInfo();
 
-  const isVrified = false;
+  const isVerified = false;
   userProfile?.data?.profile.has_verified_phone_number &&
     userProfile?.data?.profile.verified;
 
--- a/packages/app/components/drop/drop-free.tsx
+++ b/packages/app/components/drop/drop-free.tsx
@@ -71,8 +71,8 @@ export const DropFree = () => {
   const { rudder } = useRudder();
   const { data: userProfile } = useMyInfo();
 
-  const isVerified = false;
-  userProfile?.data?.profile.has_verified_phone_number &&
+  const isVerified =
+    userProfile?.data?.profile.has_verified_phone_number ||
     userProfile?.data?.profile.verified;
 
   const dropValidationSchema = useMemo(
@@ -122,7 +122,7 @@ export const DropFree = () => {
     reValidateMode: "onChange",
     defaultValues: {
       ...defaultValues,
-      editionSize: isVrified ? defaultValues.editionSize : 50,
+      editionSize: isVerified ? defaultValues.editionSize : 50,
     },
   });
 
--- a/packages/app/components/drop/drop-free.tsx
+++ b/packages/app/components/drop/drop-free.tsx
@@ -103,7 +103,7 @@ export const DropFree = () => {
         googleMapsUrl: yup.string().url(),
         radius: yup.number().min(0.01).max(10),
       }),
-    [isVrified]
+    [isVerified]
   );
 
   const {
--- a/packages/app/components/drop/drop-free.tsx
+++ b/packages/app/components/drop/drop-free.tsx
@@ -1,4 +1,4 @@
-import React, { useRef, useState } from "react";
+import React, { useRef, useState, useMemo } from "react";
 import {
   Platform,
   ScrollView as RNScrollView,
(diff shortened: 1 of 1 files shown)


packages/app/components/drop/drop-free.tsx
```diff
@@ -49,54 +50,61 @@ const SECONDS_IN_A_DAY = 24 * 60 * 60;
...
+  radius: 1, // In kilometers
+  hasAcceptedTerms: false,
+  notSafeForWork: false,
+};
 export const DropFree = () => {
   const isDark = useIsDarkMode();
   const { rudder } = useRudder();
+  const { data: userProfile } = useMyInfo();
+
+  const isVrified = false;
```
hirbod:
```suggestion
  const isVerified = false;
```

packages/app/components/drop/drop-free.tsx
```diff
@@ -49,54 +50,61 @@ const SECONDS_IN_A_DAY = 24 * 60 * 60;
...
+      yup.object({
+        file: yup.mixed().required("Media is required"),
+        title: yup.string().required().max(255),
+        description: yup.string().max(280).required(),
+        editionSize: yup
+          .number()
+          .required()
+          .typeError("Please enter a valid number")
+          .min(1)
+          .max(isVrified ? 100000 : 50)
```
hirbod:
```suggestion
          .max(isVerified ? 100000 : 50)
```

packages/app/components/drop/drop-free.tsx
```diff
@@ -49,54 +50,61 @@ const SECONDS_IN_A_DAY = 24 * 60 * 60;
...
+          .default(isVrified ? defaultValues.editionSize : 50),
```
hirbod:
```suggestion
          .default(isVerified ? defaultValues.editionSize : 50),
```

hirbod (changes requested):

hirbod:
Are these limits also enforced in the backend? @jorgeavaldez

jorgeavaldez:
> Are these limits also enforced in the backend? @jorgeavaldez 

Absolutely

packages/app/components/drop/drop-free.tsx
alantoa: damn 🤣

packages/app/components/drop/drop-free.tsx
```diff
@@ -49,54 +50,61 @@ const SECONDS_IN_A_DAY = 24 * 60 * 60;
...
+        hasAcceptedTerms: yup
+          .boolean()
+          .default(defaultValues.hasAcceptedTerms)
+          .required()
+          .isTrue("You must accept the terms and conditions."),
+        notSafeForWork: yup.boolean().default(defaultValues.notSafeForWork),
+        googleMapsUrl: yup.string().url(),
+        radius: yup.number().min(0.01).max(10),
+      }),
+    [isVrified]
```
hirbod:
```suggestion
    [isVerified]
```

hirbod (changes requested):

packages/app/components/drop/drop-free.tsx
```diff
@@ -49,54 +50,61 @@ const SECONDS_IN_A_DAY = 24 * 60 * 60;
...
+  notSafeForWork: false,
+};
 export const DropFree = () => {
   const isDark = useIsDarkMode();
   const { rudder } = useRudder();
+  const { data: userProfile } = useMyInfo();
+
+  const isVerified = false;
+  userProfile?.data?.profile.has_verified_phone_number &&
+    userProfile?.data?.profile.verified;
```
hirbod: This line is wrong. isVerified is always false, the condition is written wrong

hirbod (changes requested):

packages/app/components/drop/drop-free.tsx
hirbod: Should be:

```tsx
  const isVerified =  userProfile?.data?.profile.has_verified_phone_number && userProfile?.data?.profile.verified;
```

hirbod (changes requested):

packages/app/components/drop/drop-free.tsx
alantoa: oh damn i just do a test, will remove it

packages/app/components/drop/drop-free.tsx
```diff
@@ -71,8 +71,8 @@ export const DropFree = () => {
   const { rudder } = useRudder();
   const { data: userProfile } = useMyInfo();
 
-  const isVerified = false;
-  userProfile?.data?.profile.has_verified_phone_number &&
+  const isVerified =
+    userProfile?.data?.profile.has_verified_phone_number ||
     userProfile?.data?.profile.verified;
```
hirbod: Are you sure that an OR clause is correct here? Is a verified phone number really considered a verified user? I don't think so

packages/app/components/drop/drop-free.tsx
alantoa: yeah, it's should be 
`const isVerified =  userProfile?.data?.profile.has_verified_phone_number || userProfile?.data?.profile.verified;`  
because they all belong to verified.

packages/app/components/drop/drop-free.tsx
alantoa: yeah, i have changed https://github.com/showtime-xyz/showtime-frontend/pull/1826#discussion_r1100670822

hirbod (approved):
LGTM 🔥

PR #1830, opened 2023-02-09
chore: don't commit history files

hirbod:
# Why

Some really useful tools like local history write files into a local .history folder (others in .lh/).
This files should not be commited

# How

Added to history related files to gitignore. 

# Test Plan


alantoa (approved):
LGTM

PR #1831, opened 2023-02-09
Feature/show2 1511 ios bug create music drop modal

alantoa:
# Why

-  Create Music Drop modal “Unlimited” should not have less opacity (but the rest of Edition size, yes) 

## Before 

CleanShot 2023-02-10 at 00 45 07

## After 
CleanShot 2023-02-10 at 00 43 51


intergalacticspacehighway (approved):
LGTM ❇️

PR #1833, opened 2023-02-09
fix: tracking screens/page correctly

hirbod:
# Why

We have been tracking pages (web) and screens (app) wrong. We used a custom track event instead of using the right tools

# How

Added a check for web and app and call `page` or `screen`. I had to ignore the types for `page` because of how we import the rudderClient, but it works.

# Test Plan


intergalacticspacehighway (approved):
LGTM ❇️

PR #1834, opened 2023-02-10
fix: bail out of retry logic in swr if status 400, 403 or 404

hirbod:
# Why

Our Feed and data is cached. If we ban profiles, they might still be in the Feed cache. This causes hundreds of retries by useSWRs defaults, spamming our API and eating up client resources.

# How

While I think we should adapt a global SWRConfig strategy (even though we have exponential backoff baked in), for now I've gone with the route to only adjust this single useSWR. There is no need to retry to grab a profile with a 404 or 403 forever or hundreds of times. Therefore I have added this exception, until we decide how to proceed globally.

In my findings, a single banned profile was causing 80 XHR requests in less than 5 minutes on the home feed. This should fix this.

If you guys agree, we could ignore my PR and just add the exception to swr-provider.tsx and swr-prodiver.web.tsx and generally ignore retries for 403

# Test Plan

### DIFF
--- a/packages/app/hooks/api-hooks.ts
+++ b/packages/app/hooks/api-hooks.ts
@@ -116,7 +116,18 @@ export const useUserProfile = ({ address }: { address?: string | null }) => {
   const queryKey = address ? USER_PROFILE_KEY + address : null;
   const { data, error, isLoading } = useSWR<{
     data?: UserProfile;
-  }>(queryKey, fetcher);
+  }>(queryKey, fetcher, {
+    onErrorRetry: (error, key, _config, revalidate, { retryCount }) => {
+      // Never retry on 404 or 404.
+      if (error.status === 404 || error.status === 403) return;
+
+      // Only retry up to 5 times.
+      if (retryCount >= 5) return;
+
+      // Retry after 5 seconds.
+      setTimeout(() => revalidate({ retryCount }), 5000);
+    },
+  });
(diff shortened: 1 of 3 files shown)


packages/app/hooks/api-hooks.ts
```diff
@@ -116,7 +116,18 @@ export const useUserProfile = ({ address }: { address?: string | null }) => {
   const queryKey = address ? USER_PROFILE_KEY + address : null;
   const { data, error, isLoading } = useSWR<{
     data?: UserProfile;
-  }>(queryKey, fetcher);
+  }>(queryKey, fetcher, {
+    onErrorRetry: (error, key, _config, revalidate, { retryCount }) => {
+      // Never retry on 404 or 404.
```
hirbod:
```suggestion
      // Never retry on 403 or 404.
```

intergalacticspacehighway:
> If you guys agree, we could ignore my PR and just add the exception to swr-provider.tsx and swr-prodiver.web.tsx and generally ignore retries for 403

yeah let's do it!

alantoa:
Agreed! 💯

hirbod:
@alantoa @intergalacticspacehighway you can review again please. Should be resolved now

alantoa (approved):
LGTM 💯

intergalacticspacehighway (approved):
LGTM ❇️

PR #1835, opened 2023-02-10
feat: paid drops

intergalacticspacehighway:
# Why
- This PR integrates Stripe payments for paid drops
- Paid drops is a web-only feature for now
- All drops will be free to create by verified creators.
- Unverified creators would need to pay to create drops.

# How
- Added payment integration

# Test Plan
- Web only, make sure anything does't break on native. We hide the create button on native.
- If you're a verified creator, nothing changes for you. You can create drops without paying
- If you're a unverified creator, test the stripe success and error checkout flow. Use these dummy cards - https://stripe.com/docs/testing


package.json
```diff
@@ -36,6 +36,7 @@
     "packages/*"
   ],
   "scripts": {
+    "create-screen": "node scripts/showtime-cli",
```
intergalacticspacehighway, line 38: [experimental] trying to make a cli to create screens. Need some work but basically you can run yarn create-screen [SCREEN Name] and it'll create folders/file for you

alantoa: wow, really needed this!

packages/app/components/checkout/select-plan.tsx
```diff
@@ -0,0 +1,142 @@
...
+          throw e;
+        }
+      }
+    } catch (e: any) {
+      setError(e?.response?.data?.error.message ?? "Something went wrong");
+      Logger.error("Payment intent fetch failed ", e);
+    } finally {
+      setLoading(false);
+    }
+  };
```
hirbod: Nit: I think this could be handled without nested try/catch block using a single try/catch.
And don't you think we should be consistent and use SWR for this, instead of defining these sort of async calls inside the component?

packages/app/components/checkout/checkout-return.tsx
```diff
@@ -0,0 +1,143 @@
...
+        switch (paymentIntent?.status) {
+          case "succeeded":
+            handlePaymentSuccess();
+            break;
+          case "processing":
+            {
+              setMessage("Your payment is processing.");
+              setPaymentStatus("processing");
+              for (let i = 0; i < 20; i++) {
+                const res = await axios({
```
hirbod, line 67: Don't we want to handle this with a hook instead?

packages/app/components/drop/drop-select.tsx
```diff
@@ -20,30 +19,59 @@ export const DropSelect = () => {
   const user = useUser();
   const canCreateMusicDrop = !!user.user?.data.profile.spotify_artist_id;
   const isDark = useIsDarkMode();
+  const editionSize = user.user?.data.paid_drop_credits?.[0]?.edition_size;
+  const canUserCreateEditions =
+    (typeof editionSize === "number" && editionSize > 0) ||
+    user.user?.data.profile.verified;
+  console.log("canUserCreateEditions", canUserCreateEditions);
```
hirbod: Use Logger instead

packages/app/components/drop/drop-select.tsx
```diff
@@ -20,30 +19,59 @@ export const DropSelect = () => {
...
+              onPress={() => {
+                if (!canUserCreateEditions) {
+                  navigateToCheckout();
+                  return;
+                }
+                if (Platform.OS !== "web") {
+                  router.pop();
+                }
+                router.push("/drop/free");
+              }}
```
hirbod, line 69: This might clash when not returning after `pop`. I had weird side effects in the past when calling two navigation functions at the same time.

packages/app/components/drop/drop.tsx
```diff
@@ -10,7 +10,7 @@ const STORE_KEY = "showExplanation";
 
 export const Drop = () => {
   const [showExplanation, setShowExplanation] = useState(
-    () => store.getBoolean(STORE_KEY) ?? true
+    () => store.getBoolean(STORE_KEY + "v2") ?? true
```
hirbod: Is the anon function intentional here for the default state value? I don't see any reason for this

hirbod (commented):
LGTM ✦,  just some tiny nits/questions

alantoa:
just fixed a small modal UI bug and committed it. 

## Before 

https://user-images.githubusercontent.com/37520667/218050001-70e818be-90e8-460a-ae84-7e32e1e8826b.mp4

## After 

https://user-images.githubusercontent.com/37520667/218050021-86492f7c-27a3-441c-92ed-c2ab43e66963.mp4

packages/app/components/drop/drop-select.tsx
```diff
@@ -56,9 +84,13 @@ export const DropSelect = () => {
             }
             description="Promote your latest music: give your fans a free collectible for saving your song to their library."
             ctaLabel={
-              canCreateMusicDrop ? "Create Music Drop" : "Request Access"
+              canCreateMusicDrop ? "Create a Music Drop" : "Request Access"
             }
             onPress={() => {
+              if (!canUserCreateEditions) {
```
alantoa: AFAIK music drop should still be free for the verified account, if non-verified should be linked to the [Typeform](https://form.typeform.com/to/pXQVhkZo) to apply it.

packages/app/components/drop/drop.tsx
intergalacticspacehighway: yeah, it's nit but better for performance. It'll be only called when this component renders for the first time whereas if we do `store.getBoolean(STORE_KEY + "v2")`, it'll be called on every render although react will use the first render's value

packages/app/components/drop/drop-select.tsx
intergalacticspacehighway: oops, need to remove this. Not useful

packages/app/components/checkout/checkout-return.tsx
intergalacticspacehighway: do you mean to separate it in a different hook? yeah we can but honestly wanted to keep it stupid simple.

packages/app/components/drop/drop.tsx
alantoa: actually it missed to set `STORE_KEY + "v2"` key, so it alway is return `undefined`.  

https://user-images.githubusercontent.com/37520667/218055622-746876fc-5dd8-4dc9-90b1-ab3acdad4853.mp4

alantoa: fixed it! 
CleanShot 2023-02-10 at 17 32 26

packages/app/components/checkout/select-plan.tsx
intergalacticspacehighway: Yes! updated. Need to stop being lazy 🙈

packages/app/components/drop/drop-select.tsx
intergalacticspacehighway: yeah but we need to replace the routes. There's no .replace function right now 🤕. I think instead of modals we can use stack screen later.

packages/app/components/drop/drop-select.tsx
intergalacticspacehighway: good catch 🙏! Fixed

packages/app/components/drop/drop.tsx
intergalacticspacehighway: damn. thanks 🐐

packages/app/components/checkout/select-plan.tsx
```diff
@@ -0,0 +1,155 @@
...
+import { DropPlan, usePaidDropPlans } from "app/hooks/use-paid-drop-plans";
+import { axios } from "app/lib/axios";
+import { Logger } from "app/lib/logger";
+import { MY_INFO_ENDPOINT } from "app/providers/user-provider";
+
+export const SelectPlan = ({ setClientSecret }: { setClientSecret: any }) => {
+  const paidDropPlansQuery = usePaidDropPlans();
+  const [selectedPlan, setSelectedPlan] = useState<DropPlan | null>(null);
+  const isDark = useIsDarkMode();
+  const { trigger, data, isMutating, error } = useSWRMutation<{
```
intergalacticspacehighway, line 23: cc - @hirbod replaced it with a hook. Looks much better 😄

hirbod: We just need to keep in mind how the default error handling might kick in set in SWRConfig, but I think there is no such thing when mutating, only when fetching

packages/app/components/drop/drop.tsx
hirbod: > yeah, it's nit but better for performance. It'll be only called when this component renders for the first time whereas if we do store.getBoolean(STORE_KEY + "v2"), it'll be called on every render although react will use the first render's value


TIL, thanks!

packages/app/components/drop/drop-select.tsx
hirbod: Ok, I see. I mean, as long you know it does not clash its fine

alantoa (approved):
LGTM! 🔥 🔥 🔥

hirbod (approved):
🔥 good job

packages/app/components/checkout/checkout-form.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+    const { error } = await stripe.confirmPayment({
+      elements,
+      confirmParams: {
+        return_url: window.location.origin + "/checkout-return",
+        receipt_email: email,
+      },
+    });
+
+    Logger.error("Stripe payment failure ", error);
```
hirbod: This would log always. Is that intended?

hirbod (changes requested):

packages/app/components/checkout/checkout-form.tsx
intergalacticspacehighway: nope, good catch. My bad. Need to add an if (error). Fixed!

PR #1837, opened 2023-02-10
e2e: sign up test flow

intergalacticspacehighway:
# Why
Adds sign up test flow using a new wallet

# How
- Adds sign up test flow for iOS. Maestro cloud lacks default media in their android emulators. So it's noop on android as our sign up step depends upon media. I have asked their team if they can do something.

# Test Plan
- Sign up test should pass on maestro cloud.


alantoa (approved):
LGTM 🔥

PR #1844, opened 2023-02-11
fix: correctly handle HTTP status 401

hirbod:
# Why

401 issues are currently not working, because the `error.status` is undefined (due to our use of Axios). The status data from Axios resides in the `error.response` property.

# How

I fixed the check and moved the code a bit. I also added a bail out if the refreshToken failed once. I am still not completely satisfied with this solution, as I am seeing up to eight refreshToken requests on the Home page once we encounter a 401 (because all 401s result in at least one retry). This is still an improvement, as before (when it wasn't even executing), it would have attempted it many, many more times before eventually bailing out.

For now, this issue has been resolved, but the proper way to handle it would be to instantly log out the user if the first refreshToken results in another 401. I tried, but the AuthProvider is not available, so we need to export a generic logout function that is not related to the AuthProvider. This should prevent further 401 API spam, as once we fail to refresh the user, we should log them out.

This is at least a temporary solution to reduce the number of retries in the API.

# Test Plan

### DIFF
--- a/packages/app/providers/swr-provider.tsx
+++ b/packages/app/providers/swr-provider.tsx
@@ -72,6 +72,19 @@ export const SWRProvider = ({
           const maxRetryCount = config.errorRetryCount;
           const currentRetryCount = opts.retryCount;
 
+          if (error.response?.status === 401) {
+            // we only want to refresh tokens once and then bail out if it fails on 401
+            // this is to prevent infinite loops. Actually, we should logout the user but AuthProvider is not available here
+            if (currentRetryCount > 1) {
+              return;
+            }
+            try {
+              await refreshTokens();
+            } catch (err) {
+              return;
+            }
+          }
+
           // Exponential backoff
           const timeout =
             ~~(
@@ -84,10 +97,6 @@ export const SWRProvider = ({
             currentRetryCount > maxRetryCount
           ) {
             return;
-          }
-
-          if (error.status === 401) {
-            await refreshTokens();
           }
 
           setTimeout(revalidate, timeout, opts);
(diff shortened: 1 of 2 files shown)


packages/app/providers/swr-provider.tsx
```diff
@@ -72,6 +72,19 @@ export const SWRProvider = ({
           const maxRetryCount = config.errorRetryCount;
           const currentRetryCount = opts.retryCount;
 
+          if (error.response?.status === 401) {
+            // we only want to refresh tokens once and then bail out if it fails on 401
+            // this is to prevent infinite loops. Actually, we should logout the user but AuthProvider is not available here
+            if (currentRetryCount > 1) {
```
intergalacticspacehighway, line 78: is `currentRetryCount` 401 specific retry count?

hirbod: No its not, it is a general count, but the retry is happening for the same error, so it should be safe to add that specific here. If refreshToken would NOT fail, we'd return and not end up again inside a 401 since the timeout is coming later

intergalacticspacehighway: okay, but the retry count would have still gone up even if it's 404 for example right? since it's coming from `onErrorRetry` some internal counter in SWR

hirbod: Yes, the counter will increment for a 404 error, but only for that particular request. I don't believe there is a scenario where we have a 401 error that could result in a 404 on the next request, or vice versa. (While possible in theory)

However, it's possible that we might encounter a 401 error that could turn into a 500 error in the next request, which could potentially cause the retry-code to break. To check for this explicitly, we need to define a "let" outside the code and increment and reset it on success. Currently, without a bailout mechanism, we would retry 401 errors up to 8 times per hook, which caused the issue in my findings (and could be hundreds of times on the main feed).

intergalacticspacehighway: > but only for that particular request

Got it. I thought it was a global error request counter. My bad. LGTM ❇️

intergalacticspacehighway (approved):
LGTM ❇️

PR #1845, opened 2023-02-11
fix: cumulative layout shifts caused by cards

hirbod:
# Why

The cards on Home and Trending are causing a lot of CLS due to the "ClaimedBy" component.
Currently, we have two endpoints for NFTs: the NFT endpoint and the NFT-detail endpoint. The NFT-detail endpoint provides information about how many people have collected a specific NFT.

However, the detail endpoint is called asynchronously after a card (either the feed or trending card) has been rendered. Based on the value of this key, we decide whether to render the “claimed-by” component or not. I have temporarily fixed this issue with a simple “hack” (rendering an empty view with the final layout). This solution works well because I haven’t seen any unowned NFTs, but it could result in a gap if an NFT was not collected yet. (not looking bad though)

Knowing this information beforehand would prevent the cumulative layout shift in both cases.
I have asked for this information, in more detail [here on slack](https://showtime-rq88331.slack.com/archives/C02QD3J7DJM/p1676130946348719)

# How
Preserving the expected height upfront, until the async request is done. Works pretty well.
Also fixed some types. Drastically reduced CLS and it looks way better when scrolling through lists. 

# Test Plan

### DIFF
--- a/packages/app/components/card/card.tsx
+++ b/packages/app/components/card/card.tsx
@@ -6,6 +6,7 @@ import {
   ViewStyle,
 } from "react-native";
 
+import { ResizeMode } from "expo-av";
 import { Link, LinkProps } from "solito/link";
 
 import {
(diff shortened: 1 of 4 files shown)


intergalacticspacehighway (approved):
LGTM ❇️

PR #1847, opened 2023-02-11
fix: clamp-text not updating

hirbod:
# Why

When updating the bio on iOS or Android, the text does not change and a restart of the app is required to see the changes.

# How

This was a challenging issue to resolve. Initially, I thought the problem was due to memoizing in the Tab-View component, but it turned out to be related to how state was managed in the use-clamp-text function. A simple solution would have been to add a key to the ClampText component, but since it is widely used and also in recycling lists, I had to find a more comprehensive solution.

I solved the issue by comparing the text with a stored reference. Whenever the text changes, I update the state and reset the clamping process. This solution is stable with recycling and should not result in any additional overhead.

## Help

**@alantoa could you please update the universal design system? I had to make changes. If you don't mind, please co-author this PR to publish the changes.**

# Test Plan

| Before  | After |
| <video src="https://user-images.githubusercontent.com/504909/218285230-26feeeda-63cb-4a25-aa7b-42de36bc169a.mp4"></video> | <video alt="" src="https://user-images.githubusercontent.com/504909/218285229-6c6f33f9-62cb-4dda-97a7-2ff35c6ca7ba.mp4"></video> |

### DIFF
--- a/packages/design-system/clamp-text/use-clamp-text.tsx
+++ b/packages/design-system/clamp-text/use-clamp-text.tsx
@@ -1,4 +1,4 @@
-import { useState, useRef, useCallback, ReactNode, useEffect } from "react";
+import { useState, useRef, useCallback, ReactNode } from "react";
 import { NativeSyntheticEvent, TextLayoutEventData } from "react-native";
 
 import { ClampTextProps } from "./clamp-text";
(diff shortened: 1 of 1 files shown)


intergalacticspacehighway (approved):
LGTM ❇️

PR #1848, opened 2023-02-11
fix: long text in bio break HoverCard overflow

hirbod:
# Why

Long text in bio would break overflow in AvatarHoverCard.

# How

Added a word-break strategy.

# Test Plan


intergalacticspacehighway (approved):
LGTM ❇️

PR #1849, opened 2023-02-12
feat: optimize linkify

hirbod:
# Why

Our linkify function had several flaws. It included punctuation at the end of the URL in the final match and did not support URLs without a protocol, such as [www.google.de](http://www.google.de/).

# How

I optimized the regular expression to exclude trailing punctuation. Additionally, I added checks for "[www](http://www/)." and prepended "https://" to it. I chose not to add "http" because all websites should use "https" now. I also removed the protocol from the printed URL to save space and make it look better, while still retaining the protocol in the URL itself.

After trying 30 different regular expressions, I settled on a straightforward and simple solution based on the previous one, which provided good results.

# Test Plan

### DIFF
--- a/packages/app/lib/linkify.tsx
+++ b/packages/app/lib/linkify.tsx
@@ -27,7 +27,7 @@ export const linkifyDescription = (text?: string, tw?: TW) => {
     replacedText,
     /(https?:\/\/\S+|www\.\S+)\b/g,
     (match, i) => {
-      if (match.startsWith("www")) {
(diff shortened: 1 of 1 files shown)


alantoa (approved):
LGTM, thanks!

intergalacticspacehighway (approved):
> After trying 30 different regular expressions, I settled on a straightforward and simple solution based on the previous one, which provided good results.

LGTM 👏

PR #1851, opened 2023-02-12
feat: enableFreeze for our App

hirbod:
# Why

We can boost the performance of our App pretty heavily if we `enableFreeze`

# How

I've enbaled `enableFreeze(true)` end tested the app manually as far as I could. I haven't seen any side-effects but it could sometimes lead into missed updated, which I was not able to determine. Maybe you guys already tried in the past and had reasons not to activate it, but in general, it gives a lot of benefits without downsides.

We also have this prop available if we want to manipulate it:
https://github.com/react-navigation/react-navigation/pull/10834

# Test Plan

Please test this intensively before approving or merging, I might have missed something but it was working great in my tests.


alantoa (approved):
LGTM 💯

PR #1852, opened 2023-02-13
fix: myinfo overfetching

intergalacticspacehighway:
# Why
/myinfo is being fetching on every component mount. Thanks to @hirbod for the finding!

# How
- Set revalidateOnMount to false
- Did some cleanup

# Test Plan
- Test myinfo is only fetched when stuff is mutated

### DIFF
--- a/packages/app/hooks/api-hooks.ts
+++ b/packages/app/hooks/api-hooks.ts
@@ -290,12 +290,13 @@ export const useComments = ({ nftId }: { nftId: number }) => {
 
 export const useMyInfo = () => {
   const { accessToken } = useAuth();
-  const queryKey = "/v2/myinfo";
-  const { mutate } = useSWRConfig();
   const navigateToLogin = useNavigateToLogin();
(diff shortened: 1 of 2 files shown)


packages/app/hooks/api-hooks.ts
```diff
@@ -290,12 +290,13 @@ export const useComments = ({ nftId }: { nftId: number }) => {
...
-  const { mutate } = useSWRConfig();
   const navigateToLogin = useNavigateToLogin();
-  const { data, error } = useSWR<MyInfo>(
-    accessToken ? queryKey : null,
-    fetcher
+  const { data, error, mutate } = useSWR<MyInfo>(
+    accessToken ? "/v2/myinfo" : null,
+    fetcher,
+    {
+      revalidateOnMount: false,
```
intergalacticspacehighway, line 298: main fix is here, rest everything is ocd cleanup

hirbod (approved):
Awesome, that should already help quite a lot

PR #1856, opened 2023-02-13
fix: listemptycomponent breaking tabs

hirbod:
# Why

-mt-10 caused list empty component in comments overlap the tabs, making it non clickable.

# How

Removed pointer events.

# Test Plan

Bildschirm­foto 2023-02-13 um 17 33 52

### DIFF
--- a/packages/app/components/comments/index.tsx
+++ b/packages/app/components/comments/index.tsx
@@ -153,10 +153,11 @@ export function Comments({ nft, webListHeight }: CommentsProps) {
   const listEmptyComponent = useCallback(
     () =>
       !isLoading && !error && !dataReversed.length ? (
-        <View tw="absolute -mt-10 h-full w-full flex-1 items-center justify-center">
+        <View tw="absolute h-full w-full flex-1 items-center justify-center">
           <EmptyPlaceholder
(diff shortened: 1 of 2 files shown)


alantoa (commented):
CleanShot 2023-02-14 at 01 00 19
nit: if set `pointer-events-none` this button will be not working when user didn't login

hirbod:
@alantoa good find. I didn't think about that. I will change that to box-none instead of none

hirbod:
@alantoa I chose another solution (using z-index: 1) - wdyt?

Edit: weird, it is still not clickable. There is another issue oO

intergalacticspacehighway (approved):
LGTM ❇️  (with box-none?)

alantoa:
@hirbod maybe you can try set tab bar `z-index:1`

hirbod:
@alantoa I did that, but I still can't click the button inside list empty component

alantoa (approved):
LGTM

alantoa:
@hirbod I just made a small change and fixed this! I'm going to merge and deploy it.

hirbod:
@alantoa LGTM, thanks. Sorry, internet dead today! Thanks for taking over.

PR #1859, opened 2023-02-13
fix: remove nft title and desc escaping

hirbod:
# Why

We need to remove an old hack, since it will be handled from contract.
https://showtime-rq88331.slack.com/archives/C02PXGK3V8D/p1676305085989649

# How

Reverted  https://github.com/showtime-xyz/showtime-frontend/pull/1260
We need to wait before we merge, because @jorgeavaldez needs to address this in the backend first.

https://linear.app/showtime/issue/SHOW2-1527/when-fetching-edition-details-properly-handle-unescaped-strings-in

# Test Plan

### DIFF
--- a/packages/app/hooks/use-drop-nft.ts
+++ b/packages/app/hooks/use-drop-nft.ts
@@ -196,17 +196,11 @@ export const useDropNFT = () => {
         return;
       }
 
-      const escapedTitle = JSON.stringify(params.title).slice(1, -1);
-      const escapedDescription = JSON.stringify(params.description).slice(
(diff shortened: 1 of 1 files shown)


hirbod:
Regarding to https://linear.app/showtime/issue/SHOW2-1527/when-fetching-edition-details-properly-handle-unescaped-strings-in we should now be safe to merge

intergalacticspacehighway (approved):
LGTM ❇️

alantoa (approved):
LGTM

PR #1858, opened 2023-02-13
fix: notification time darkmode

hirbod:
# Why

Notification create date was invisible in dark mode.

# How

Added text-white for dark mode.

# Test Plan

Bildschirm­foto 2023-02-13 um 22 03 54
Bildschirm­foto 2023-02-13 um 22 01 24

PR #1860, opened 2023-02-14
feat: add new notification types

hirbod:
# Why

We have two new notification types. Added those to the notification.
https://showtime-rq88331.slack.com/archives/C02QD3J7DJM/p1675987736390959

Supersedes #1858

# How

Added new notification types (one to update metadata when spotify link is missing and the other when a song was released.
@alantoa @intergalacticspacehighway I think we also need to add new Notification settings. Could someone please take care of that? Not sure if we actually have those already. 

ID 13 = MISSING_MUSIC_RELEASE_METADATA
ID 14 = RELEASE_SAVED_TO_SPOTIFY

I had to change the NotificationItem, because these ones are new and need special handling.

# Test Plan

Bildschirm_foto 2023-02-14 um 02 15 46

Bildschirm­foto 2023-02-14 um 02 23 23

### DIFF
--- a/packages/app/components/notifications/nfts-display-name.tsx
+++ b/packages/app/components/notifications/nfts-display-name.tsx
@@ -52,3 +52,13 @@ export const NFTSDisplayName = ({ nfts }: NotificationDescriptionProps) => {
     </>
   );
 };
+
+export const NFTSDisplayNameText = ({ nfts }: NotificationDescriptionProps) => {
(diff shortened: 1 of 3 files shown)


intergalacticspacehighway (approved):
LGTM ❇️ 🎶

alantoa (approved):
nice! LGTM!

PR #1863, opened 2023-02-14
fix: update presave drop link

intergalacticspacehighway:
# Why
- Allow users to update the song link and release date if the song is presave drop.


alantoa (approved):
LGTM

hirbod:
@intergalacticspacehighway the time check is still there, isn't?

intergalacticspacehighway:
> the time check is still there, isn't?

was trying something, removed it!

PR #1864, opened 2023-02-14
feat: improve date time picker

alantoa:
# Why

- following Isha and Alex's feedback, we should set Release Date to 12:00 AM by default.
- currently, DateTimePicker UI is not good on the web, we need to improve it. 

# How
- modify the default Release Date to tomorrow at 12:00 AM
- add a `date-time-picker.css` to the `DateTimePicker` component on the web. 

# Test Plan

- check the `DateTimePicker` component UI looks good.
- check if the default Release Date field is correctly

https://user-images.githubusercontent.com/37520667/218703835-3452de91-ffff-455b-afd1-1289865fc367.mp4


packages/app/components/drop/drop-music.tsx
```diff
@@ -450,6 +454,7 @@ export const DropMusic = () => {
                       <DateTimePicker
                         disabled={isSaveDrop}
                         onChange={(v) => {
+                          console.log(v);
```
hirbod: nit: can be removed

hirbod (approved):
LGTM

intergalacticspacehighway (approved):
LGTM ❇️ . Looking great 👏

hirbod (approved):

PR #1869, opened 2023-02-15
fix: form draft persistence and file saving support

intergalacticspacehighway:
# Why
- Form persistence is not working. Race condition between default values and the ones set by usePersistForm hook. 

## Before

https://user-images.githubusercontent.com/23293248/218955968-2296c7b7-d3d1-476a-aaa2-edd494827197.mov

## After

https://user-images.githubusercontent.com/23293248/218956195-f44f5684-7012-44f6-b29a-80d4b8f541b2.mov

# How
- Add defaultValue support in usePersistForm hook to prevent the race condition
- Use indexed db to store/retrieve file on web. On native, it already works as file is just an URI.

# Test Plan
- Test form persistence in all drops on native and web

# P.S. 
- I'll be testing this more across browsers
- Also need to fix that flickering.

### DIFF
--- a/packages/app/hooks/use-persist-form.ts
+++ b/packages/app/hooks/use-persist-form.ts
@@ -37,13 +37,15 @@ export const usePersistForm = (
   }: FormPersistConfig
 ) => {
   const watchedValues = watch();
-
-  const clearStorage = () => store.delete(name);
-
   const fileStorage = useMemo(
     () => new FileStorage("persist_db", name),
     [name]
   );
+
+  const clearStorage = () => {
+    store.delete(name);
+    fileStorage.clearStorage();
+  };
 
   useEffect(() => {
     const str = store.getString(name);
--- a/packages/app/hooks/use-persist-form.ts
+++ b/packages/app/hooks/use-persist-form.ts
@@ -33,7 +33,7 @@ export const usePersistForm = (
     dirty = false,
     touch = false,
     onTimeout,
-    timeout,
+    timeout = 86400000,
   }: FormPersistConfig
 ) => {
   const watchedValues = watch();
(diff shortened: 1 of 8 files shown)


packages/app/hooks/use-persist-form.ts
```diff
@@ -24,63 +27,96 @@ export const usePersistForm = (
...
-      ? Object.entries(watchedValues)
-          .filter(([key]) => !exclude.includes(key))
-          .reduce((obj, [key, val]) => Object.assign(obj, { [key]: val }), {})
-      : Object.assign({}, watchedValues);
+    let stringValues: any = {};
+    for (let key in watchedValues) {
+      if (!exclude.includes(key)) {
+        if (watchedValues[key] instanceof File) {
+          fileStorage.saveFile(watchedValues[key], key);
+          stringValues[key] = "instanceof File";
```
intergalacticspacehighway, line 106: doing something smart here 😝 so IndexedDB is called only if file is saved in it.

packages/app/hooks/use-persist-form.ts
```diff
@@ -24,63 +28,96 @@ export const usePersistForm = (
...
+      Object.keys(values).forEach(async (key) => {
         const shouldSet = !exclude.includes(key);
         if (shouldSet) {
-          dataRestored[key] = values[key];
-          setValue(key, values[key], {
-            shouldValidate: validate,
-            shouldDirty: dirty,
-            shouldTouch: touch,
-          });
+          if (values[key] === "instanceof File") {
```
intergalacticspacehighway, line 64: related to that ["smart thing"](https://github.com/showtime-xyz/showtime-frontend/pull/1869/files#r1106729112). It compares the string value

alantoa (approved):
LGTM 🔥 🔥 🔥

PR #1872, opened 2023-02-15
fix: updating expo-av patch from last PR

intergalacticspacehighway:


patches/expo-av+13.0.1.patch
```diff
@@ -1,3 +1,51 @@
...
+diff --git a/node_modules/expo-av/android/build.gradle b/node_modules/expo-av/android/build.gradle
+index 9198d33..49edd82 100644
+--- a/node_modules/expo-av/android/build.gradle
++++ b/node_modules/expo-av/android/build.gradle
+@@ -5,7 +5,7 @@ apply plugin: 'kotlin-android'
+ apply plugin: 'maven-publish'
+ 
+ group = 'host.exp.exponent'
+-version = '13.0.2'
++version = '13.0.1'
```
intergalacticspacehighway, line 27: i don't know why but even clean install, running patch-package with just JS changes is changing the build.gradle file, some post install script is doing some magic. Ignoring for now

PR #1876, opened 2023-02-16
feat: page auth permission on web

alantoa:
# Why

We haven't handled user authentication permission on the router, which allows users to access certain pages that require signing in by directly entering their URLs.

# How

Add the `redirectTo` parameter for the 'useUser' hooks. You can pass this parameter to specify the page you want to redirect to. If the user is not authenticated, they will be redirected to that page. 

# Test Plan 

Please ensure that you are not authenticated and try accessing a page that requires signing in. Check if you are redirected to the login page.

## Before 

https://user-images.githubusercontent.com/37520667/219308516-6f492443-ca7b-4311-ab85-93d9d8685460.mp4

## After 

https://user-images.githubusercontent.com/37520667/219308355-803eafe6-cf61-4955-b928-c49934dc130a.mp4

### DIFF
--- a/packages/app/components/drop/drop-event.tsx
+++ b/packages/app/components/drop/drop-event.tsx
@@ -122,7 +122,7 @@ export const DropEvent = () => {
   const bottomBarHeight = useBottomTabBarHeight();
 
   const { state, dropNFT } = useDropNFT();
-  const user = useUser();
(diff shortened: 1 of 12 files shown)


intergalacticspacehighway (approved):
LGTM ❇️

PR #1877, opened 2023-02-16
fix: rename free drop CTA

alantoa:
# Why

Feedback from Alex: 
Rename “Create Free Drop” to Create Drop because we already write “free digital collectibles”, but the CTA should omit “Free” since now it’s paid for creators


intergalacticspacehighway:
@alantoa we can make it live in paid drops PR.

alantoa:
@intergalacticspacehighway closed it, just found out you already changed it! 😃

PR #1878, opened 2023-02-20
fix: remove email not working and improve error logging

intergalacticspacehighway:
# Why
Remove email is not working. Fix is [here](https://github.com/showtime-xyz/showtime-frontend/compare/fix/email-wallet?expand=1#diff-036fd24d587207a2f630f5e24c7de19f1e9f1f31ecbe07804932a372ea579ee6R66). 

# How
Log error and show backend error messages when available

# Test Plan
- Test add wallet, email and social. Sometimes API returns something went wrong for some users. (for my user, for new user it worked). https://linear.app/showtime/issue/SHOW2-1552/add-wallet-is-throwing-something-went-wrong


hirbod (approved):
LGTM 🔥
Good step forward regarding UX. We should try to address all the important Form related returns which are currently not really helpful in terms of error-handling.

PR #1881, opened 2023-02-21
fix: double page tracking

hirbod:
# Why

Web page tracking has been fired twice on initial load and reload.

# How

Removed the generic `.page` track, since our tracking function already tracks that for us.

# Test Plan


intergalacticspacehighway (approved):
LGTM ❇️

alantoa (approved):
LGTM 👍

PR #1882, opened 2023-02-21
fix: form media persist

intergalacticspacehighway:
# Why
Sometimes media does't persist due to race conditions on setting default value and retrieving from IndexedDB.

# How
- Wait to set default values till media is retrieved from IndexedDB. Max wait for 2 seconds.
- Clear form persisted data on logout

# Test Plan
- Test form persistence.


packages/app/hooks/use-persist-form.ts
```diff
@@ -38,87 +37,133 @@ export const usePersistForm = (
 ) => {
   const watchedValues = watch();
   const fileStorage = useMemo(() => new FileStorage(name), [name]);
-
+  let persistDebounceTimeout = useRef<any>(null);
```
intergalacticspacehighway, line 40: also added debounce when saving to localStorage/mmkv. Before it used to store on each keystroke.

packages/app/hooks/use-persist-form.ts
```diff
@@ -38,87 +37,133 @@ export const usePersistForm = (
...
       }
-      store.set(name, JSON.stringify(stringValues));
     }
-  }, [watchedValues, timeout, exclude, fileStorage, name]);
+
+    if (persistDebounceTimeout.current) {
+      clearTimeout(persistDebounceTimeout.current);
+    }
+    persistDebounceTimeout.current = setTimeout(() => {
+      if (Object.keys(restoringFiles).length === 0) persistFormValues();
```
intergalacticspacehighway, line 159: we only start storing values in localStorage after restoringFiles state becomes empty (can take max [2 seconds](https://github.com/showtime-xyz/showtime-frontend/pull/1882/files#diff-e984ccef2e92f4513764a35a6c62974e4997d0379850b0cb20b1312ee4113cc0R87) in worst case.)

alantoa (approved):
LGTM

PR #1887, opened 2023-02-21
fix: edit/complete profile form UX

hirbod:
# Why

Aligned Edit Profile with our new Form optimizations

# How

Added scroll to error fields like for the drops, specially the social links have been not very clear on small screens in case of an error. Optimized error messages, placeholders and error labels

# Test Plan


intergalacticspacehighway (approved):
LGTM 🔥

PR #1888, opened 2023-02-21
feat: paid drops and billing in settings

intergalacticspacehighway:
# Why
- We're making free drops paid for non verified creators.
- Add billing page in settings screen. This screen will show purchased drop credits

# How
- Added new screens for checkout, select plan, checkout success, and billing (by @alantoa 🐐)
- Integrated stripe checkout, default payment method.

# Testcases
All users should be able to see the billing page in settings screen.

## Native
1. Verified creators can create music and free drops
2. Non verified creators can request for music drop verification. Paid drops button will be invisible on native. 

## Web
1. Same as Native
2. Non verified creators can go through checkout flow to create drops and request for music drop verification.

You can test and share the bugs/feedback on this thread! https://showtime-rq88331.slack.com/archives/C02S5PQ1XJT/p1677014700203479


packages/app/components/checkout/checkout-form.tsx
```diff
@@ -0,0 +1,131 @@
...
+
+    setIsLoading(true);
+
+    const { error } = await stripe.confirmPayment({
+      elements,
+      confirmParams: {
+        return_url:
+          window.location.origin +
+          "/checkout-return?setAsDefaultPaymentMethod=" +
+          setAsDefaultPaymentMethod,
```
hirbod: I know we're not using this form in our native apps, but checkout-form.tsx is not suffixed with `.web.tsx`. In theory, this would be loaded on native. `window.location` is not available. We should either make that an web only component or add some gating with Platform.OS

packages/app/components/checkout/checkout-return.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+    const setAsDefaultPaymentMethod = new URLSearchParams(
+      window.location.search
+    ).get("setAsDefaultPaymentMethod");
+
+    const stripe = await stripePromise;
+
+    const clientSecret = new URLSearchParams(window.location.search).get(
+      "payment_intent_client_secret"
+    );
```
hirbod, line 37: Same issue as pointed out before. Do we already want to make this future proof and define our components as web-only?

hirbod: URLSearchParams are also not available without Polyfill

packages/app/components/checkout/checkout-return.tsx
```diff
@@ -0,0 +1,124 @@
...
+        typeof paymentIntent?.payment_method === "string"
+      ) {
+        setPaymentByDefault(paymentIntent.payment_method);
+      }
+    }
+  }, [router, setPaymentByDefault]);
+
+  useEffect(() => {
+    const paymentIntentId = new URLSearchParams(window.location.search).get(
+      "payment_intent"
```
hirbod, line 55: Same as before

packages/app/components/checkout/select-plan.tsx
```diff
@@ -0,0 +1,255 @@
...
+        method: "POST",
+        url: "/v1/payments/start",
+        data: {
+          payment_plan: arg.dropPlan.name,
+          use_default_payment_method: arg.useDefaultPaymentMethod,
+        },
+      });
+
+      return res;
+    } catch (e: any) {
```
hirbod: e is type of AxiosError

packages/app/components/drop/drop-free.tsx
```diff
@@ -69,10 +71,24 @@ const defaultValues = {
...
 export const DropFree = () => {
   const isDark = useIsDarkMode();
-  const { data: userProfile } = useMyInfo();
-  const maxEditionSize = userProfile?.data?.profile.verified ? 100000 : 50;
+  const { user: userProfile } = useUser();
+  const [checkoutSuccess] = useParam("checkoutSuccess", {
+    parse: (value) => value === "true",
+    initial: false,
+  });
+  const editionSizeCredit =
```
hirbod, line 89: Could you add some code comments what this is?

packages/app/components/drop/drop-music.tsx
```diff
@@ -222,6 +222,17 @@ export const DropMusic = () => {
     previousIsSaveDrop.current = isSaveDrop;
   }, [isSaveDrop, trigger]);
 
+  useEffect(() => {
+    resetDropState();
+  }, [resetDropState]);
```
hirbod, line 231: Is this intended?

packages/app/components/settings/tabs/billing.tsx
```diff
@@ -0,0 +1,266 @@
...
+      </View>
+      <View tw="h-8 flex-row">
+        {item?.receipts?.length > 0 && (
+          <Button
+            variant="tertiary"
+            onPress={() => {
+              for (let link of item.receipts) {
+                Linking.openURL(link);
+              }
+            }}
```
hirbod: I'd rather pass a `useCallback` here and determine `link` in a `useMemo`

hirbod (commented):
I haven't tested yet, but from a code-reviewing perspective, LGTM!
Some questions and tiny nits added.

packages/app/components/checkout/checkout-form.tsx
intergalacticspacehighway: yeah, I considered it. This component is not imported (in root-navigator) so won't be included in bundle. But yeah better to add the suffix. Will do!

packages/app/components/checkout/checkout-return.tsx
intergalacticspacehighway: will do! good catch 🙏

packages/app/components/settings/tabs/billing.tsx
alantoa: yeah, absolutely can optimize a little bit here, but I think using `useCallback` and `useMemo` will make the code more complicated, WDTY?

packages/app/components/drop/drop-music.tsx
intergalacticspacehighway: yes, I have moved the drop creation state to context, so we need to reset it when the component mounts. Moved it to context to prevent some weird issues in redirection from checkout to drop form. Also since one drop is allowed at any moment, I think having a single context can be useful if we decide to show some indicator elsewhere similar to claim for e.g.

packages/app/components/settings/tabs/billing.tsx
hirbod: You might not need useCallback, but I would change the way you retrieve the link. Currently, this would, in theory trigger multiple `openURL` calls. If you want to map or filter, better to store result as useMemo and pass as dep to a useCallback.

intergalacticspacehighway: Fixed

packages/app/components/drop/drop-free.tsx
intergalacticspacehighway: Done

packages/app/components/checkout/select-plan.tsx
intergalacticspacehighway: Done! thanks for catching, i sometimes do this out of laziness. Will take care next time 🥺

packages/app/components/settings/tabs/billing.tsx
```diff
@@ -105,6 +105,12 @@ const HistoryItem = memo(function HistoryItem({
 }: {
   item: PaymentsHistory;
 }) {
+  const handleViewPress = useCallback(() => {
+    for (let link of item.receipts) {
+      Linking.openURL(link);
+    }
```
hirbod: Why are you looping inside a press handler?

packages/app/components/checkout/checkout-form.tsx
intergalacticspacehighway: @hirbod adding some conditionals for now. When we add checkout on native, it'll be a complete rewrite, we'll not use stripe. We can spend more time later cleaning some stuff. Ignoring it for now.

packages/app/components/drop/drop-music.tsx
intergalacticspacehighway: > so we need to reset it when the component mounts. 

Ideally, we can reset it when user closes the success modal but this works for now. Global state has resetting issues 🤷

packages/app/components/checkout/checkout-form.tsx
hirbod: Ok, makes sense.

packages/app/components/settings/tabs/billing.tsx
intergalacticspacehighway: good question. I copy-pasted it without even thinking 🤣

intergalacticspacehighway: cc - @alantoa

alantoa: receipts are an array of URLs because one payment intent may have resulted in multiple charges being made (including refunds most likely) and we get a receipt for each charge.
so we are probably wrong to use only one of them, now we are not sure multiple receipts arrays are not very common, so I think we can use this code first, at least it doesn't appear invisible bug, WDYT?

hirbod: But this would cause multiple openURL triggers. These will do what? Open InAppBrowsers? App Intents? For me this looks wrong. Only ok when these would be something like background fetches

alantoa: yeah, the issue is we don't have background fetches about this right now, and this page is only available on the web, we don't need to consider native.
now I came up with a plan that uses the Dropdown component to handle this, WDTY?

hirbod: I think I don't really understand this from just reading code.

alantoa (approved):
🔥 🔥 🔥

PR #1889, opened 2023-02-22
fix: form default values

intergalacticspacehighway:
# Why
- Form default values are not working properly for first mount

# How
- Move setting default values outside of a condition 🤦 

# Test Plan
Test default values are set in form


alantoa (approved):
LGTM

alantoa:
@intergalacticspacehighway I tested it and work well but now this issue is happening on the product, so I'll merge it and deploy it first!

PR #1893, opened 2023-02-22
fix: clear persisted form data on submit

intergalacticspacehighway:
# Why
- Form data doesn't clear after submit.

# How
- Re-renders messing with equality check of `useEffect`. `watchedValues` and `exclude` were newly created object on each render so when we dispatch success event after drop creation, re-render was causing useEffect to trigger again and ends up saving in localStorage again.
- react-hook-form is the best forms library hands down. They have a way to use imperative subscription on watch(), using that we prevent getting new object a.k.a new `watchedValues` on each render

# Test Plan
Test form gets clear after submission success


hirbod (approved):
LGTM 🔥

PR #1898, opened 2023-02-23
fix: reanimated web without babel plugin

intergalacticspacehighway:
# Why
- `useAnimatedReaction` was not firing on web. 
- We're not using a reanimated babel plugin on web. I think this bug got introduced while experimenting with babel and swc reanimated plugin.

# How
Added patch from this [PR](https://github.com/software-mansion/react-native-reanimated/pull/3997) by Fernando.

# Test Plan
Test video pauses/plays on feed.


alantoa (approved):
LGTM 💯

hirbod (approved):
LGTM

PR #1899, opened 2023-02-24
Feature/show2 1010 add creator name in drop metadata

alantoa:
# Why

just add Creator_Name in the Drop metadata.


intergalacticspacehighway (approved):
LGTM 💯

PR #1901, opened 2023-02-24
feat: improve check receipt

alantoa:
# Why

receipts are an array of URLs because one payment intent may have resulted in multiple charges being made (including refunds most likely) and we get a receipt for each charge, which now maybe cause multiple `openURL` triggers at the same time. 

# How

use the `Dropdown` component to handle this, so: 
- if only one receipt will display `Check Receipt` directly,  for example: 
CleanShot 2023-02-24 at 10 06 52

- if has multiple receipts, will display `Check Receipt #index`, for example:  
CleanShot 2023-02-24 at 10 07 48 

I think this way will be better! WDYT? cc @hirbod

### DIFF
--- a/packages/app/components/settings/tabs/billing.tsx
+++ b/packages/app/components/settings/tabs/billing.tsx
@@ -114,10 +114,6 @@ const HistoryItem = memo(function HistoryItem({
 }: {
   item: PaymentsHistory;
 }) {
-  const handleViewPress = useCallback((link: string) => {
-    Linking.openURL(link);
-  }, []);
-
   return (
     <View tw="flex-row justify-between py-3.5 px-4 md:px-0">
       <View tw="flex-col items-start justify-center md:flex-row md:items-center">
@@ -146,7 +142,7 @@ const HistoryItem = memo(function HistoryItem({
               {item.receipts.map((link, index) => {
                 return (
                   <DropdownMenuItem
-                    onSelect={() => handleViewPress(link)}
+                    onSelect={() => Linking.openURL(link)}
                     key={link}
                   >
                     <MenuItemIcon Icon={Receipt} />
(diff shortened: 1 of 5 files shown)


hirbod:
@alantoa much much much better! LGTM.
Are these PDF files? I would name it `View Receipt` or `Download Receipt` but not `Check` though.

packages/app/components/settings/tabs/billing.tsx
```diff
@@ -128,9 +135,31 @@ const HistoryItem = memo(function HistoryItem({
...
+            <DropdownMenuContent loop>
+              {item.receipts.map((link, index) => {
+                return (
+                  <DropdownMenuItem
+                    onSelect={() => handleViewPress(link)}
+                    key={link}
+                  >
+                    <MenuItemIcon Icon={Receipt} />
+                    <DropdownMenuItemTitle tw="font-semibold text-gray-700 dark:text-neutral-300">
+                      {`Check Receipt ${
```
hirbod:
```suggestion
                      {`View Receipt ${
```

hirbod (changes requested):

packages/app/components/settings/tabs/billing.tsx
```diff
@@ -105,11 +114,9 @@ const HistoryItem = memo(function HistoryItem({
...
   item: PaymentsHistory;
 }) {
-  const handleViewPress = useCallback(() => {
-    for (let link of item.receipts) {
-      Linking.openURL(link);
-    }
-  }, [item]);
+  const handleViewPress = useCallback((link: string) => {
+    Linking.openURL(link);
+  }, []);
```
hirbod: I misunderstood the code before. You are right, useCallback is not needed here and can be inlined

intergalacticspacehighway (approved):
LGTM ❇️

packages/app/components/settings/tabs/billing.tsx
alantoa: DONE!

PR #1902, opened 2023-02-24
fix: Android StatusBar shift on cold boot

hirbod:
# Why

When cold-booting Android, the status bar is not translucent and will be changed declaratively, which causes ugly and expensive layout shifts on boot

# How

Fixing this is not as easy as it seems. There are different results on Android versions less than 10, 10, and 12+, and it gets even worse with Edge-to-Edge mode. Luckily, I've invested a lot of time in this issue in the past and could copy my solution, which just works™. It needs a config plugin that changes something specific caused by Expo.

This should improve initial perf.

# Test Plan

Here is a slow downed video. Left is the current live behavior and right is the fixed one.

https://user-images.githubusercontent.com/504909/221229231-9719e9fb-8e12-44ed-b6fb-161dd2e88dab.mp4

### DIFF
--- a/apps/expo/app.config.js
+++ b/apps/expo/app.config.js
@@ -102,6 +102,10 @@ export default {
     barStyle: "dark-content",
     backgroundColor: "#FFFFFF",
   },
+  androidStatusBar: {
+    backgroundColor: "#00000000",
+    barStyle: "light-content",
+  },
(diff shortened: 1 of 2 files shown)


intergalacticspacehighway (approved):
LGTM ❤️

alantoa (approved):
LGTM 💯

PR #1903, opened 2023-02-27
chore: upgrade Expo SDK 48

hirbod:
# Why

Upgrade to Expo SDK 48

# How

Upgraded all dependencies across the monorepo and fixed version mismatches, removed react-native types, as they are shipped now with RN. Removed some wrong installed deps (expo-modules-core) and some unneeded packages.

iOS already done and working perfectly fine.

## Blockers 

Android is not building currently. Some core packages we need upgrades for:

- react-native-bignumber
- react-native-blob-jsi-helper
- ~~react-native-random-values-jsi-helper~~
- react-native-quick-crypto

~~We could replace `react-native-random-values-jsi-helper` against https://github.com/LinusU/react-native-get-random-values. We could also use `expo-crypto`, but it does not ship with a Polyfill, we would need to add ourselves. I'd wait for SDK 49 before opting for expo-crypto~~

I replaced it with LinusU's version. Should not have any perf impact for us.

#### Following patch is broken and needs an update once blob-jsi-helper has 0.71 support for Android:
- react-native+0.70.5.patch

#### Following patches were successfully applied for on old version - working but need a bump (I could not, patch-package crashes for me)
- npx patch-package expo-notifications
- npx patch-package react-native-fast-image

# Test Plan

- [x] iOS
- [ ] Android

@intergalacticspacehighway can you take over once we got more infos regarding the failing packages?


alantoa:
@hirbod looks like it related to `yarn.lock` file, could you go to `root` folder and run `yarn` command?

hirbod:
@alantoa I did but no change. Might be really related to the missing RN patch.

alantoa:
@hirbod got it, looking into it!

hirbod:
Closing in favor of #2061

PR #1905, opened 2023-02-28
fix: list is not recycle

alantoa:
# Why

https://showtime-rq88331.slack.com/archives/C02Q0E4PBD0/p1677499269539919 

# How 

change list render layout. 

# Test Plan

check if the list is performance better for now.

### DIFF
--- a/packages/app/components/profile/profile-tab-list.tsx
+++ b/packages/app/components/profile/profile-tab-list.tsx
@@ -86,7 +86,6 @@ export const ProfileTabList = forwardRef<ProfileTabListRef, TabListProps>(
       }: ListRenderItemInfo<NFT & { loading?: boolean }>) => {
         return (
           <Card
-            key={item.nft_id}
             nft={item}
             onPress={() => onItemPress(itemIndex)}
             numColumns={NUM_COLUMNS}
--- a/packages/app/components/profile/profile-tab-list.tsx
+++ b/packages/app/components/profile/profile-tab-list.tsx
@@ -63,6 +63,14 @@ export const ProfileTabList = forwardRef<ProfileTabListRef, TabListProps>(
     useImperativeHandle(ref, () => ({
       refresh,
     }));
+    const onItemPress = useCallback(
+      (currentIndex: number) => {
+        router.push(
+          `/list?initialScrollIndex=${currentIndex}&tabType=${list.type}&profileId=${profileId}&collectionId=${filter.collectionId}&sortType=${filter.sortType}&type=profile`
+        );
+      },
+      [list.type, profileId, filter.collectionId, filter.sortType, router]
+    );
 
     const ListFooterComponent = useCallback(
       () => <ProfileFooter isLoading={isLoadingMore} />,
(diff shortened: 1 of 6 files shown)


packages/app/components/trending/tabs/nfts-list.tsx
```diff
@@ -1,33 +1,19 @@
...
 import { ListFooter } from "app/components/footer/list-footer";
-import { withViewabilityInfiniteScrollList } from "app/hocs/with-viewability-infinite-scroll-list";
 import { useTrendingNFTS } from "app/hooks/api-hooks";
 import { useContentWidth } from "app/hooks/use-content-width";
 import { useScrollToTop } from "app/lib/react-navigation/native";
 import { NFT } from "app/types";
 
 import { TrendingTabListProps, TrendingTabListRef } from "./";
 
-const ViewabilityInfiniteScrollList = withViewabilityInfiniteScrollList(
```
alantoa, line 27: remove this, I think we shouldn‘t need this in the Trending list.

hirbod: I think that's wrong. I have seen that we're playing videos in web, even in the trending list. So it might have its purpose.

packages/app/components/profile/profile-tab-list.tsx
```diff
@@ -86,33 +77,19 @@ export const ProfileTabList = forwardRef<ProfileTabListRef, TabListProps>(
...
-                  marginRight: StyleSheet.hairlineWidth,
-                  marginBottom: StyleSheet.hairlineWidth,
-                }}
-              />
-            ))}
-          </View>
+          <Card
+            nft={item}
+            onPress={() => onItemPress(itemIndex)}
+            numColumns={NUM_COLUMNS}
```
hirbod, line 91: It's preferred to `useCallback` here.
Define onPress as a const on line 87.

packages/app/components/trending/tabs/nfts-list.tsx
```diff
@@ -59,36 +41,23 @@ export const NFTSList = forwardRef<TrendingTabListRef, TrendingTabListProps>(
...
-                  marginBottom: StyleSheet.hairlineWidth,
-                }}
-              />
-            ))}
-          </View>
+          <Card
+            nft={item}
+            onPress={() => onItemPress(index)}
+            numColumns={NUM_COLUMNS}
+          />
```
hirbod, line 50: It's preferred to `useCallback` here.
Define onPress as a const with useCallback on line 45.

hirbod (changes requested):

packages/app/components/trending/tabs/nfts-list.tsx
alantoa: yeah, I noticed that, but looks like we didn't using this on profile list. 😃

alantoa: anyways, adding it at least should be better I think.

alantoa: haha, just I found this list didn't use on the web, accurately now using [this file](https://github.com/showtime-xyz/showtime-frontend/blob/staging/packages/app/components/trending/trending.web.tsx) on the web.

hirbod: Oh ok :)

hirbod (approved):
LGTM 🔥

PR #1909, opened 2023-03-01
fix: cap drop editions to 350

alantoa:
# Why

[Verified Creator Free Drop: cap at 350 editions maximum](https://linear.app/showtime/issue/SHOW2-1576/verified-creator-free-drop-cap-at-350-editions-maximum)


intergalacticspacehighway (approved):
LGTM ❇️

PR #1910, opened 2023-03-01
fix: carousel pagination color

alantoa:
# Why 

fix: slider dots do have not enough contrast.

## Before 
image

## After 
CleanShot 2023-03-01 at 6 41 48
CleanShot 2023-03-01 at 6 41 43


hirbod (approved):
🔥

PR #1912, opened 2023-03-02
fix: complete prompt screen

intergalacticspacehighway:
# Why
- Complete prompt screen stays fixed after successful signup. This was breaking test on maestro.
- I think we don't need a separate screen and we could just toggle the complete prompt content with state, similar to drop pages.

# How
- Replace Complete prompt screen with a component in edit profile screen

# Test Plan
Test e2e passes and complete prompt screen is not visible after successful sign up.


packages/app/providers/auth-provider.tsx
```diff
@@ -99,7 +99,7 @@ export function AuthProvider({
             router.pop();
           }
           setTimeout(() => {
-            router.push("/profile/complete-prompt");
```
alantoa, line 102: sometimes the `isAuthenticated` is `false` even if the sign-in API was successful, so will redirect to the login modal when `1000ms` time out to enter the `edit-profile` screen. it's related to [this issue](https://showtime-rq88331.slack.com/archives/C02PXGK3V8D/p1677594471875929). 

btw, IMO we can remove this line, I feel it's a little bit weird for the user because the pop-up is without any prompting.
or we use `toast` instead it?

alantoa (approved):
LGTM ✨

packages/app/providers/auth-provider.tsx
intergalacticspacehighway: yeah, let's revisit when we add the new flow!

hirbod: Yeah, the new flow overwrites ALL of this.

PR #1914, opened 2023-03-03
feat: improve toast

alantoa:
# Why

https://showtime-rq88331.slack.com/archives/C02Q0E4PBD0/p1677755775617289 

# How 

- on the native is using [Burnt](https://github.com/nandorojo/burnt)
- on the web is using [react-hot-toast](https://github.com/timolins/react-hot-toast) 
- make the web and native have the same APIs. 
- clear up the storybook example. 
- change all used toast files.

# Test Plan

check the toast message if is correct since I changed a lot of files and maybe missed some copies.  

https://user-images.githubusercontent.com/37520667/222733749-abf5c59a-7a86-4eb3-94cb-23e9cd4d88e2.mp4


packages/app/hooks/use-block.ts
```diff
@@ -57,9 +57,9 @@ function useBlock() {
         }),
         true
       );
-      toast?.show({ message: "Blocked!", hideAfter: 4000 });
+      toast("Blocked!");
```
intergalacticspacehighway, line 60: is this API valid? `toast(message)`

intergalacticspacehighway (approved):
LGTM 😍

packages/app/hooks/use-block.ts
alantoa: yeah, if this method will be without any icon. 
I just think that the `blocked` and `reported` is a little bit negative, so I didn't add any positive icons to the above. 😃

PR #1917, opened 2023-03-06
fix: toast duration & memorize toast options

alantoa:
# Why

- fix toast duration since the duration is `s` instead of `ms`. 🤣  
- memorize the toast options will be better. 

# How

# Test Plan


intergalacticspacehighway (approved):
LGTM. LOL

PR #1920, opened 2023-03-06
Feature/show2 1553 settings moved privacy and security

alantoa:
# Why  

- we need to move the "Privacy and Security", `Clear cache` and `Delete account` to a new tab "Advanced" that will be clear.  
- add a feedback button to the `Advanced` tab and support the user with feedback any issue or question will be good. 

# How
- add a new `Advanced` tab to the Setting screen. 
- use `expo-mail-composer` lib to support preset text copy to the Mail app, if failed will use the `Alert` component to prompt users to copy their email address manually.
 

# Test Plan

check if the feedback button works.


intergalacticspacehighway (approved):
LGTM ❇️. wow `expo-mail-composer` 💯

PR #1921, opened 2023-03-07
fix: crop image cannot view button

alantoa:
# Why
fix some feedback from users on the mobile web:  

- user cannot view the modal header in the edit-profile modal.
- user cannot view the `Looks good` button on the crop image modal.  

screenshot just like 

this should be due to the modal too height causing that.

# How

- remove the `Open in App` banner in the `profile/*` screen since going to the user profile anyway will redirect to `showtime.xyz/@username`,  so we don't need to consider including the `profile/*` screen. 
- reduce modal height.  

# Test Plan

I need to deploy on the product to test it since local is without the `Open in App` banner.

### DIFF
--- a/packages/app/components/media-cropper/index.web.tsx
+++ b/packages/app/components/media-cropper/index.web.tsx
@@ -95,7 +95,7 @@ export const MediaCropper = ({
             )}
           />
           <View tw="max-h-[90vh] min-h-[560px]">
-            <View tw="h-[60vh] w-full md:w-[480px]">
+            <View tw="h-[60vh] min-h-[480px] w-full md:w-[480px]">
               {src && (
                 <Cropper
                   image={src}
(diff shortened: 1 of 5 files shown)


packages/app/components/media-cropper/index.web.tsx
```diff
@@ -95,7 +95,7 @@ export const MediaCropper = ({
             )}
           />
           <View tw="max-h-[90vh] min-h-[560px]">
-            <View tw="h-[576px] w-full md:w-[480px]">
+            <View tw="h-[60vh] w-full md:w-[480px]">
```
hirbod: Do you want to add a min-height just in case?

hirbod (commented):
LGTM

packages/app/components/media-cropper/index.web.tsx
alantoa: yeah, that makes sense,  min-height should be the same as the `width`.

intergalacticspacehighway (approved):
LGTM ❇️

PR #1923, opened 2023-03-07
fix: collect button text color on music drop

alantoa:
# Why

The music drop's collect button text color is not correct in dark mode. 

## Before 

CleanShot 2023-03-07 at 8 46 44

## After 

CleanShot 2023-03-07 at 8 47 14


alantoa:
merged it first to fix that since it this bug is already in the product.

PR #1924, opened 2023-03-07
fix: collect button text

alantoa:
# Why

The text on the collect button cannot be fully displayed. 
# Before 

CleanShot 2023-03-08 at 4 55 54
# After 
CleanShot 2023-03-08 at 4 57 30


hirbod (approved):
LGTM

PR #1925, opened 2023-03-08
feat: new onboarding

hirbod:
# Why

This PR brings the long awaited new onboarding flow.

# How

This is a big PR, bear with me, but it was very tangled across the app and plenty of tiny changes have been necessary. @intergalacticspacehighway, you are free to add the Twitter/Insta to select-social.tsx. We can have a little chat about that on Slack.

# Test Plan


apps/expo/package.json
```diff
@@ -194,6 +195,7 @@
     "react-native-level-fs": "^3.0.0",
     "react-native-localhost": "^1.0.0",
     "react-native-mmkv": "2.5.1",
+    "react-native-modal": "^13.0.1",
```
intergalacticspacehighway, line 200: are we using react-native-modal anywhere?

hirbod: Yes, hCaptcha has this dependency. But they haven’t released a new version on npm and I had to import from git directly and this forced me to install this dep

packages/app/components/onboarding/hcaptcha/challenge.web.tsx
```diff
@@ -0,0 +1,157 @@
...
+      );
+    } finally {
+      // this has to be called to reset the captcha once validated with the server
+      captchaRef.current?.resetCaptcha();
+      setChallengeIsRunning(false);
+    }
+  };
+
+  const onExpire = () => {
+    console.log("hCaptcha Token Expired");
```
hirbod: could be inlined, it is a no-op but we have to define it, because the lib is calling it anyway..

packages/app/components/onboarding/hcaptcha/challenge.web.tsx
```diff
@@ -0,0 +1,157 @@
...
+  };
+
+  const onError = (err: string) => {
+    console.log(`hCaptcha Error: ${err}`);
```
hirbod: could be inlined, it is a no-op but we have to define it, because the lib is calling it anyway..

packages/app/components/onboarding/hcaptcha/challenge.web.tsx
```diff
@@ -0,0 +1,157 @@
...
+  };
+
+  // hCaptcha injects a very ugly square on the page when the challenge is running
+  // this function removes it even though its super hacky
+  const fixUglySquare = () => {
+    const sq = document.querySelector('div[style*="border-width: 11px" i]');
+    if (sq) sq.remove();
+  };
```
hirbod, line 112: I absolutely hate this, since the DOM can change, but there was no other way to prevent that weird square (I think they tried to build an X button, lol). So this needs to stay unless you guys know a better way.

packages/app/components/onboarding/hcaptcha/hcaptcha-utils.ts
```diff
@@ -0,0 +1,50 @@
...
+  return { validate };
+};
+
+export const useFinishOnboarding = () => {
+  const router = useRouter();
+  const redirectToClaimDrop = useRedirectToClaimDrop();
+
+  const finishOnboarding = useCallback(
+    (redirectUri?: string) => {
+      console.log(redirectUri);
```
intergalacticspacehighway: nit: log

packages/app/components/onboarding/hcaptcha/hcaptcha-utils.ts
```diff
@@ -0,0 +1,50 @@
...
+      if (redirectUri) {
+        if (redirectUri.includes("/claim/")) {
```
intergalacticspacehighway: not a blocker: do you think we can handle it better if we implement that promise based approach?

await finishOnboarding() 😄

packages/app/components/onboarding/onboarding.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+export const Onboarding = () => {
+  const { user, isIncompletedProfile } = useUser();
+  const router = useRouter();
+
+  const [redirectUri] = useParam("redirectUri");
+
+  // redirect to home if user is not incompleted profile
+  // this should prevent user from going back to onboarding
+  // with the back button
```
intergalacticspacehighway: i think if we do router.replace post onboarding, we might not this special handling. wdyt?

packages/app/components/onboarding/select-picture.tsx
```diff
@@ -0,0 +1,300 @@
...
+
+          setStep(OnboardingStep.Social);
+
+          // TODO: optimise to make fewer API calls!
+          mutate(MY_INFO_ENDPOINT);
+          matchMutate(
+            (key) => typeof key === "string" && key.includes(USER_PROFILE_KEY)
+          );
+        } catch (e) {
+          setError("profilePicture", {
```
intergalacticspacehighway, line 110: we could also Logger.error here

intergalacticspacehighway (approved):
Added some nits, but overall this looks great 👏

packages/app/components/onboarding/hcaptcha/challenge.web.tsx
hirbod: We could patch the library though. WDYT? Running this little hack or going for a patch?

packages/app/components/onboarding/hcaptcha/hcaptcha-utils.ts
hirbod: I think we should definitely add that approach soon. We could tackle this in one rush and add the same for liking, clicking on comments etc. as we talked in our huddle. For now, leaving as is?

packages/app/components/onboarding/onboarding.tsx
hirbod: There are cases though which require router.pop, like the claim. Main reason was that I was calling an existing hook which used `push` to open the claim dialog. I could duplicate that hook though and make a replace version, or pass a param. WDYT? 

To be specific, it is `useRedirectToClaimDrop`

packages/app/components/onboarding/select-picture.tsx
hirbod: Done

packages/app/components/onboarding/hcaptcha/hcaptcha-utils.ts
hirbod: Done

packages/app/components/onboarding/hcaptcha/challenge.web.tsx
hirbod: Done, using a noop now defined outside the component

packages/app/components/onboarding/hcaptcha/challenge.web.tsx
hirbod: Done, using a noop now defined outside the component

packages/app/components/onboarding/onboarding.tsx
hirbod: I added a second param `shouldUseReplace` to useRedirectToClaimDrop (default false) so I can reuse that hook and get rid of that special handling

hirbod: Even though I added this this logic https://github.com/showtime-xyz/showtime-frontend/pull/1925/commits/26755351e6ccaefe5a2a35e19042a7f4c54f298e#diff-20211b750f4af859e9d85d45e05e02fc483793752948b7b525cb18c2bf48abcbR46-R56 it still pushes on top instead of replacing it. Very odd

hirbod: I found another solution, so this is resolved now

hirbod: I thought about it once more and I decided to add this back, because even when we replace, users on web could still directly call /onboarding

packages/app/components/onboarding/hcaptcha/challenge.web.tsx
intergalacticspacehighway: can we do it with CSS though? some namespaced selector or something more unique?

packages/app/components/onboarding/onboarding.tsx
intergalacticspacehighway: okay, let's keep it 👍

packages/app/components/onboarding/hcaptcha/challenge.web.tsx
hirbod: Good luck 🍀 😂. This was the most unique selector I could find lol. No detectable hierarchy, no class names etc.

intergalacticspacehighway: leave it for now! 😂

packages/app/components/onboarding/hcaptcha/challenge.web.tsx
```diff
@@ -0,0 +1,151 @@
...
+        size="invisible"
+        onError={noop}
+        onExpire={noop}
+        onVerify={noop}
+        theme={isDark ? "dark" : "light"}
+        languageOverride="en"
+      />
+      {challengeRunning && (
+        <Portal.Root>
+          <View
```
alantoa, line 142: not a blocker, i prefer to this code style. 😃 
```
<View 
tw="items-center justify-center bg-black/60 dark:bg-black/60"    
style={StyleSheet.absoluteFill}        
pointerEvents="none"
/>
```

packages/app/components/onboarding/hcaptcha/index.web.ts
```diff
@@ -0,0 +1 @@
+export { Challenge } from "./challenge.web";
```
alantoa: we can delete this file since anyway the `hcaptcha/index.ts` file will import the `challenge.web` file on web.

alantoa (approved):
LGTM! 🔥 🔥 🔥 
I left some comments on Slack.

PR #1929, opened 2023-03-08
feat: twitter/instagram connect in new onboarding flow

intergalacticspacehighway:
# Why
Right now printing the token in console. We need to do API call there!

# Test Plan
Test Instagram connect and Twitter on onboarding screen. It should log tokens on console


hirbod (approved):
Very nice, LGTM. I'll add a different button state in case a user already has a connection and disable that specific button. We can add that later with the server step.

intergalacticspacehighway:
@hirbod will merge when API endpoints are up. Would be easier to review.

alantoa (approved):
LGTM

PR #1930, opened 2023-03-09
fix: remove divider border

alantoa:
# Why 

we need to remove the line `Separator` on list, It will looks more modern. 

## Before 

CleanShot 2023-03-09 at 10 38 39

## After

CleanShot 2023-03-09 at 10 38 57


intergalacticspacehighway (approved):
Nice one. LGTM ❇️

PR #1931, opened 2023-03-09
fix: message text cannot select

alantoa:
# Why

currently, the comments text cannot select. we need to make it can be selected copy.

# How

instead of using the Swipeable component, we will use View on the web.


intergalacticspacehighway (approved):
LGTM ❇️

hirbod:
Was this an actual request to make it selectable?

PR #1932, opened 2023-03-10
feat: login promise

intergalacticspacehighway:
# Why
- Adds a way to continue user-initiated authenticated actions after the user is authenticated.
- Realised it would be better to add it before doing Spotify decoupling PR as it would need it for better UX.

```
// User taps on authenticated action e.g. like
await loginPromise()
// continue with the action
```

https://user-images.githubusercontent.com/23293248/224194391-4b7eec06-3701-456f-9685-c843598780be.mov

# How
- Some promise hacking

# Test Plan
- Try liking, following when you are not logged in. Action should complete post login.

P.S. for these actions we might also want to add `await onboarding()`. Not a blocker for now.

### DIFF
--- a/packages/app/lib/login-promise/index.ts
+++ b/packages/app/lib/login-promise/index.ts
@@ -1,0 +1,29 @@
+import { useAuth } from "app/hooks/auth/use-auth";
+import { useStableCallback } from "app/hooks/use-stable-callback";
+import { useNavigateToLogin } from "app/navigation/use-navigate-to";
+
+export let loginPromiseCallbacks = {
+  resolve: null as ((v: unknown) => void) | null,
+  reject: null as ((v: unknown) => void) | null,
+};
+
+export const useLogInPromise = () => {
+  const { accessToken } = useAuth();
+  const navigateToLogin = useNavigateToLogin();
+  const loginPromise = useStableCallback(
+    () =>
+      new Promise((resolve, reject) => {
+        if (accessToken) {
+          resolve(true);
+        } else {
+          navigateToLogin();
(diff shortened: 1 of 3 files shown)


packages/app/lib/login-promise/index.ts
```diff
@@ -0,0 +1,29 @@
...
+import { useStableCallback } from "app/hooks/use-stable-callback";
+import { useNavigateToLogin } from "app/navigation/use-navigate-to";
+
+export let loginPromiseCallbacks = {
+  resolve: null as ((v: unknown) => void) | null,
+  reject: null as ((v: unknown) => void) | null,
+};
+
+export const useLogInPromise = () => {
+  const { accessToken } = useAuth();
```
intergalacticspacehighway, line 11: we use useAuth instead of useUser because of some context order issues. UserProvider is down in the hierarchy. I don't like context due to these reasons 😤 . This hook is used in useMyInfo. which I think is above the user provider at some place. It's a bit weird right now. Will spend some time cleaning it up.

packages/app/providers/auth-provider.tsx
```diff
@@ -217,6 +218,13 @@ export function AuthProvider({
     };
   }, [doRefreshToken]);
 
+  useEffect(() => {
+    if (authenticationStatus === "AUTHENTICATED") {
```
intergalacticspacehighway, line 222: we call resolve here instead of login handler. Not sure if it's a good idea. Looks safe for now.

hirbod (approved):
Love it, works great! Same integration done for Onboarding. We can chain the flow now once this is merged and remove some if/else conditions on createDrop and claim

PR #1937, opened 2023-03-13
fix: integrate social endpoints

intergalacticspacehighway:
# Why
Adds new Instagram/Twitter connect API in onboarding. Backend PR is [here](https://github.com/showtime-xyz/stbackend/pull/760)

# Test Plan
- Test Social connect in login and settings page

### DIFF
--- a/packages/app/components/onboarding/select-social.tsx
+++ b/packages/app/components/onboarding/select-social.tsx
@@ -8,7 +8,7 @@ import { View } from "@showtime-xyz/universal.view";
 
 import { useAddMagicSocialAccount } from "app/hooks/use-add-magic-social-account";
 
-import { Instagram, Twitter } from "design-system/icon";
+import { InstagramColorful, Twitter } from "design-system/icon";
 import { Spinner } from "design-system/spinner";
 
 import { Challenge } from "./hcaptcha";
@@ -57,7 +57,7 @@ export const SelectSocial = () => {
           />
           <ConnectButton
             title="Connect Instagram"
-            icon={<Instagram color="#4A99E9" width={20} height={20} />}
+            icon={<InstagramColorful color="#4A99E9" width={20} height={20} />}
             type="instagram"
           />
 
--- a/packages/app/components/onboarding/select-social.tsx
+++ b/packages/app/components/onboarding/select-social.tsx
@@ -12,6 +12,7 @@ import { Instagram, Twitter } from "design-system/icon";
 import { Spinner } from "design-system/spinner";
 
 import { Challenge } from "./hcaptcha";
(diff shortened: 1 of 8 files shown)


packages/app/components/onboarding/select-social.tsx
```diff
@@ -64,35 +50,60 @@ export const SelectSocial = () => {
...
+          <ConnectButton
+            title="Connect Twitter"
+            icon={<Twitter color="#4A99E9" width={20} height={20} />}
+            type="twitter"
+          />
+          <ConnectButton
+            title="Connect Instagram"
+            icon={<Instagram color="#4A99E9" width={20} height={20} />}
+            type="instagram"
+          />
```
hirbod: The color here is wrong. We have a new `InstagramColorful` Icon which does not need any color.

packages/app/components/settings/tabs/account.tsx
```diff
@@ -76,7 +85,45 @@ const ConnectSpotify = ({ redirectUri }: ConnectSocialProps) => {
...
+          }
+        }}
+      >
+        {user.user?.data.profile.social_login_connections.instagram
+          ? "Disconnect"
+          : "Connect"}
+      </Button>
+    </View>
+  );
+};
```
hirbod, line 126: For the profile, we need the tiny buttons + state (EditProfile.tsx)

And for the connect buttons, the layout should look like so:

Screenshot 2023-03-12 at 11 55 37 PM

hirbod (changes requested):
LGTM, the Buttons need some changes though.

packages/app/components/settings/tabs/account.tsx
hirbod: FYI: screenshot only one color mode. Inverted colors for dark/light

packages/app/components/onboarding/select-social.tsx
intergalacticspacehighway: Done, thanks!

packages/app/components/settings/tabs/account.tsx
intergalacticspacehighway: do we need to show the connected state? we show the toast and the router pops.

intergalacticspacehighway: merging this for now. Can do in a follow up PR if required!

hirbod: Yeah, Alex asked for "connected". We can do a follow up PR. It can be states where the user signed up with Twitter already. So it needs to reflect that (make it also disabled in the onboarding) so the user can only pick up Instagram or Skip

PR #1942, opened 2023-03-15
feat: music drop/auth decoupling

intergalacticspacehighway:
# Why
Allow user to save the song before sign up.

# How
Use loginPromise, onboardingPromise to do trigger saving spotify token. Follows [backend spec](https://www.notion.so/showtime-xyz/Spotify-de-coupling-5c5c71d6e194459587dcad8c0b9711f1)

# Test Plan
- Create a new wallet account
- Try to claim this drop while you are signed out. http://localhost:3000/nft/mumbai/0x71778d2126CBbaf8BaaB2E22127E9bddf5c821fA/0

### DIFF
--- a/apps/next/src/pages/login.tsx
+++ b/apps/next/src/pages/login.tsx
@@ -1,1 +1,3 @@
-export { default } from "app/pages/login";
+import { Fragment } from "react";
+
+export default Fragment;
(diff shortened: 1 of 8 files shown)


alantoa (approved):
LGTM 🔥 🔥 
I just tested it and works well!

PR #1946, opened 2023-03-15
fix: login stays open sometimes

intergalacticspacehighway:
# Why
- Noticed during spotify decoupled flow, login modal stayed open.
### Root cause
During login we do `router.push(/login)`. router.pop was being done in a callback that would be called in a useEffect post login. Leading to weird race conditions. 

# How
We make sure to call router.pop in login() function synchronously instead of useEffect. This makes sure it works well with our promise based await loginPromise(); await onboardingPromise() setup.

# Test Plan
- Test music drop decoupled flow

### DIFF
--- a/packages/app/components/login/index.tsx
+++ b/packages/app/components/login/index.tsx
@@ -11,11 +11,7 @@ import { BottomSheetScrollView } from "app/components/bottom-sheet-scroll-view";
 import { LoginComponent } from "./login";
 import { useLogin } from "./use-login";
 
-interface LoginProps {
-  onLogin?: () => void;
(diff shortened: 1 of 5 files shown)


alantoa (approved):
LGTM! 
do we need to rebuild the native app?

hirbod (approved):
🔥

intergalacticspacehighway:
> do we need to rebuild the native app?

we can do a codepush!

hirbod:
> we can do a codepush!

I think we can't. Stage has some new native libraries. Maybe just cherry picking

PR #1950, opened 2023-03-15
improve deleting account

alantoa:
# Why 

- we should use Alert component to prompt users to double-check before deleting their account.
- we should prompt the status when a user's account deletion succeeds or fails.

CleanShot 2023-03-16 at 2 40 26

# How

# Test Plan

### DIFF
--- a/packages/app/components/settings/settings-account-item.tsx
+++ b/packages/app/components/settings/settings-account-item.tsx
@@ -16,6 +16,8 @@ import { useDeleteUser } from "app/hooks/use-delete-user";
 import { useUser } from "app/hooks/use-user";
 import { Logger } from "app/lib/logger";
 
+import { toast } from "design-system/toast";
+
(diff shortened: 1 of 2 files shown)


packages/app/components/settings/settings-account-item.tsx
```diff
@@ -28,14 +31,24 @@ export const SettingDeleteAccount = () => {
...
     }
-
-    try {
-      await deleteUser();
-      logout();
-    } catch (e) {
-      Logger.error(e);
-      setError("Error deleting account");
-    }
+    Alert.alert("Delete Account", "This action cannot be undone.", [
```
hirbod:
```suggestion
    Alert.alert("Delete Account", "Are you sure you want to delete your account? This action cannot be undone.", [
```

hirbod (approved):

PR #1952, opened 2023-03-16
fix: sign up maestro test

intergalacticspacehighway:
# Why
- Fixes new onboarding sign up test. 
- This test is not 100% perfect as we skip the captcha, but not sure what we can do about it. We can potentially create a dummy insta or twitter account for testing 🤔

# How
Updated maestro sign up test flow

# Test Plan
Test everything passes on maestro

### DIFF
--- a/packages/app/components/onboarding/hcaptcha/challenge.tsx
+++ b/packages/app/components/onboarding/hcaptcha/challenge.tsx
@@ -36,6 +36,12 @@ export const Challenge = () => {
   const captchaRef = useRef<ConfirmHcaptcha>(null);
 
   const showCaptcha = () => {
+    // we don't need captcha on e2e tests
+    if (process.env.E2E) {
+      finishOnboarding();
+      return;
+    }
+
     // skip directly to the next step if user has already a social account
     // connected or if the user has already completed the captcha challenge
     if (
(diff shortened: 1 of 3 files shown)


packages/app/components/onboarding/hcaptcha/challenge.tsx
```diff
@@ -36,6 +36,12 @@ export const Challenge = () => {
   const captchaRef = useRef<ConfirmHcaptcha>(null);
 
   const showCaptcha = () => {
+    // we don't need captcha on e2e tests
+    if (process.env.E2E) {
```
intergalacticspacehighway, line 40: we should avoid changing code for tests. But I couldn't find any better way to do this. We can end the test on captcha screen though or create a dummy twitter/insta accounts.

hirbod: There is another way: hCaptcha has a "dummy" sitekey which always results to true, we could use that maybe?

hirbod: But LGTM anyway

hirbod (approved):

packages/app/components/onboarding/hcaptcha/challenge.tsx
intergalacticspacehighway: oh wow. There is? sure we should use that!

alantoa (approved):
LGTM

PR #1953, opened 2023-03-16
feat: create drop testcase

intergalacticspacehighway:
# Why
Adds create drop testcase for iOS. Android still lacks media picker. Maestro team just mentioned they are working on the feature.

# Test Plan
Test should pass on maestro


alantoa (approved):
LGTM 🔥

hirbod (approved):
LGTM

PR #1955, opened 2023-03-17
fix: pass token in login APIs

intergalacticspacehighway:
# Why
- We can pass `provider_access_token` in login APIs. 
- Also did some small refactor in web login with wallet. Made it more similar to native.

# Test Plan
- Test login with wallet works on web.


alantoa (approved):
LGTM! good job! 🔥

PR #1956, opened 2023-03-17
fix: video preview 

alantoa:
# Why 

video preview border radius is not correct.  

# How  

simplfy it and reduce an image use. 

## Before  

CleanShot 2023-03-18 at 1 49 43

## After 
CleanShot 2023-03-18 at 2 16 29

# How

# Test Plan

still need to wait for test some video content.


hirbod (approved):
LGTM

intergalacticspacehighway (approved):
LGTM ❇️

PR #1958, opened 2023-03-19
fix: follow amplitude naming settings

hirbod:
# Why

# How

# Test Plan


hirbod:
tiny change, just to comply with amplitude naming convention

PR #1959, opened 2023-03-19
fix: enforce captcha gate to fix mismatch with backend

hirbod:
# Why

The gating was wrong and not in-line with the backend.
We had hundreds of issues raised in our Sentry logs, because the frontend logic did not align to https://github.com/showtime-xyz/stbackend/commit/a1b00f612cebc5499a050c4b2a8f7c21c9ee42d3#diff-ea7a8b8d5961cb00e2715e7572e49422b5ed37b7a328c4a9c77e2b66a90e3d46R19-R30

Our issue was that we focused on Instagram and Twitter connections and did not test the fact with a skipped hCaptcha

# How

We will enforce the onboarding again for users without Instagram or Twitter, since only checking for SocialAccounts is not enough. Before, we did not show hCaptcha to users as soon as they had a username, which caused a wrong state.

# Test Plan

### DIFF
--- a/packages/app/components/onboarding/hcaptcha/challenge.tsx
+++ b/packages/app/components/onboarding/hcaptcha/challenge.tsx
@@ -36,6 +36,9 @@ export const Challenge = () => {
   const captchaRef = useRef<ConfirmHcaptcha>(null);
 
   const showCaptcha = () => {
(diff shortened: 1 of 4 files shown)


alantoa (approved):
LGTM 🙏

PR #1960, opened 2023-03-19
fix: instagram connect/disconnect

hirbod:
# Why

The Instagram connect/disconnect was disappearing from the settings due to test leftovers.

# How

Removed condition, optimized check in component.

# Test Plan


alantoa (approved):
LGTM 🙏

hirbod:
Sometimes, `instagramProviderId` is undefined. Guess we need to mutate accounts. Not a blocker, we should check it though.

PR #1961, opened 2023-03-20
fix: remove alert on spotify cancel

intergalacticspacehighway:
# Why
Remove alert if user presses cancel on spotify connect popup


alantoa (approved):
LGTM

PR #1962, opened 2023-03-20
fix: scroll position not preserved when opening modals

intergalacticspacehighway:
# Why
Scroll position is not preserved when modal is openend in feed.

https://user-images.githubusercontent.com/23293248/226291680-27986fe7-030b-4668-8292-01e3819601cc.mov

# How
Improved logic to lock body scroll

# Test Plan
Scroll position is preserved when modal is opened


alantoa (approved):
LGTM ✨

PR #1963, opened 2023-03-20
fix: dont refresh without refreshToken

hirbod:
# Why

Currently, the refresh endpoint is being called even when there is no refreshToken stored. (or undefined)

# How

Wrapping the axios call with an if-condition can reduce calls to the refresh endpoint when there is no need. Ultimately, we should decode the JWT and also check if 'exp' has been reached before refreshing in the first place, but that's just a small detail. Since our tokens are short-lived, it's not a big issue.

# Test Plan


intergalacticspacehighway (approved):
LGTM ❇️

alantoa (approved):
LGTM

PR #1965, opened 2023-03-20
Update error popup copy

alantoa:
# Why  

the message 'experiencing high usage' is not accurate when encountering a 500 error during the Claiming and Dropping processes. 
so if we change it to `Please contact us at help@showtime.xyz if this persists. Thanks!` will be more accurate.


hirbod (approved):
LGTM

PR #1966, opened 2023-03-20
add “Frequently Asked Questions” to our footer

alantoa:
# Why

# How

# Test Plan

### DIFF
--- a/packages/app/hooks/use-footer.ts
+++ b/packages/app/hooks/use-footer.ts
@@ -18,6 +18,10 @@ const links = [
     link: "/assets/Showtime-Assets.zip",
     title: "Brand Assets",
   },
+  {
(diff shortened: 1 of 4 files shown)


packages/app/hooks/use-footer.ts
```diff
@@ -18,6 +18,10 @@ const links = [
     link: "/assets/Showtime-Assets.zip",
     title: "Brand Assets",
   },
+  {
+    link: "https://showtime-xyz.notion.site/Frequently-Asked-Questions-Showtime-f57d533e7fb84c7c9728cd0248764c78",
+    title: "Frequently Asked Questions",
+  },
```
hirbod, line 24: Do we want to add alias /faq like we did for /apply or is this useless?

hirbod (approved):

packages/app/hooks/use-footer.ts
alantoa: i'm not sure so let keep it now

PR #1968, opened 2023-03-20
fix: nsfw-gate-blur-web

hirbod:
# Why

Expo Blur had some weird results on Web.

# How

I replaced it against a solution with View and Tailwind, which also removes the need of an absolute positioned View for Android.

# Test Plan

### DIFF
--- a/apps/next/src/pages/nft/[chainName]/[contractAddress]/[tokenId]/index.tsx
+++ b/apps/next/src/pages/nft/[chainName]/[contractAddress]/[tokenId]/index.tsx
@@ -31,7 +31,7 @@ export async function getServerSideProps(context) {
               nft.creator_name ?? getCreatorUsernameFromNFT(nft)
             } | Showtime`,
             description: nft.token_description,
-            image: imageUrl,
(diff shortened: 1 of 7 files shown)


apps/next/src/pages/nft/[chainName]/[contractAddress]/[tokenId]/index.tsx
```diff
@@ -31,7 +31,7 @@ export async function getServerSideProps(context) {
               nft.creator_name ?? getCreatorUsernameFromNFT(nft)
             } | Showtime`,
             description: nft.token_description,
-            image: imageUrl,
+            image: nft?.nsfw ? null : imageUrl,
```
Comment (alantoa) [inline apps/next/src/pages/nft/[chainName]/[contractAddress]/[tokenId]/index.tsx]: should we use our logo image by default?

alantoa (approved):
LGTM 🔥

apps/next/src/pages/nft/[chainName]/[contractAddress]/[tokenId]/index.tsx
Comment (hirbod) [reply apps/next/src/pages/nft/[chainName]/[contractAddress]/[tokenId]/index.tsx]: I would like to. Do we have a CDN image for that?

PR #1969, opened 2023-03-21
fix: don't refresh accessToken while still valid

hirbod:
# Why

Currently, we refresh our access token every time the app is reloaded or booted up from a cold start, as well as when the app is brought back from being in the background for more than 8 hours. This results in excessive calls to the API.

# How

We decode the JWT and extract the exp field to check its validity. We refresh the token only if its lifetime is less than 8 hours, although we could choose an even shorter duration. Regardless, this approach should significantly reduce the number of API calls.

Since we have a JSI btoa/Buffer Polyfil, the operation is really fast.

# Test Plan


intergalacticspacehighway (approved):
LGTM ❇️

alantoa (approved):
LGTM 🔥

PR #1971, opened 2023-03-21
remove bit

intergalacticspacehighway:
# Why
- bit is becoming painful to use with weird errors.
- let's remove the bit/universal setup for now so we can focus on shipping stuff. 
- Let's revisit how we need to handle the versioning of modules. I think a single tree-shakable `showtime-universal` package makes most sense.

# How
Find and replace

# Test Plan
Test app works


alantoa (approved):
LGTM 🔥 🔥 🔥

hirbod (approved):
🙏🏼

PR #1976, opened 2023-03-21
fix: align onboard gating with backend

hirbod:
# Why

Backend adjusted the booleans, so we can now rely on them to detect gating

# How

Changed the logic when to show the captcha based on `has_social_login` which defaults true for:

- Twitter, Instagram, Spotify, Apple and Google
Defaults to false for Email and Phone.

# Test Plan


intergalacticspacehighway (approved):
LGTM ❇️

alantoa (approved):
LGTM 🖖

PR #1977, opened 2023-03-21
fix: dont log while on prod

hirbod:
# Why

Currently, we log on prod, which is bad for performance. If we decide to actually do something with our logs in the future, we can add a strategy.

# How

Only output console.log on __DEV__

# Test Plan


alantoa (approved):
LGTM

PR #1978, opened 2023-03-22
feat: amplitude

intergalacticspacehighway:
# Why?
- We want to directly send the event to amplitude instead of using rudderstack.
- Track all events on frontend
- Moved all the backend events. Here is the [list of events](https://github.com/showtime-xyz/showtime-frontend/pull/1978/files#diff-6c1a1cc5b0a570bb287b28680033a73e6588c2f5a5abffbbdab0d17ed5785f8dR29) we'll be tracking as of now
- Created a new amplitude project. [This project](https://data.amplitude.com/its-showtime/Showtime/events/main/latest) has dev and prod environments. 

# How?
- Remove rudder, use amplitude sdk
- Move backend event tracking to frontend.

# Test Plan
- Test events are getting tracked on amplitude on dev environment.


alantoa (approved):
LGTM 🔥 🔥 🔥

hirbod (approved):
LGTM. Awesome, this should help to get reliable funnel data.

PR #1979, opened 2023-03-22
fix: feed-item on mobile had wrong height

hirbod:
# Why

iOS Smart App banners break viewport height calculations, because they inject late. This caused a long lasting bug on our mobile view (hiding the collect buttons behind the toolbar)

# How
Usually you would mess around with weird JS hacks or try to fix that weirdness with some sort of `-webkit-fill-available`.
Not anymore. Safari finally supports `100svh, 100lvh and 100dvh`. A simple `max-height: 100dvh` was enough to fix the issue for us. Also added min of 100dvh, so it can even paint faster. This is truly amazing to finally have a reliable way of working with full height units on Safari. 

# Test Plan

https://user-images.githubusercontent.com/504909/227058178-b23b9f6a-318a-4e4e-ab1c-99aaafc93d7b.MP4

### DIFF
--- a/packages/app/components/feed-item/feed-item.web.tsx
+++ b/packages/app/components/feed-item/feed-item.web.tsx
@@ -85,7 +85,10 @@ export const FeedItem = memo<FeedItemProps>(function FeedItem({
   return (
     <>
       <LikeContextProvider nft={nft} key={nft.nft_id}>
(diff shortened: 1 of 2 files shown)


intergalacticspacehighway (approved):
LGTM 🔥🔥🔥

alantoa (approved):
LGTM 🔥

PR #1980, opened 2023-03-23
analytics: track play on spotify pressed event

intergalacticspacehighway:
# Why
Added a new event type to track play on spotify


hirbod (approved):
LGTM

PR #1981, opened 2023-03-24
Feat/improve design system import

alantoa:
# Why

- We cannot directly change `@showtime-xyz/universal.` to  `design-system/`  because it will disrupt our design system releases, so I’m trying to use local packages instead of remote design system packages, it will always use the latest design system, so now we don’t have to release every change. 
- We need to upgrade our Next.js version because `@next/font` has been moved to Next.js's 13.2.x version. 
- Since we started importing local files, we have been encountering a lot of hydration issues, and we need to fix them.

# How 

- Updated `tsconfig.json`
- To resolve import issues on native, we need to use Babel's `module-resolver` plugin.
- Many pages do not support hydration so we need to disable SSR. 
 
# Test Plan

Check and run if everything works.


apps/expo/babel.config.js
```diff
@@ -27,282 +51,7 @@ module.exports = function (api) {
...
-            "packages/design-system/utils",
-            "index.tsx"
-          ),
-          "@showtime-xyz/universal.verification-badge": path.join(
-            __dirname,
-            "../../",
-            "packages/design-system/verification-badge",
-            "index.tsx"
-          ),
+          ...resolveDesignSystemPackages(),
```
hirbod, line 54: 💯

intergalacticspacehighway (approved):
LGTM ❇️

intergalacticspacehighway (approved):

PR #1986, opened 2023-03-29
fix: sign message prompt

intergalacticspacehighway:
# Why
Sometimes wallet session stays connected after browser reload so it doesn't show wallet select modal and directly shows the signature modal. I managed to reproduce the issue locally. Looks like a bug in wallet connect.

# How
Disconnect wallet before doing connect call. This makes sure we show the wallet selection modal

# Test Plan
Test sign in with wallet


alantoa (approved):
This is a huge improvement for our testing. I always wanted to give this feedback! 
LGTM!

PR #1987, opened 2023-03-29
fix: Select component invisible

alantoa:
# Why

resolve visibility issue with Select component

# How

# Test Plan


intergalacticspacehighway (approved):
LGTM ❇️

PR #1989, opened 2023-03-29
fix: modal auto-responsive height

alantoa:
# Why

Currently, our modal height is not perfect and sometimes exceeds the window screen. 
Now that CSS supports `svh` and `lvh` units, we can use them to improve it. 

# How

Use `svh` instead of `vh` on mobile browsers. If the current browser does not support svh, it will fallback to using `vh`. 

## Before 
CleanShot 2023-03-29 at 4 07 01

## After 
CleanShot 2023-03-29 at 4 07 51


hirbod (approved):
LGTM. The new units are a game changer.

intergalacticspacehighway (approved):
LGTM ❇️

PR #1992, opened 2023-03-29
fix: don't prefetch links visible in viewport

hirbod:
# Why

The Link component in Next.js prefetches links that are visible in the viewport, leading to inaccurate analytics, unnecessary fetches, and increased AppEngine costs.

Some more background: https://github.com/vercel/next.js/discussions/24009

# How

I have disabled that feature since it only provides a minor speed improvement at the cost of unnecessary prefetching. The Link component still preloads on hover by default, so there is no need to worry about losing that functionality.

# Test Plan

### DIFF
--- a/packages/app/navigation/link/link-core.web.tsx
+++ b/packages/app/navigation/link/link-core.web.tsx
@@ -20,7 +20,7 @@ function LinkCore({
   componentProps?: any;
 }) {
   return (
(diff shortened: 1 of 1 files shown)


intergalacticspacehighway (approved):
LGTM ❇️ wow that is default true 🤯

PR #1994, opened 2023-03-29
fix: mute button broken on mobile web

hirbod:
# Why

Mute button was broken on web

# How

# Test Plan


hirbod:
Merging, since its urgent and a tiny fix

PR #1995, opened 2023-03-30
fix: correctly handle video clicks

hirbod:
# Why

Directly clicking on a video would unmute and pause at the same time, which lead into a weird UX.

# How

The global click interceptor checks for the target now and let handle muted videos directly in the item.
So on web, a click on a muted video will first unmute it and every further click will pause/resume.

# Test Plan


alantoa (approved):
LGTM 💯

PR #1996, opened 2023-03-30
fix: don't show `block user` to self

hirbod:
# Why

Block user (even when gated by backend) was shown to own NFTs.

# How

Added `isSelf` condition around the button.

# Test Plan


alantoa (approved):
LGTM

PR #1997, opened 2023-03-30
feat: improve drop slug

intergalacticspacehighway:
# Why
- Improve drop slug. `/@:username/:dropSlug`

# How
Add a new URL on native and web to handle

# Test Plan
Test slugs work on native and web

P.S. Prev button is broken as found by @hirbod. I have a fix local, testing it. I'll add it in a separate PR.


apps/next/src/pages/profile/[username]/[dropSlug].tsx
```diff
@@ -0,0 +1,50 @@
...
+            } | Showtime`,
+            description: nft.token_description,
+            image: nft?.nsfw ? fallbackImage : imageUrl,
+            deeplinkUrl: `/@${username}/${dropSlug}`,
+          },
+        },
+      };
+    }
+  } catch (e) {
+    console.error(e);
```
Comment (hirbod) [inline apps/next/src/pages/profile/[username]/[dropSlug].tsx]: Nit: use logger

apps/next/src/pages/api/.well-known/apple-app-site-association.ts
```diff
@@ -1,6 +1,7 @@
 import type { NextApiRequest, NextApiResponse } from "next";
 
-const isDevelopment = process.env.STAGE === "development";
+const isDevelopment =
+  process.env.NEXT_PUBLIC_WEBSITE_DOMAIN === "dev.showtime.xyz";
```
hirbod, line 4: sure it should not also contain ` || process.env.STAGE === "development"` as OR operator?
Beside that questing, LGTM!

hirbod (approved):

alantoa (approved):
LGTM 🔥 🔥 🔥

apps/next/src/pages/api/.well-known/apple-app-site-association.ts
intergalacticspacehighway: Actually, `STAGE` doesn't seem to be visible. It's value was undefined. So i thought of using the NEXT_PUBLIC* variable 😅. will dig and improve it. STAGE should be visible since it's running the code server side. But i need to check what's wrong

PR #2000, opened 2023-03-31
fix: forward button browser

intergalacticspacehighway:
# Why
- This is a workaround solution to fix the forward button. Reported by @hirbod https://showtime-rq88331.slack.com/archives/C02QD3J7DJM/p1680177169673189?thread_ts=1680117135.634529&cid=C02QD3J7DJM
- Long term fix would be use a context to store the ephemeral state instead of this URL hack. Keeping it low priority if this solution is good enough 😅 

# How
- Use pushState and listen to window popstate to inform next router to go back. It is very hacky. 

# Test Plan
Test forward and back button from feed to detail screen

### DIFF
--- a/packages/app/components/swipe-list.web.tsx
+++ b/packages/app/components/swipe-list.web.tsx
@@ -72,6 +72,10 @@ export const SwipeList = ({
   );
 
   useEffect(() => {
+    const popStateListener = () => {
+      window.removeEventListener("popstate", popStateListener);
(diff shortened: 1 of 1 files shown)


alantoa (approved):
LGTM 💯

hirbod (approved):
Since it was broken for a while, I think this workaround is good enough for now.
LGTM

PR #2001, opened 2023-03-31
fix: vscode auto import

alantoa:
# Why

When using automatic import in vscode, there is an issue where it adds the `packages/*` prefix when importing any component from the `packages/app/*` files.  

For example: auto import is `"packages/app/context/like-context";` ranther than `"app/context/like-context";`

# How 

Update ts `tsconfig.json` file. 

# Test Plan

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -34,7 +34,6 @@ module.exports = {
     "react/jsx-uses-react": "off",
     "react/react-in-jsx-scope": "off",
     "react/prop-types": "off",
-    "unused-imports/no-unused-imports": "warn",
     "unused-imports/no-unused-vars": "warn",
     "no-empty": "warn",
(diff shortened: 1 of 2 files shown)


hirbod (approved):

hirbod:
Can we please also remove the rule which removes unused imports on save? It can warn, but it should not automatically remove. It is so annoying when copying stuff from one file to the other and it automatically removes stuff

intergalacticspacehighway (approved):
LGTM ❇️

hirbod:
I pushed a change @alantoa, you can review. This will warn if we have unused imports but will not automatically remove them.

alantoa:
@hirbod yeah, agreed with you!

hirbod:
Ok, I can't describe how often it annoyed me :D

PR #2005, opened 2023-04-02
fix: XCode 14.3 temp issues with config plugin

hirbod:
# Why

Our App can't compile with the new XCode 14.3 currently. 

# How

Looks like there are packages which haven't updated their podspecs to target the correct deployment target. Added a Config plugin to fix this issue for now, since build properties alone are not enough to fix this.

Credit goes to: https://github.com/facebook/react-native/issues/34106#issuecomment-1493040686

The upstream bugfix is already merged but I don't know if this is going to be backported, so we need to monitor new releases.
https://github.com/facebook/react-native/pull/36759

# Test Plan

### DIFF
--- a/apps/expo/app.config.js
+++ b/apps/expo/app.config.js
@@ -160,6 +160,7 @@ export default {
     "./plugins/react-native-cronet.js",
     "./plugins/with-spotify-sdk.js",
     "./plugins/with-android-splash-screen.js",
+    "./plugins/fix-deployment-target.js",
     [
       withInfoPlist,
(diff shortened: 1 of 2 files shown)


Kudo:
hi there! hope react native will release a new patch version with the fix this week and we will update with a new `expo` package then. let's see if react native could release the fix this week.

intergalacticspacehighway (approved):
LGTM ❇️

alantoa (approved):
LGTM 🙇‍♂️

PR #2008, opened 2023-04-03
fix: allow display name change for verified when empty

hirbod:
# Why

Verified users can't change name once they are verified. We have edge cases, where we have verified users but they still haven't entered a display name.

https://showtime-rq88331.slack.com/archives/C02PXGK3V8D/p1680502899279759

# How
Allow change of the name (not username) if the field is empty. @jorgeavaldez can you take care of the backend part please? Thanks King!

# Test Plan

### DIFF
--- a/packages/app/components/edit-profile.tsx
+++ b/packages/app/components/edit-profile.tsx
@@ -115,7 +115,6 @@ export const EditProfile = () => {
   const { width } = useWindowDimensions();
   const isMdWidth = width >= breakpoints["md"];
   const { isValid, validate } = useValidateUsername();
-  const socialLinks = useLinkOptions();
   const isDark = useIsDarkMode();
   const pickFile = useFilePicker();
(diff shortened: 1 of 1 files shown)


packages/app/components/edit-profile.tsx
```diff
@@ -135,24 +134,10 @@ export const EditProfile = () => {
...
-        const foundLink = user.data.profile.links.find(
-          (l) => l.type_id === s.id
-        );
-
-        if (foundLink) {
-          links[s.id] = foundLink.user_input;
-        }
-      });
-    }
-
```
hirbod, line 150: I removed these since we're not using this fields anymore

intergalacticspacehighway (approved):
LGTM ❇️ (post BE changes)

alantoa (approved):
LGTM

hirbod:
Backend was merged, we can merge as well now.

PR #2009, opened 2023-04-03
analytics: before login save music press event

intergalacticspacehighway:
@BossLevan needs this event 🫡


intergalacticspacehighway:
we could also track the success event. @BossLevan let me know!

hirbod (approved):
LGTM ✦

PR #2011, opened 2023-04-03
feat: improve video backdrop

alantoa:
# Why

context: https://showtime-rq88331.slack.com/archives/C02Q0E4PBD0/p1680522765432229

# How

# Test Plan


packages/app/components/feed-item/feed-item.md.tsx
```diff
@@ -138,10 +138,16 @@ export const FeedItemMD = memo<FeedItemProps>(function FeedItemMD({
   };
 
   const mediaHeight = Math.min(windowWidth, feedItemStyle.height) - 160 - 80;
```
hirbod: Could you add some comments for the magic numbers?

hirbod (approved):
LGTM. I'd extract all magic numbers to CONST or add comments so its clear what 160 or 80 is. Thank you!

PR #2013, opened 2023-04-03
fix: detail screen url handling

intergalacticspacehighway:
# Why

- Fix was to use the query object which also changes the query parameter in the URL. 
https://twitter.com/leeerob/status/1642927637860433924
- I also did some cleanup. Removed the `/list` route. We can manage it in the drop detail screen.

# How
Use query object in next router to change the query parameters. 

# Test Plan
- Test browser back/forward button works when switching to drop detail screen from card items


packages/app/lib/constants.ts
```diff
@@ -182,4 +182,5 @@ export const CURRENCY_NAMES = {
 export const HIDE_LINK_FOOTER_ROUTER_LIST = [
   "/nft/[chainName]/[contractAddress]/[tokenId]",
   "/list",
+  "/profile/[username]/[dropSlug]",
```
alantoa, line 185: we missed here, I just added it!

alantoa (approved):
LGTM ✨

PR #2014, opened 2023-04-03
fix: navigateToOnboarding replace error

alantoa:
# Why

`navigateToOnboarding` params is not correct, I just fix it. 

# How

# Test Plan


hirbod:
I said untested :D

PR #2017, opened 2023-04-04
fix: delay avatar hover card hooks rendering

hirbod:
# Why

While working on Trending 2.0 and the "Claimed by" component, I realized that our AvatarHoverCard for the web was spamming our servers. Although Radix-UI is clever enough to render the Avatar Card lazily, we made the mistake of putting all the data fetching hooks in the root component. This resulted in every small avatar triggering 2 API fetches, and at times, we had more than 20 visible Avatar Cards

# How

I moved the content of the Avatar Hover Card into a separate component, and now both the fetches and the content are rendered lazily, resulting in a nearly 30% reduction in XHR requests on the web. Also optimized the closing behavior a bit.

# Test Plan


hirbod:
@alantoa @intergalacticspacehighway merging, since its a tiny change with big impact.

intergalacticspacehighway:
Good catch sir 🫡

PR #2018, opened 2023-04-05
fix: swipe list from feed on web

intergalacticspacehighway:
# Why
Swipe list on web got broken in this PR's some last few cleanup commits - https://github.com/showtime-xyz/showtime-frontend/pull/2013

# How
Pass queryParams to the swipe list component for other feed items. Still not the cleanest approach, but faster. Ideally, imo we should stop using URL params in this case. If it breaks now, I'll do a rewrite!

# Test Plan
Test swipe list on web from lists


alantoa (approved):
LGTM 🙏

PR #2021, opened 2023-04-05
fix: modal not popping when route pattern matches url

intergalacticspacehighway:
# Why
I don't think we're supporting the below usecase in any of the URLs. So I thought of removing the special handling.

# Test Plan
Test all modals with close button on web


hirbod (approved):
Tested all the modals on the Vercel preview, couldn't find anything breaking beside the Trending thing, but this was fixed in a separate PR already.

LGTM

alantoa (approved):
yeah! LGTM 💯

PR #2024, opened 2023-04-05
fix: dont output "creator" by default

hirbod:
# Why

We're outputting "creator" everywhere, which is really redundant. 
Talked with Alex, we agreed to remove it globally. If its needed somewhere specific, the label can be passed.

# How

Removed default value.

# Test Plan

Bildschirm_foto 2023-04-05 um 23 25 34


hirbod:
@alantoa @intergalacticspacehighway merging since little change

PR #2027, opened 2023-04-06
feat: Trending 2.0

hirbod:
# Why

This PR brings the all new Trending view. 
And its pretty fast :)

# How

# Test Plan


packages/app/components/card/list-card.tsx
```diff
@@ -0,0 +1,356 @@
...
+
+function ListCard(props: Props) {
+  const { width } = useWindowDimensions();
+  const isLgWidth = width >= breakpoints["md"];
+
+  if (!isLgWidth) {
+    return <ListCardSmallScreen {...props} />;
+  }
+
+  return <ListCardLargeScreen {...props} />;
```
intergalacticspacehighway, line 99: not a blocker: i have seen stuff jumping as we use JS based logic for screen resolution mounting. We should probably start using tailwind's `hidden` in future.

packages/app/components/card/list-card.tsx
```diff
@@ -0,0 +1,356 @@
...
+      nft?.token_description
+        ? linkifyDescription(removeTags(nft?.token_description))
+        : "",
+    [nft?.token_description]
+  );
+
+  return (
+    <View
+      // @ts-expect-error TODO: add accessibility types for RNW
+      accessibilityRole="article"
```
intergalacticspacehighway, line 267: 💯

packages/app/components/card/list-card.tsx
hirbod: This was the logic of the old `Card`, I just optimized it a bit. We can keep optimizing in the future for sure.

intergalacticspacehighway (approved):
LGTM 🔥🔥🔥

packages/app/components/trending/trending.web.tsx
```diff
@@ -65,56 +71,38 @@ export const Trending = () => {
...
-            new Array(numColumns - chuckItem.length)
-              .fill(0)
-              .map((_, itemIndex) => (
-                <View key={itemIndex.toString()} tw="flex-1" />
-              ))}
+        <View tw="mx-auto max-w-screen-xl flex-row md:px-4">
+          <ListCard
+            nft={item}
+            showClaimButton
+            href={`/${getNFTSlug(
```
intergalacticspacehighway: `getNFTSlug` adds the `slash` so we can remove it from here.

hirbod: fixed

PR #2036, opened 2023-04-07
fix: perf of NFTDropdown Menu

hirbod:
# Why

# How

# Test Plan


hirbod:
Turns out it does not work on iOS, even though giving good perf on web :(

PR #2050, opened 2023-04-09
fix: SDK 47 minor upgrades

hirbod:
# Why

Expo released a new version of RN for SDK 47 which fixes the build issues on XCode 14.3

# How

Upgraded packages. Also added a config plugin which fixes forced dark mode on Android (causing weird colors based on heuristics). This upgrade will force us to release a new major release to the stores.

# Test Plan


alantoa (approved):
LGTM! 
also, It means we can't use hot update when we merge this PR because it has some native lib updates.

hirbod:
Exactly, that's why I haven't merged it yet. But if we upgrade to SDK 48 anyway, me maybe can skip this one here anyway. But this PR also added a fix for dark mode on MIUI platforms and some other Android phones.

hirbod:
I'll close this PR, since SDK 48 is almost ready.

PR #2057, opened 2023-04-11
fix: expo app building

alantoa:
# Why

i’m not sure why storybook-react-native project affected expo app now, so I just removed `react-native-gesture-handler ` from Storybook  project to ensure expo project can working will first. will dig it later. 
# How

# Test Plan


alantoa:
it's exists in https://github.com/showtime-xyz/showtime-frontend/pull/2050

PR #2058, opened 2023-04-11
feat: apple music save drop

intergalacticspacehighway:
# Why
We want to enable creators create apple music gated drops. This PR adds apple music gated Save drop. Note: Only Save drops are supported for now, not pre-save drops.

# How
### Creator side
- When creator has `apple_music_artist_id` set in backend, they would see apple music track id field in create music drop form. 
- The apple music track id field will only be visible for a Save drop. 

### Collector side
- Users would be able to either connect Spotify or Apple music during collecting the drop.
- Users will see 2 buttons on drop card - Apple music and Spotify. They can choose either and collect the drop.
- Users would be able to disconnect/connect apple music from account settings.

# Test Plan
- Set apple_music_artist_id from the admin panel and try creating a Save music drop (keep `is Live` checked).
- Try claiming the music drop authenticated and unauthenticated

Testcases i tried are below.
```
iOS, Android Web

### Authenticated/Unauthenticated flow
- Apple music + Spotify 
- Spotify
- Apple music 
```

### Known issues
- There is a known issue when disconnecting apple music token and immediately trying to connect to new account fails. Max will be looking into it. I have reported him the issue.
- There are some button design issues. I'll post it on Slack. Right now deploying it on dev.showtime for testing


packages/app/components/drop/drop-music.tsx
```diff
@@ -174,9 +174,9 @@ export const DropMusic = () => {
         if (isSaveDrop) {
           return baseSchema.concat(
             yup.object({
-              spotifyUrl: yup
-                .string()
-                .required("Spotify Song URL is required for saved drops."),
+              spotifyUrl: yup.string(),
+              appleMusicTrackId: yup.string(),
```
intergalacticspacehighway: // TODO - will be fixing this before merging. One of the two needs to be mandatory.

packages/app/hooks/use-connect-apple-music/utils.ts
```diff
@@ -0,0 +1,6 @@
+export const developerToken =
```
intergalacticspacehighway: oh this needs to be moved into Next's API or our backend. I think backend is better. Will ping Max.

packages/app/components/claim/claim-button.tsx
```diff
@@ -221,12 +252,79 @@ export const ClaimButton = ({
...
+      <View tw="w-full flex-row">
+        {edition?.apple_music_track_url ? (
+          <Button
+            {...buttonProps}
+            onPress={() => handleCollectPress("appleMusic")}
+            tw="flex-1 flex-row items-center justify-center"
+            disabled={isAppleMusicCollectLoading}
+          >
+            <AppleMusic
+              width={22}
```
alantoa: I didn't check this icon on the website, but I feel that its size may be incorrect because it appears to lack spacing. I'll check it!
- Spotify 
CleanShot 2023-04-25 at 12 45 32

- Apple 
CleanShot 2023-04-25 at 12 44 17

alantoa (approved):
LGTM 🔥 🔥🔥

PR #2063, opened 2023-04-12
fix: description breaking app

hirbod:
# Why

We have old cases where there is no description text. This was breaking our code.

# How

Added a null check

# Test Plan


alantoa (approved):
LGTM 🙏

PR #2064, opened 2023-04-12
feat: virtualized list

intergalacticspacehighway:
# Why
We need to preserve scroll position. It was getting tricky with virtuoso, still tricky with tanstack virtual. But this seems to be working. Also, it is much faster.

# How
Replace Virtuoso with tanstack virtual. Should only affect desktop web and mobile web

# Test Plan
Test all the lists in desktop and mobile web


packages/app/components/notifications/index.tsx
```diff
@@ -121,10 +121,6 @@ export const Notifications = ({
             ios: windowHeight,
           }),
         }}
-        overscan={{
```
intergalacticspacehighway, line 124: hardcoding overscan to 2 in the list component itself. It will render 2 additional items above and below

packages/app/components/profile/profile.web.tsx
```diff
@@ -298,14 +298,10 @@ const Profile = ({ username }: ProfileScreenProps) => {
                 useWindowScroll={isMdWidth}
                 ListHeaderComponent={Header}
                 numColumns={1}
+                preserveScrollPosition
```
intergalacticspacehighway, line 301: didn't intend to add an extra prop. But it was getting tricky with many lists in single page. I was assuming that we'd require scroll position saving just for window scroll lists, but we're not using window scroll on mobile web, so we need to save scroll position for window and normal virtualiser  😕

hirbod:
I'd increase the overscan for comments even more, since small lists act weird with the scroll bar and momentum when you try to scroll up. I think we can handle some more comments easily.

packages/app/components/profile/profile.web.tsx
hirbod: Why can't we actually use window scroll on mobile? With the new dvh it should be pretty doable, or?

packages/app/components/notifications/index.tsx
hirbod: I think it should still be overwritable on a component basis, because specially on short lists it acts a bit weird when it messes with the elements. It breaks momentum scrolling and stops scrolling.

hirbod (approved):
Incredible work @intergalacticspacehighway - works pretty good now after so many iterations and it added a really good performance boost and released stress from the main thread thanks to translateY instead of height manipulation.

I think some components still have some room for a tiny manual intervention regarding overscan but we can address this separately.

apps/next/package.json
```diff
@@ -35,6 +35,7 @@
     "@sentry/nextjs": "7.7.0",
     "@stripe/react-stripe-js": "2.1.0",
     "@stripe/stripe-js": "^1.52.1",
+    "@tanstack/react-virtual": "^3.0.0-alpha.0",
```
alantoa: should we replace it with `"^3.0.0-beta.54"`? because we are actually using `beta.54` in our project and this will provide clarity.

yarn.lock
```diff
@@ -11317,6 +11317,24 @@ __metadata:
   languageName: node
   linkType: hard
 
+"@tanstack/react-virtual@npm:^3.0.0-alpha.0":
+  version: 3.0.0-beta.54
+  resolution: "@tanstack/react-virtual@npm:3.0.0-beta.54"
```
alantoa, line 11322: here is real lib version

apps/next/package.json
hirbod: Oh good catch @alantoa - we're on the earliest alpha, which is crazy

alantoa (approved):
LGTM 🐐 🐐 🐐

apps/next/package.json
intergalacticspacehighway: Done! good catch.

intergalacticspacehighway:
@hirbod added overscan support back and increased it to 6 for comments.

packages/app/components/profile/profile.web.tsx
intergalacticspacehighway: i think yes, it should be doable. I am not sure what was the exact reason of removing it. @alantoa would know better. if we do that, we can remove this prop for sure and just support scroll position for window lists.

packages/app/components/notifications/index.tsx
intergalacticspacehighway: done

packages/app/components/profile/profile.web.tsx
intergalacticspacehighway: it can be done in a separate PR though!

PR #2070, opened 2023-04-13
fix: toast duration inconsistency between web and native

intergalacticspacehighway:
# Why
Made toast duration to accept milliseconds on native same as on web


alantoa (approved):
LGTM 💯 
It's a little bit annoying, by the way. Thanks for fixing it 🙏

PR #2073, opened 2023-04-13
lazy load stripe

intergalacticspacehighway:
# Why
Lazy load stripe at the checkout form

# Test Plan
Tested checkout flow works on local. Stripe only loads when the form shows up.


alantoa (approved):
LGTM 💯

hirbod (approved):
Perfect. +6 on LightHouse. I have more ideas to get the number higher

PR #2074, opened 2023-04-13
feat: optimize notification usability

hirbod:
# Why

The Notification UX was not very good. You could only tap/click the bold part to navigate to.

# How
I haven't removed the single actions but rather added a Pressable around in case the whole notification was touched.
Also fixed an issue with nested anchor tags.

# Test Plan


alantoa (approved):
LGTM ✨

PR #2076, opened 2023-04-14
fix: grid item performance and image crashes

hirbod:
# Why

PressableScale, Expo-AV for still previews, and PinchToZoom are pretty expensive for ListItems, while not adding any benefits.

# How

Changed "PressableScale" to "Pressable", used still images for Grid-Items instead of paused videos, and added a new "GridMedia" component, which also does not include PinchToZoom. Additionally, a function was added to clear the image memory on every navigation state change.

# Test Plan

### DIFF
--- a/packages/app/navigation/index.tsx
+++ b/packages/app/navigation/index.tsx
@@ -1,6 +1,8 @@
 import { useState, useRef, useMemo } from "react";
 import { Platform } from "react-native";
 
+import { Image } from "expo-image";
+
 import { useIsDarkMode } from "@showtime-xyz/universal.hooks";
 
 import {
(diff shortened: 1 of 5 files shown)


intergalacticspacehighway (approved):
LGTM ❇️

packages/app/navigation/index.tsx
```diff
@@ -62,6 +64,13 @@ export function NavigationProvider({
         formatter: (options) =>
           options?.title ? `${options.title} | Showtime` : "Showtime",
       }}
+      onStateChange={() => {
+        // this is a temporary workaround to reduce crashes.
+        // Once we have optimized images and pull from disk, we can remove this
+        // Currently we're collecting images into memory until the app crashes,
+        // therefore we want to clear it on every screen change
+        Image.clearMemoryCache();
+      }}
```
hirbod, line 73: First, I've registered a `setInterval` but I think clearing the memory on navigation state change is the better solution. This will clear the memory when navigating away from the screen and reduces crashes.

alantoa (approved):
LGTM 💯

PR #2077, opened 2023-04-14
feat: upgrade to reanimated v3

alantoa:
# Why

We want to use Reniamteed v3 because It's beneficial for us!

# How

- Upgrade Reanimated version to `3.0.2`.
- Check every component and test them.
- Fixed some broken components.
- Unify the Babel version and clean up some code.

# Test Plan

Run it on your device and check if everything is working well!


intergalacticspacehighway (approved):
LGTM 🔥

hirbod (changes requested):
Working great so far. I think I have tested all components and you did an awesome job. The final thing that happens is the crash I've reported to you on Profile (both iOS and Android). If this is fixed, this is good to go!

PR #2079, opened 2023-04-14
fix: white flash on feed

hirbod:
# Why

Fixes a report by @alantoa that the Home Feed has a white flash on load and I found it also happen on fast swipe with Android.

# How

Added black background color to contentStyle of HomeNavigator

# Test Plan


alantoa (approved):
LGTM 🙏

PR #2083, opened 2023-04-17
Fix/lcp

hirbod:
# Why
Still had a tiny CLS and LCP issue.

# How
This PR fixes CLS on Home (suggested user box was shifting) and eagerly loads the first Image on trending and home.

# Test Plan


alantoa (approved):
LGTM 💯

PR #2094, opened 2023-04-19
fix: send logs to sentry

intergalacticspacehighway:
# Why
Added some logging before and after update happens to debug expo-update related crash.


hirbod (approved):
Just afraid that this will flood our limits pretty quick, since Backend is already doing this a lot. But we can try it.

alantoa (approved):
LGTM

PR #2096, opened 2023-04-19
wip: reload crash

intergalacticspacehighway:
# Why
Seems combination of gesture-handler and reanimated is causing the reload crash. Maybe gesture handler's callbacks capture stale shared values 🤔 ?

# How

# Test Plan


hirbod (commented):
This PR would remove our PinchToZoom Component?

intergalacticspacehighway:
> This PR would remove our PinchToZoom Component?

Not final PR, this is just for testing.

PR #2097, opened 2023-04-19
fix: card to detail swipe list url hack

intergalacticspacehighway:
# Why
Use `as` prop on web to make the URL look as expected


alantoa (approved):
LGTM 💯

PR #2101, opened 2023-04-20
feat: update gating type

alantoa:
# Why

Backend has updated the `gating_type` field and created a new type called `spotify_presave ` to replace the `music_presave` field. 
This is in anticipation of Apple Music Presaves, so that we can have clear gating types for each platform.

# How

Just replaced instances of the `spotify_presave` type with `music_presave`.

# Test Plan

This PR needs to wait for the backend team to merge their PR into staging before we can test it.

### DIFF
--- a/packages/app/components/claim/claim-button.tsx
+++ b/packages/app/components/claim/claim-button.tsx
@@ -83,7 +83,7 @@ export const ClaimButton = ({
   const raffleConcludedAt = useMemo(() => {
     if (!isSelf || !isRaffleDrop) return null;
     if (
-      edition.gating_type === "music_presave" &&
+      edition.gating_type === "spotify_presave" &&
       edition?.presave_release_date
     ) {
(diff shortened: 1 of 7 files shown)


packages/app/components/claim/claim-button.tsx
```diff
@@ -83,7 +83,7 @@ export const ClaimButton = ({
   const raffleConcludedAt = useMemo(() => {
     if (!isSelf || !isRaffleDrop) return null;
     if (
-      edition.gating_type === "music_presave" &&
+      edition.gating_type === "spotify_presave" &&
```
intergalacticspacehighway: if existing drops are `music_presave` it'll hide the button right? Do you think we need `edition.gating_type === "music_presave" || edition.gating_type === "spotify_presave"`?

alantoa: that's makes sense! we need to compatible with it!

PR #2102, opened 2023-04-20
fix: user list momentum scroll

intergalacticspacehighway:
# Why
Makes momentum scroll on userlist smoother. Increased overscan on comments and user list and some tiny optimisation in infinite scroll list.


hirbod (approved):
LGTM - much better

PR #2103, opened 2023-04-20
fix: plenty of perf issues

hirbod:
# Why

Trending List with Video previews is currently to slow and consume too much traffic. Furthermore, backdrop-blur-md was expensive and had weird effects on Desktop in the full detail preview

# How

For now, removed video previews in the Trending List (will add back with Bunny but optimized) and also introduced a new prop called `withBackDropBlur` to control the blurred background image more granular. Furthermore, on web, instead of rendering a second video placeholder, we use the native solution inside videos (usePoster prop was missing while we did provide a posterSource)

# Test Plan


intergalacticspacehighway (approved):
LGTM ❇️

PR #2104, opened 2023-04-20
fix: reserve button space better

hirbod:
# Why

I figured that buttons would still jump on mounting/remounting nodes.

# How

Unmount the children instead of the whole parent.

# Test Plan


intergalacticspacehighway (approved):
LGTM ❇️

PR #2110, opened 2023-04-21
fix: comments lag on momentum scroll

intergalacticspacehighway:
# Why
ClientSideOnly triggered async mounting was causing the issue. Fixed it by triggering setState before render.

### DIFF
--- a/packages/design-system/client-side-only/client-side-only.web.tsx
+++ b/packages/design-system/client-side-only/client-side-only.web.tsx
@@ -1,7 +1,5 @@
 import { useState } from "react";
 
-import { useIsomorphicLayoutEffect } from "app/hooks/use-isomorphic-layout-effect";
-
 export const ClientSideOnly = ({ children }: any) => {
   const [clientSide, setClientSide] = useState(false);
   if (typeof window !== "undefined" && clientSide === false) {
--- a/packages/design-system/client-side-only/client-side-only.web.tsx
+++ b/packages/design-system/client-side-only/client-side-only.web.tsx
@@ -1,10 +1,12 @@
-import { useEffect, useState } from "react";
+import { useState } from "react";
+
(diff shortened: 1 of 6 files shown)


packages/design-system/client-side-only/client-side-only.web.tsx
```diff
@@ -1,10 +1,12 @@
...
+import { useState } from "react";
+
+import { useIsomorphicLayoutEffect } from "app/hooks/use-isomorphic-layout-effect";
 
 export const ClientSideOnly = ({ children }: any) => {
-  const [mounted, setMounted] = useState(false);
-  useEffect(() => {
-    setMounted(true);
-  }, []);
+  const [clientSide, setClientSide] = useState(false);
```
intergalacticspacehighway, line 4: main fix is here, rest all is cleanup 😅

packages/design-system/hooks/index.ts
```diff
@@ -1,14 +1,13 @@
...
 import { LayoutChangeEvent, Platform } from "react-native";
 
 import { useSharedValue } from "react-native-reanimated";
 
 import { useColorScheme } from "@showtime-xyz/universal.color-scheme";
 import { getScrollParent } from "@showtime-xyz/universal.utils";
 
-const useIsomorphicLayoutEffect =
-  typeof window !== "undefined" ? useLayoutEffect : useEffect;
+import { useIsomorphicLayoutEffect } from "app/hooks/use-isomorphic-layout-effect";
```
alantoa: I think our design system folder components shouldn't rely on any `app` folder files. Otherwise, if we want to release them someday, it will be troublesome. wdty?

intergalacticspacehighway: oh yes, totally missed it's importing in design folder

hirbod (approved):
Woah! Awesome find.

PR #2111, opened 2023-04-21
fix: sidebar jumping on loading

hirbod:
# Why
Currently, the sidebar (web desktop version) for an NFT detail jumps around while loading.

# How

Reserve space for LikedBy, Claimers, and the Button, so they load without shifting too many elements around. Also, reduce the width of "LikedBy" to show only one name instead of two, as it frequently runs into two lines

# Test Plan


intergalacticspacehighway (approved):
LGTM ❇️

alantoa (approved):
LGTM

packages/app/components/camera/camera-buttons.tsx
```diff
@@ -34,7 +34,7 @@ export function CameraButtons({
   const isDark = useIsDarkMode();
 
   return (
-    <View tw="flex-row items-center justify-between py-8 px-6">
+    <View tw="flex-row items-center justify-between px-6 py-8">
```
alantoa: btw I removed the files in this folder, so there is a conflict. I will solve it!

PR #2113, opened 2023-04-21
fix: profile list height incorrect

alantoa:
# Why
our profile list is not correct on web. 
# How

# Test Plan

# Before 
CleanShot 2023-04-21 at 11 15 41@2x 

# After 

CleanShot 2023-04-21 at 11 15 11@2x


intergalacticspacehighway (approved):
LGTM ❇️

PR #2118, opened 2023-04-25
Fix/UI bugs

alantoa:
# Why
This PR is just for some UI fixes and code improvements. 
- [Fixed spacing issue for drop-down menu, view share screen on web.](https://github.com/showtime-xyz/showtime-frontend/commit/e845d8a03a4938b2843ffa7c8b71190bba68bca3)
- [Fixed double tap to open heart icon position is not correct.](https://github.com/showtime-xyz/showtime-frontend/commit/d423e16b0eaac5c50f1740cb57ba6e9b52d98371)
- [Improved useBottomTabHeight hooks](https://github.com/showtime-xyz/showtime-frontend/commit/76e920bfbc9d3eeb47ae2fedc1880b768f593587)
# How

# Test Plan


packages/app/lib/react-navigation/bottom-tabs/index.web.ts
```diff
@@ -1,10 +1,5 @@
 import { createContext } from "react";
 
-const useBottomTabBarHeight = () => {};
```
alantoa, line 3: I removed this because this is incorrect for the web, we should use the `usePlatformBottomHeight` hook instead.

packages/app/components/feed/feed-item-tap-gesture.tsx
```diff
@@ -161,7 +165,12 @@ export const FeedItemTapGesture = ({
         style={[heartContainerStyle, heartStyle, topOffset]}
         pointerEvents="none"
       >
-        <HeartFilled width={90} height={90} color={colors.rose[500]} />
+        <HeartFilled
+          style={shadowStyle}
```
alantoa, line 169: This change will fix an anoying warning log called `(ADVICE) View #563 of type RCTView has a shadow set but cannot calculate shadow efficiently...` cc @hirbod

intergalacticspacehighway (approved):
LGTM ❇️

PR #2121, opened 2023-04-26
fix: temporary CDN/NFT image selection

hirbod:
# Why

This PR supports media directly hosted on bunny and with some optimization fallbacks for cloudinary.

# How

# Test Plan


hirbod:
@alantoa @intergalacticspacehighway once tested intensively against prod data it should be good to go.
I'll do a follow up with more fine-tuned cloudinary video-stills, once this is done

alantoa (approved):
LGTM ⚡

PR #2122, opened 2023-04-26
fix: drop form shake

alantoa:
# Why 

- It will be shaken when toggling the `Accordion` component in the drop form. This is because we have used `flex-1` for some special fields. Removing it will make it work properly. 
- We already have bottom padding in the modal header, so it's unnecessary to add extra top padding in `ScrollView`.

# How 

- I removed `tw="flex-1"` on the title field element.
- I updated the `AnimateHeight` component without using Moti because it was previously using Reanimated v1 code in this component.
- Updated ScrollView's padding in drop form.

## Before  

https://user-images.githubusercontent.com/37520667/234527685-486f26a9-ad61-450b-8231-1c035a36614b.mp4

## After 

https://user-images.githubusercontent.com/37520667/234527568-490a99a2-9b47-4362-96ae-9e8c310ff16b.mp4


intergalacticspacehighway (approved):
LGTM ❇️

hirbod:
Can you check if that fixed also Android? I disabled AnimateHeight for Android in the past with a Platform Check

alantoa:
I just tried it and it is working well on Android and Web, so [I made the FieldSet animated height component support them](https://github.com/showtime-xyz/showtime-frontend/pull/2122/commits/a03b48454bc7c96dd34eecfe63a13994912dccb5). @hirbod 

https://user-images.githubusercontent.com/37520667/234614350-e73fdacf-55be-43ae-9a35-b6fc3baaae3e.mp4

hirbod:
In the past, it happened on long error messages. Can you enter some invalid characters into the username for example and test that case as well on Android?

hirbod:
Looks like the error message on web is cutted (maybe only in the login form though)
Beside that, LGTM. Very smooth!

Bildschirmfoto 2023-04-26 um 16 58 40

hirbod (approved):

alantoa:
Yeah, I have tried them. Both of them are working! @hirbod 
Andorid:
CleanShot 2023-04-26 at 11 07 28 
Web: 
CleanShot 2023-04-26 at 11 07 43

PR #2125, opened 2023-04-26
Fix/eslint report

hirbod:
# Why

Fixed all Linting issues.

# How

Most of the changes are real fixes, I did silence a few which would cause too much refactor, like making sure a Promise Function Executor is not async. It's bad practice but if we have time, we can search for disabled rules an try to tackle them

# Test Plan


alantoa (approved):
LGTM 💯

PR #2124, opened 2023-04-26
Upgrade RN & fix storybook

alantoa:
# Why

- Update react-native to `0.71.7`
- Fix storybook page layout
- Improve the import of design system components.

# How

# Test Plan

PR #2126, opened 2023-04-26
fix: maestro build

intergalacticspacehighway:
wip


intergalacticspacehighway:
Fixed by @alantoa here https://github.com/showtime-xyz/showtime-frontend/pull/2124

PR #2132, opened 2023-04-27
fix: improve apple music auth on native

intergalacticspacehighway:
# Why
Adds the webview page in a separate HTML in the public folder. This makes sure we don't load any libraries on this page so it loads faster on native. On iOS we can add a native module, I created [here](https://github.com/showtime-xyz/apple-music-auth) but there's no android equivalent so i thought we can keep the webview if it's fast enough. Will do some more testing.

# How
Create a new html page in public folder and perform apple music auth there for native.

# Test Plan
- Test apple music auth on native. It should be fast enough.
- We still keep the apple-music-auth page in next js for backward compat. We can remove it later in future.

Before/After: Almost 2x difference in load speed.

https://user-images.githubusercontent.com/23293248/234754965-4d1d1be1-b88e-4f0a-8542-c038a1ef91be.mov

https://user-images.githubusercontent.com/23293248/234754983-0e935ea2-2e03-40d2-aea4-a2fa596250ee.mov


intergalacticspacehighway:
Merging to test on prod staging 🙏

PR #2133, opened 2023-04-27
fix: buttons on list card

intergalacticspacehighway:
# Why
Revert to previous designs

### DIFF
--- a/packages/app/components/claim/claim-button.tsx
+++ b/packages/app/components/claim/claim-button.tsx
@@ -263,7 +263,7 @@ export const ClaimButton = ({
           <Button
             {...buttonProps}
             onPress={() => handleCollectPress("appleMusic")}
-            tw="grow flex-row items-center justify-center bg-black dark:bg-white"
+            tw="flex-row items-center justify-center bg-black dark:bg-white"
             disabled={isAppleMusicCollectLoading}
           >
             <View tw="mt-[2px]">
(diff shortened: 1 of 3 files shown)


packages/app/components/claim/claim-button.tsx
```diff
@@ -320,7 +319,7 @@ export const ClaimButton = ({
             tw="ml-1 text-sm font-semibold leading-5"
             style={{ color: isDark ? colors.black : colors.white }}
           >
-            {isAuthenticated ? "Pre-Save to Collect" : "Pre-Save on Spotify"}
+            Pre-Save on Spotify
```
hirbod: What about this line?

hirbod (commented):
Did you test this also on small devices? iPhone SE SIM maybe?

packages/app/components/claim/claim-button.tsx
intergalacticspacehighway: yeah thought to make the copy a bit consistent. Pre-Save to collect actually pre-saves on Spotify so thought we should maybe change it. wdyt?

intergalacticspacehighway:
> Did you test this also on small devices? iPhone SE SIM maybe?

Nope, it's breaking there pretty bad!

hirbod:
Do you have a screenshot?

hirbod (approved):
LGTM

PR #2136, opened 2023-04-27
fix: forward ref warning in new list and content type tool tip icon

intergalacticspacehighway:


packages/app/components/content-type-tooltip.tsx
```diff
@@ -41,6 +41,11 @@ export const contentGatingType = {
     text: "Pre-Save to collect",
     typeName: "Pre-Save",
   },
+  multi_provider_music_save: {
+    icon: Spotify,
```
intergalacticspacehighway, line 45: we need an icon that can show it supports apple music + spotify

PR #2140, opened 2023-04-27
fix: canceling dynamic import

alantoa:
# Why 

I removed some dynamic imports because I haven't found them to be causing any hydration issues at the moment. 

# How

# Test Plan


intergalacticspacehighway (approved):
LGTM ❇️

PR #2149, opened 2023-05-01
switch to solito router

alantoa:
# Why 

We should move [expo-next-react-navigation](https://github.com/nandorojo/expo-next-react-navigation) to [Solito](https://github.com/nandorojo/solito) because it has already been abandoned and is no longer being maintained. 

# How

Replace our `createParam` with Solito's `createParam` method. 

# Test Plan


intergalacticspacehighway (approved):
LGTM ❇️

PR #2150, opened 2023-05-01
feat:  improve modal animation on web

alantoa:
# Why 

We want to have a more graceful UX when the modal is displayed, so I added an animation for the modal.

# How 
I'm using the `Transition` component from `@headlessui/react` since we have already installed it in our project.
# Test Plan
#### Desktop

https://user-images.githubusercontent.com/37520667/235616570-f6e8a470-9002-4e8d-b359-749dc7079832.mp4

#### Mobile: 

https://user-images.githubusercontent.com/37520667/235616306-7e4e9158-f450-4a38-8d79-b5300b3fb284.MP4

### DIFF
--- a/packages/design-system/modal/modal.container.web.tsx
+++ b/packages/design-system/modal/modal.container.web.tsx
@@ -6,11 +6,11 @@ import {
   useCallback,
   forwardRef,
   useImperativeHandle,
+  Fragment,
 } from "react";
 
-import { useEscapeKeydown } from "@radix-ui/react-use-escape-keydown";
+import { Transition } from "@headlessui/react";
 
-import { useLockBodyScroll } from "@showtime-xyz/universal.hooks";
 import { View } from "@showtime-xyz/universal.view";
 
 import { WEB_HEIGHT } from "./constants";
@@ -19,7 +19,7 @@ import { ModalHeader } from "./modal.header";
 import type { ModalContainerProps, ModalMethods } from "./types";
 
 const CONTAINER_TW = [
-  "top-0 right-0 bottom-0 left-0 fixed z-[999]",
+  "fixed z-[999] inset-0",
   "flex items-center justify-end sm:justify-center",
 ];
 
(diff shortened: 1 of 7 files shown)


packages/design-system/modal/modal.container.web.tsx
```diff
@@ -61,27 +57,47 @@ const ModalContainerComponent = forwardRef<ModalMethods, ModalContainerProps>(
...
-      event.preventDefault();
-
-      return disableBackdropPress ? noop() : onClose?.();
-    });
-
-    // Prevent scrolling the body when the modal is open
-    useLockBodyScroll();
-
     return (
-      <FocusTrap aria-modal>
```
alantoa, line 74: @intergalacticspacehighway I have removed this component from the Modal because I didn't find it useful now, and it's cuasing a bug in our project, you can see [this thread](https://showtime-rq88331.slack.com/archives/C02Q0E4PBD0/p1682945414129099). 
Maybe it's because we have already upgraded RNW to v18 and switched to NativeWind. I think the previous issue no longer exists, or could you check it agian in this PR?

intergalacticspacehighway, line 74: did we remove the focus trap here?

intergalacticspacehighway (approved):
Smooth! LGTM ❇️

packages/design-system/modal/modal.container.web.tsx
alantoa: yes! see this [comment](https://github.com/showtime-xyz/showtime-frontend/pull/2150#discussion_r1182248670)

alantoa: if still exist I'll fix it again.

hirbod (approved):

packages/design-system/modal/modal.container.web.tsx
intergalacticspacehighway: @alantoa i think [this line](https://github.com/showtime-xyz/showtime-frontend/blob/staging/packages/design-system/modal/modal.backdrop.web.tsx#L19) is causing the bug. Can you check by removing it? I tried and it fixed the issue. I am not sure why onTouchEnd was used though

PR #2158, opened 2023-05-02
fix: reanimated+gesturehandler crash

intergalacticspacehighway:
# Why
- We get a crash when doing reload via dev menu. The stack trace of crash looks same when we apply codepush updates and do ExpoUpdates.reloadAsync.
- Also fixes crash caused by coinbase sdk due to failing assertion.

# How

Follows this [comment](https://github.com/software-mansion/react-native-gesture-handler/issues/2319#issuecomment-1375255933). This would make all the callback passed in gesture handler run on JS thread. Right now this seems to be a good tradeoff, we're mostly setting shared values in callbacks so this should not be a big issue. Best solution would be to migrate to v3 once it's stable enough.

# Test Plan
- Run the app, do cmd+d and press reload. App should not crash

### DIFF
--- a/apps/expo/package.json
+++ b/apps/expo/package.json
@@ -38,7 +38,7 @@
     "@amplitude/analytics-react-native": "^1.1.3",
     "@bacons/link-assets": "^1.1.0",
     "@capsizecss/core": "^3.1.0",
-    "@coinbase/wallet-mobile-sdk": "^1.0.5",
+    "@coinbase/wallet-mobile-sdk": "1.0.6",
     "@ethersproject/shims": "^5.7.0",
     "@gorhom/bottom-sheet": "4.4.5",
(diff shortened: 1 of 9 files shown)


hirbod (approved):
We should mark all of those lines with 

// TODO: remove once fixed

or something like that, so we don't forget about them

LGTM

intergalacticspacehighway:
this is the first thing i'll remove when we migrate to v3 😤

hirbod:
Tested with a prod release and confirmed that reloading does not crash anymore.
This example calls `ExpoUpdate.reloadAsync()` every 5 seconds on App start to confirm that it is actually working.

https://user-images.githubusercontent.com/504909/235762954-49555342-2cc2-4bb2-8c34-4999c121a7e4.mp4

PR #2162, opened 2023-05-03
fix: bypass username lookup validation

hirbod:
# Why

There was a flaw in the username lookup feature in both EditProfile and Onboarding. If a user typed too quickly, it was possible to bypass the check (because button would turn active on change), which could ultimately lead to an error. This is because the backend would not accept the already-taken username, but the UX would indicate something different

# How

The issue we encountered was with the debounced input. Because of the 400ms delay, `isLoading` was only triggered with the debounce, resulting in an odd button/validation state. To resolve this, we stored the last value in a ref and checked if it had changed. We then triggered `setIsLoading` separately from the actual debounced lookup.

# Test Plan

### DIFF
--- a/packages/app/components/edit-profile.tsx
+++ b/packages/app/components/edit-profile.tsx
@@ -113,7 +113,7 @@ export const EditProfile = () => {
   const router = useRouter();
   const { width } = useWindowDimensions();
   const isMdWidth = width >= breakpoints["md"];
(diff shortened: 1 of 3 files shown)


intergalacticspacehighway (approved):
LGTM ❇️

alantoa (approved):
LGTM

PR #2163, opened 2023-05-03
fix: tapping light box crashing app

intergalacticspacehighway:
# Why
Tapping lightbox in profile is crashing the app. Caused by https://github.com/showtime-xyz/showtime-frontend/pull/2158 changes

# How
gesture handler is using measure function from reanimated, so we need to wrap it in runOnUI.

# Test Plan
Tapping user profile picture should not crash the app


hirbod:
LGTM

hirbod (approved):

PR #2165, opened 2023-05-03
update patch package

alantoa:
# Why
I just checked [v7 code](https://github.com/ds300/patch-package/blob/0b9b0150a2348bf53087198fcad84e862a35eebb/src/getPackageResolution.ts#L11) and it’s using our related patch code, so I think we don’t need this patch file anymore! 

# How

# Test Plan


hirbod (approved):

intergalacticspacehighway (approved):
LGTM ❇️

PR #2167, opened 2023-05-03
fix: android reanimated crash

intergalacticspacehighway:
# Why
we've been seeing a lot of crashes on post code-push reload on android. The stack trace points to [this](https://github.com/software-mansion/react-native-reanimated/issues/1912). I also managed to reproduce it locally with reloadAsync. This PR adds a [patch](https://github.com/software-mansion/react-native-reanimated/issues/1912#issuecomment-1449278418) to fix it.

# Test Plan
Build android in release mode and trigger reloads using Updates.reloadAsync();


patches/react-native-reanimated+2.14.4.patch
```diff
@@ -144,47 +187,3 @@ index 46e3e88..b2af463 100644
...
-     component.previousStyle = currentStyle;
--    component.setNativeProps({ style: currentStyle });
-+    const domStyle = createReactDOMStyle(currentStyle);
-+    for (const key in domStyle) {
-+      if(key === 'transform') {
-+          (component.style).transform = createTransformValue(domStyle[key]);
-+      }else{
-+          (component.style)[key] = domStyle[key];
-+      }
-+    }
```
hirbod: These parts are missing and needed

hirbod (changes requested):
LGTM beside the missing web patches

patches/react-native-reanimated+2.14.4.patch
intergalacticspacehighway: added!

hirbod (approved):

PR #2175, opened 2023-05-05
fix: improve apple music auth experience on native

intergalacticspacehighway:
# Why
- When user closes auth webview modal without authorizing, the button stays `loading...`, we need to reject the promise.
- Instead of making http script request, i inlined the apple music auth javascript in the public html page. This makes it much faster.
- Request apple dev token on native app and pass as parameter to the web view. NextJS functions had a cold start issue (if not used often), adding significant delay. We could use edge functions but better to make the request from the native app itself.
- Everything is backward compatible

# Test Plan
- Apple music auth should be faster on native.


apps/next/public/apple-music-auth-native-webview/apple-music-auth-native-webview.html
```diff
@@ -1,6 +1,23 @@
 <html>
   <head>
-    <script src="https://js-cdn.music.apple.com/musickit/v3/musickit.js"></script>
+    <script>
```
intergalacticspacehighway, line 3: clearly we don't need this much JS, but there's no modularized apple music auth only SDK by apple 🤦

packages/app/screens/apple-music-auth-native-webview.tsx
```diff
@@ -1,23 +1,94 @@
...
+        }}
+      >
+        <Pressable
+          style={{
+            position: "absolute",
+            top: 0,
+            right: 0,
+            zIndex: 1,
+            padding: 8,
+          }}
```
alantoa, line 71: Is there any reason why not to use `tw=absolute right-0 top-0 z-10 p-2` ?

alantoa (approved):
LGTM ✨

packages/app/screens/apple-music-auth-native-webview.tsx
intergalacticspacehighway: no, was just playing around 😅. Will update!

PR #2178, opened 2023-05-05
feat: cherry pick some CC changes

hirbod:
# Why

I made some adjustments in CC which could already be cherry picked.

# How

# Test Plan


alantoa (approved):
LGTM

PR #2183, opened 2023-05-08
remove haptics from tabs.

intergalacticspacehighway:
# Why
On android, it's kinda annoying. It vibrates.

### DIFF
--- a/packages/app/navigation/bottom-tab-bar.tsx
+++ b/packages/app/navigation/bottom-tab-bar.tsx
@@ -8,7 +8,6 @@ import Animated, {
   useDerivedValue,
 } from "react-native-reanimated";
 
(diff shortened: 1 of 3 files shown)


alantoa (approved):
LGTM 😆

hirbod:
On iOS it was kinda neat though :D
But yeah, we can remove it.

hirbod (approved):

intergalacticspacehighway:
> On iOS it was kinda neat though :D

I think haptics should be used for some rewarding actions. Feels weird when it happens on tab presses. But let me know if you think otherwise.

hirbod:
No, I absolutely agree. Showtime was the first app I used that had these interactions on Tabs, I liked it but I totally think you're right. For example, liking with a heart could trigger a Haptic

PR #2184, opened 2023-05-08
feat: change default duration to 1 month

hirbod:
# Why

Alex asked to change default duration to 1 month instead of 1 week. This is indeed the better default option to prevent frequent time outs.

# How

# Test Plan


intergalacticspacehighway (approved):
LGTM ❇️

alantoa (approved):
LGTM

PR #2186, opened 2023-05-15
chore: update tw intellisense

alantoa:
# Why

- [prettier-plugin-tailwindcss](https://github.com/tailwindlabs/prettier-plugin-tailwindcss) support[ custom attributes PR](https://github.com/tailwindlabs/prettier-plugin-tailwindcss/pull/155) merged, so we don't need to use [this patch](https://github.com/showtime-xyz/showtime-frontend/blob/staging/patches/prettier-plugin-tailwindcss%2B0.2.5.patch) anymore!
- update TW intellisense. It now supports all `tw` regex intellisense.

image

image

image

# How

# Test Plan


.vscode/settings.json
```diff
@@ -8,13 +8,7 @@
     // already have a Prettier plugin, so don't need this.
```
hirbod, line 8: Comments are not allowed in .json. This could / should actually fail

package.json
```diff
@@ -21,7 +21,7 @@
     "nativewind": "^2.0.11",
     "patch-package": "^7.0.0",
     "prettier": "^2.8.7",
-    "prettier-plugin-tailwindcss": "0.2.5",
+    "prettier-plugin-tailwindcss": "insiders",
```
hirbod, line 24: What is this? Never seen before

hirbod (changes requested):
LGTM, but comments should be removed

.vscode/settings.json
alantoa: No, now it's the age of JSON5 and we are using it. This is not a problem.

hirbod: This does not work in every env though. But yeah, works in VSCode.

hirbod (approved):
LGTM

hirbod (approved):
LGTM

PR #2193, opened 2023-05-17
fix: auto close modal on web

alantoa:
# Why

This PR is to fix an issue that was reported from https://showtime-rq88331.slack.com/archives/C02PXGK3V8D/p1684336323639999

I found that this issue is due to the modal screen not automatically closing when using the `router.pop()` method.

# How

Automatically close the modal when URL params are lost in the router. 

# Test Plan


packages/design-system/modal-screen/with-modal-screen.tsx
```diff
@@ -20,15 +24,34 @@ function withModalScreen<P extends {}>(
...
+      closeCallback.current?.();
     }, [router]);
 
-    const contextValues = useMemo(() => ({ setTitle }), []);
+    useLayoutEffect(() => {
+      if (modalRef.current?.close) {
+        setContextValues({
+          setTitle,
+          pop: (params) => {
+            closeCallback.current = params?.callback;
```
alantoa, line 49: I added this method, which is copied from the creator-channels branch code, It's useful to do some callback when you want to wait for the modal show/hide animation to be dismissed.

PR #2197, opened 2023-05-18
fix: poll every 2 hours and not every minute

hirbod:
# Why

We don't need to poll the token every minute, it's enough to check for it every two hours
https://showtime-rq88331.slack.com/archives/C02QD3J7DJM/p1684427015334339

# How

Increased timeout duration.

@intergalacticspacehighway, we currently don't have a check for hard reloads, which means returning users or a reload would still trigger polling. Should we consider storing the token inside localStorage/MMKV and perform a validity check before initiating the polling? The response is a JWT, and we already have implemented logic for this with our AuthToken.

https://github.com/showtime-xyz/showtime-frontend/blob/9e5dde54b7ee88176f31873f9f979a162e1d2c73/packages/app/hooks/auth/use-access-token-manager.ts#L63-L77

WDYT? I think reducing every impactful request is a good thing. If you agree, feel free to add it to the PR.

# Test Plan

### DIFF
--- a/packages/app/lib/apple-music-auth/apple-music-auth.ts
+++ b/packages/app/lib/apple-music-auth/apple-music-auth.ts
@@ -23,7 +23,7 @@ export const initialiseAppleMusic = async () => {
       },
     });
 
-    // TODO: increment this expiration from backend
-    await delay(60 * 1000);
+    // 2 hours validity
+    await delay(2 * 60 * 60 * 1000);
   }
 };


intergalacticspacehighway (approved):
LGTM ❇️ 

> Should we consider storing the token inside localStorage/MMKV and perform a validity check before initiating the polling?

sure, we already store the token in localstorage, right? can we just use that?

hirbod:
Is the Apple dev token inside LS, if yes, we could do that. But I don't see it being stored inside LS in the poll function.

alantoa (approved):
LGTM

intergalacticspacehighway:
> Is the Apple dev token inside LS, if yes, we could do that. But I don't see it being stored inside LS in the poll function.

okay, I misunderstood. Yeah, will push in the next PR, I think we should do that, not a blocker for this one. LGTM.

PR #2198, opened 2023-05-18
feat: support optimized thumbnails

hirbod:
# Why

A lot of video stills are black since the first frame is fading or dark, resulting in strange-looking Lists and Grids.
Bunny creates 6 thumbnails for each video, so we're going to introduce a new key in lists, graph-images, and grids.
However, we still depend on the first frame for the video placeholders because they work best in most cases.

# How

Use the new optimized_thumbnail (only for Bunny URLs)

# Test Plan


alantoa (approved):
lgtm

intergalacticspacehighway (approved):
LGTM ❇️

PR #2202, opened 2023-05-22
feat: add optimizer to og-image for bunny uploads

hirbod:
# Why

Add transformation to Bunny uploads for our meta-images/og:graph. This should fix all size and file-size related issues, allowing our preview images to load much faster.

# How

Added a special bunny transformation class with fixed setups.
Any format -> jpg
Quality -> 70
Any width (if > 1200) = 1200

# Test Plan

### DIFF
--- a/apps/next/src/pages/nft/[chainName]/[contractAddress]/[tokenId]/index.tsx
+++ b/apps/next/src/pages/nft/[chainName]/[contractAddress]/[tokenId]/index.tsx
@@ -20,11 +20,17 @@ export async function getServerSideProps(context) {
     };
 
     const nft = res.data?.data?.item as NFT;
-    const imageUrl = getMediaUrl({
+    let imageUrl = getMediaUrl({
       nft,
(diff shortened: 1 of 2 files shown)


apps/next/src/pages/nft/[chainName]/[contractAddress]/[tokenId]/index.tsx
```diff
@@ -20,12 +20,18 @@ export async function getServerSideProps(context) {
...
       stillPreview: nft?.mime_type?.startsWith("video"),
       optimized: true,
     });
 
+    // lets check if the image is from showtime.xyz (eg Bunny,
+    // since they start with media.showtime.xyz and video.showtime.xyz)
+    if (imageUrl && imageUrl.includes("showtime.xyz/")) {
+      imageUrl = imageUrl + "?class=ogimage";
+    }
+
```
Comment (hirbod) [inline apps/next/src/pages/nft/[chainName]/[contractAddress]/[tokenId]/index.tsx:34]: This is needed for now, since we still have 3 different mixed sources of images. This can be removed once we have migrated all existing data to Bunny.

PR #2213, opened 2023-05-31
fix(android): some screens can show up with opacity

intergalacticspacehighway:
# Why
Some screens show up with opacity > 0. I think it has to do with rn-screens + when headerShown is false + fade animation. The animation of screen transition gets stuck. Changing to animation to something else fixes the issue. Right now changed it to `fade_from_bottom`. This PR adds a quick solution. I am trying to create a minimal repro to know the root cause. 

https://github.com/showtime-xyz/showtime-frontend/assets/23293248/5d6d864a-b976-478c-a570-c9a09cbb3066

# How

# Test Plan

### DIFF
--- a/packages/app/navigation/root-stack-navigator.tsx
+++ b/packages/app/navigation/root-stack-navigator.tsx
@@ -62,7 +62,7 @@ export function RootStackNavigator() {
         screenOptions={{
           headerShown: false,
           fullScreenGestureEnabled: true,
-          animation: Platform.OS === "android" ? "fade_from_bottom" : "default",
+          animationDuration: 400,
         }}
       >
         <Stack.Screen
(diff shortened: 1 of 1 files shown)


hirbod:
I also think it looks even better :) 👍

hirbod (approved):

packages/app/navigation/root-stack-navigator.tsx
```diff
@@ -62,7 +62,7 @@ export function RootStackNavigator() {
         screenOptions={{
           headerShown: false,
           fullScreenGestureEnabled: true,
-          animationDuration: 400,
+          animation: Platform.OS === "android" ? "fade_from_bottom" : "default",
```
hirbod, line 66: please add the animationDuration back for iOS

hirbod (changes requested):

PR #2215, opened 2023-06-01
feat: add support for new notification types

hirbod:
# Why

This is a cherry pick from Creator Channels, since it fixes an issue with the notification date/time being cut/pushed out of view

# How

# Test Plan


hirbod:
Nevermind, this would break currently

PR #2224, opened 2023-06-06
fix: status bar management

intergalacticspacehighway:


packages/app/lib/delete-cache.web.ts
```diff
@@ -6,4 +6,6 @@ export function deleteAppCache() {
   localStorage.removeItem("showExplanationv2");
   // TODO: showClaimExplanation is not used anymore, remove soon
   localStorage.removeItem("showClaimExplanation");
+  storage.delete("showCreatorChannelTip");
```
alantoa: can you remove this? this is my intentional change

alantoa (approved):
Just need to delete that two lines

PR #2237, opened 2023-06-07
feat: handle internal links correctly

hirbod:
# Why

Internal showtime links would open the browser inside the app, instead of redirecting

# How

# Test Plan


intergalacticspacehighway (approved):
LGTM ❇️

PR #2241, opened 2023-06-08
feat: add support for eas-updates from background

hirbod:
# Why

Our app currently only updates during a cold boot, never when transitioning from the background to the foreground. Given that many users, especially those on iOS, tend not to kill apps, we are potentially missing out on over-the-air updates.

# How

Added some checks. We agreed not showing a Snack and just update. The App must have been backgrounded at least 15min though.

# Test Plan


intergalacticspacehighway (approved):
LGTM ❇️

PR #2246, opened 2023-06-10
Fix/comments keyboard UI

hirbod:
# Why
Removed buggy InputAccessoryView for comments.

# How

Keyboard handling interactively done with react-native-keyboard-controller, like Creator channels.
Todo for future: Scroll down button, comment highlight, edit etc.

Also fixed web appearance of our alert, since buttons have been styled very inconsistently. 

Re-added autofocus. 

# Test Plan


alantoa (approved):
LGTM 💯

PR #2247, opened 2023-06-12
fix: clear input on edit message

intergalacticspacehighway:
# Why
Sometimes fast tap on edit message also sends the message. 

# How
- Disable submit if edit message is loading and also clear the input on button tap.


hirbod (approved):
LGTM

alantoa (changes requested):
LGTM

alantoa (approved):
LGTM

PR #2252, opened 2023-06-14
fix: make spotify URI mandatory

intergalacticspacehighway:
# Why
- Make Spotify URI mandatory 
- Fix ugly link from profile to drop redirect


hirbod (approved):
LGTM

alantoa (approved):
LGTM 🔥

PR #2253, opened 2023-06-14
chore: upgrade native packages

hirbod:
# Why

Fixes some perf issues and regressions and prevents from big surprises with SDK 49

# How

# Test Plan


alantoa (approved):
LGTM 🔥

PR #2255, opened 2023-06-16
Feat/new notifications

hirbod:
# Why

Our current notification settings are very bad, because we have no description.

# How

Mapped backend keys to titles and description and added some context. Also made it so that we change the settings optimistically with SWR

# Test Plan

### DIFF
--- a/packages/app/hooks/use-push-notifications-preferences.ts
+++ b/packages/app/hooks/use-push-notifications-preferences.ts
@@ -24,7 +24,6 @@ async function editPushSettings(
   url: string,
   { arg }: { arg: { pushKey: any; pushValue: boolean } }
 ) {
-  console.log(arg);
   return axios({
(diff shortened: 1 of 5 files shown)


packages/app/hooks/use-push-notifications-preferences.ts
```diff
@@ -16,3 +19,30 @@ export function usePushNotificationsPreferences() {
     refresh: mutate,
   };
 }
+
+async function editPushSettings(
+  url: string,
+  { arg }: { arg: { pushKey: any; pushValue: boolean } }
+) {
+  console.log(arg);
```
intergalacticspacehighway: 🕵️

intergalacticspacehighway (approved):
LGTM ❇️

alantoa (approved):
LGTM

PR #2256, opened 2023-06-22
fix: optimize image quality

hirbod:
# Why

Some users are not very happy with our image quality. After checking I could agree that our quality on web is not the best.

# How

Bunny offers a feature called 'sharpen', which utilizes a different algorithm (albeit a bit slower for the initial generation due to the use of bicubic transformation) to produce sharper and less fragmented images. I also increased web quality from 70 to 80.

# Test Plan


alantoa (approved):
LGTM ⚡

PR #2263, opened 2023-06-30
fix: failing testcases

intergalacticspacehighway:


hirbod (approved):

hirbod:
@intergalacticspacehighway can we merge this?

intergalacticspacehighway:
@hirbod yes, i just deleted free drop test case because the feature won't exist after star drops. I'll add the star drop test case once it's done

PR #2265, opened 2023-07-04
feat: add support for unread messages in channels list

hirbod:
# Why
Currently, we had no way to distinguish between read and unread inside our channel messages.

# How

Adds a bold text status for unread messages (API value).

Nice to have: A useful feature could be scrolling to the latest unread message, as the messages endpoint provides a return a boolean for each individual message. However, since the data is paginated, it can be challenging to locate unread messages in older list entries.

FYI: this PR don't add any orange bubble to the menu, as the endpoint for this is not ready yet. We'll have to address this later.

# Test Plan


alantoa (approved):
LGTM

PR #2266, opened 2023-07-04
feat: implement Shift-Enter / Enter Submit support by default for Message-Box

hirbod:
# Why
Now, pressing the 'Enter' key will submit our creator channels and comment box forms on the web. However, pressing 'Shift + Enter' will create a new line.

# How
Added a prop and a listener. Defaults to true.

# Test Plan


intergalacticspacehighway (approved):
LGTM ❇️

PR #2278, opened 2023-07-06
Claim status sold out if claim count is greater than edition size

jorgeavaldez:
# Why

Looks like there was a race condition and we ended up getting more claimers than the edition size. We weren't checking the edition size on the backend (lmao i goofed) so the sold out message on the button allowed claiming because we were checking equality. This should catch that case now.

# How

Change `===` to `>=`

# Test Plan

Yes

### DIFF
--- a/packages/app/components/claim/claim-button.tsx
+++ b/packages/app/components/claim/claim-button.tsx
@@ -43,8 +43,7 @@ export enum ClaimStatus {
 export const getClaimStatus = (edition?: CreatorEditionResponse) => {
   if (!edition) return undefined;
   if (
-    edition.total_claimed_count ===
-    edition.creator_airdrop_edition?.edition_size
+    edition.total_claimed_count >= edition.creator_airdrop_edition?.edition_size
   )
     return ClaimStatus.Soldout;
 


intergalacticspacehighway:
@jorgeavaldez we can merge this but i think this scenerio will not occur. the race condition might have happened imo when a lot of users opened the app at the same time (they all see 45/50). If 6 user press it then it'll go 51 and it should fail for the last user. I am trying to say that claimed size can not go above the edition size or can it?

packages/app/components/claim/claim-button.tsx
```diff
@@ -43,8 +43,7 @@ export enum ClaimStatus {
 export const getClaimStatus = (edition?: CreatorEditionResponse) => {
   if (!edition) return undefined;
   if (
-    edition.total_claimed_count ===
-    edition.creator_airdrop_edition?.edition_size
+    edition.total_claimed_count >= edition.creator_airdrop_edition?.edition_size
```
alantoa, line 46: We need to add a condition that checks if the `edition_size` is not equal to 0.

PR #2282, opened 2023-07-07
feat: enable windowScroll

alantoa:
# Why
This PR is to improve our website UX:

- Before, we used swipe.js to implement a SwiperList. However, it was not always good because it didn't have 120+ FPS and smoothness. This time, I'm using CSS snap to implement the swiper list, which should be better. Since SwiperList has been changed on the web, we can now enable `windowScroll` now. This will make scrolling more natural and reduce the need for `desktop-specific` code.

- This PR also fixed some UI issues, such as media flickering when scrolling to the next drop, UI improvements, and CSS compatibility etc.

# How

- Use [CSS snap](https://developer.mozilla.org/en-US/docs/Web/CSS/scroll-snap-type) to implemented a SwiperList. 

# Test Plan

This PR most of the changes come from the web and shouldn't affect the native. I have tested iOS, and it's not breaking anything. I think Android will be fine too, but I still need to check it once.


packages/app/components/feed/feed-item-tap-gesture.web.tsx
```diff
@@ -86,45 +86,45 @@ export const FeedItemTapGesture = ({
...
+
+  // const gesture = useMemo(
+  //   () => Gesture.Exclusive(singleTapHandle),
+  //   [singleTapHandle]
+  // );
 
   return (
     <>
-      <GestureDetector gesture={gesture}>{children}</GestureDetector>
+      <div onClick={toggleVideoPlayback}>{children}</div>
```
hirbod: Any special reason?

alantoa: this gesture is stopping CSS snap gestures, I don't have enough time to dig it now, so use click method to instead

hirbod: Ok

PR #2290, opened 2023-07-12
Fix/list appearance

hirbod:
# Why

Fixes several issues with our lists and items on home and trending.
Also added a gray background on trending for the items so empty cards won't look weird while images loading.

# How

# Test Plan


alantoa (approved):
LGTM ⚡

PR #2292, opened 2023-07-14
fix: sentry upgrade breaking build

intergalacticspacehighway:
https://docs.sentry.io/platforms/javascript/performance/instrumentation/automatic-instrumentation/


alantoa (approved):
LGTM

PR #2302, opened 2023-07-18
fix: bring instagram back

hirbod:
# Why

We have finally being approved by Meta, so we can return our Instagram connect button.

# How

# Test Plan


alantoa (approved):
LGTM

PR #2312, opened 2023-07-20
fix: do not show switch network popup

intergalacticspacehighway:
# Why

# How

Resolves - https://showtime-rq88331.slack.com/archives/C02PXGK3V8D/p1689876398632069

Connecting a wallet shows switch network option in wallet. We can prevent it by supporting all chains. We only care about the user wallet and signature, not about the chain since we don't do direct smart contract transactions.


AlexMasmej:
@intergalacticspacehighway Vercel Preview (https://showtime-4vkk7irmb-showtime.vercel.app/) does not work for me when I'm on Base, see screenshot: 
ScreenShot 2023-07-20 at 14 52 47

PR #2314, opened 2023-07-21
fix: notification icon on android was missing

hirbod:
# Why

Turns out our notification icon on Android did not work correctly, since it was missing and because we changed the icons, the way the OS generates the icon stopped working.

It has to be a monochromatic image.

# How

# Test Plan

Before:

F1jhAoXWIAA1eyJ


hirbod:
FYI: this needs a new release to the stores

PR #2316, opened 2023-07-24
fix: coinbase wallet not showing qr code

intergalacticspacehighway:
# Why
Tapping coinbase wallet in rainbowkit modal does not show QR code.

# How
- Coinbase internally uses `eth-block-tracker` during connecting wallet but it was not working as yarn resolution was forcing it to use another version.
- Remove `eth-block-tracker` from yarn resolutions.

# Test Plan
Tapping coinbase should show QR code

### DIFF
--- a/apps/next/next.config.js
+++ b/apps/next/next.config.js
@@ -76,6 +76,7 @@ const nextConfig = {
     "react-native-reanimated-carousel",
   ],
   webpack: (config, options) => {
+    config.resolve.fallback = { fs: false, net: false, tls: false };
     // Mix in aliases
     if (!config.resolve) {
       config.resolve = {};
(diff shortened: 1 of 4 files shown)


apps/next/next.config.js
```diff
@@ -76,6 +76,7 @@ const nextConfig = {
     "react-native-reanimated-carousel",
   ],
   webpack: (config, options) => {
+    config.resolve.fallback = { fs: false, net: false, tls: false };
```
intergalacticspacehighway: this is recommended in latest rainbowkit as it doesn't polyfill these deps anymore

hirbod: This is not save enough

hirbod:
```suggestion
    if (!options.isServer) {
        config.resolve.fallback = { fs: false, net: false, tls: false };
    }
```

hirbod (changes requested):

hirbod (approved):
LGTM

PR #2321, opened 2023-07-27
chore: remove walletconnect dapp

intergalacticspacehighway:
# Why
Remove wallet connect dapp dependency, we're not using it


hirbod:
Seems like there is a conflict

PR #2322, opened 2023-07-28
fix: submit on edit would create new message

hirbod:
# Why

Editing a message on Desktop and submitting via ENTER would send a new message instead of editing.

# How

Fixed by swapping the handleSubmit functions

# Test Plan


alantoa (approved):
LGTM

PR #2323, opened 2023-07-28
Fix/dedupe requests and refetches

hirbod:
# Why

Most of our SWR hooks are over-reactive and refetch way too often. I increased the delays and deduping intervals to reduce the big amount of refetching

Furthermore, I fixed an issue with search, where 1 search input would create 4 requests due to an redirecting issue

# How

# Test Plan

### DIFF
--- a/packages/app/hooks/api/use-search.ts
+++ b/packages/app/hooks/api/use-search.ts
@@ -20,7 +20,7 @@ export const useSearch = (term: string) => {
   const debouncedSearch = useDebounce(term, 200);
   const { data, error } = useSWR<SearchResponse>(
     term.length >= 2 && debouncedSearch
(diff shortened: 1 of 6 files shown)


packages/app/hooks/api/use-search.ts
```diff
@@ -20,7 +20,7 @@ export const useSearch = (term: string) => {
   const debouncedSearch = useDebounce(term, 200);
   const { data, error } = useSWR<SearchResponse>(
     term.length >= 2 && debouncedSearch
-      ? "/v2/search?q=" + debouncedSearch
+      ? "/v2/search/?q=" + debouncedSearch
```
hirbod, line 23: this bad body was causing 4 requests

alantoa (approved):
LGTM 👍

PR #2325, opened 2023-07-31
fix: url in drop screen

intergalacticspacehighway:
# Why
URL doesn't represent the currently selected drop sometimes.

### DIFF
--- a/packages/app/components/swipe-list.web.tsx
+++ b/packages/app/components/swipe-list.web.tsx
@@ -49,7 +49,7 @@ export const SwipeList = ({
 }: Props) => {
   const router = useRouter();
   const [activeIndex, setActiveIndex] = useState(0);
-  const [initialParamProp] = useParam("initialScrollIndex");
+  const [initialParamProp] = useParam("initialScrollItemId");
   const isSwipeListScreen = typeof initialParamProp !== "undefined";
   const isSwiped = useRef(false);
   const swiper = useRef<any>(null);
@@ -89,25 +89,23 @@ export const SwipeList = ({
         e.activeIndex + 1 < data.length ? e.activeIndex + 1 : undefined,
       ];
       if (isSwipeListScreen) {
-        scrollTimer.current = setTimeout(() => {
-          router.replace(
-            {
-              pathname: "/profile/[username]/[dropSlug]",
-              query: {
-                ...router.query,
-                initialScrollIndex: activeIndex,
-                username: data[activeIndex].creator_username,
-                dropSlug: data[activeIndex].slug,
-              },
+        router.replace(
+          {
+            pathname: "/profile/[username]/[dropSlug]",
+            query: {
+              ...router.query,
+              initialScrollItemId: data[e.activeIndex].nft_id,
+              username: data[e.activeIndex].creator_username,
+              dropSlug: data[e.activeIndex].slug,
             },
-            getNFTSlug(data[activeIndex]),
-            { shallow: true }
-          );
-        }, 700);
+          },
+          getNFTSlug(data[e.activeIndex]),
+          { shallow: true }
+        );
       }
       setActiveIndex(e.activeIndex);
     },
-    [router, visibleItems, data, isSwipeListScreen, activeIndex]
+    [router, visibleItems, data, isSwipeListScreen]
   );
 
   if (data.length === 0) return null;
--- a/packages/app/screens/swipe-list.tsx
+++ b/packages/app/screens/swipe-list.tsx
@@ -1,3 +1,4 @@
+import { useMemo } from "react";
 import { Platform } from "react-native";
 
 import { useSafeAreaInsets } from "@showtime-xyz/universal.safe-area";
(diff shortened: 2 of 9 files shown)


packages/app/components/swipe-list.web.tsx
```diff
@@ -89,25 +89,23 @@ export const SwipeList = ({
...
-                initialScrollIndex: activeIndex,
-                username: data[activeIndex].creator_username,
-                dropSlug: data[activeIndex].slug,
-              },
+        router.replace(
+          {
+            pathname: "/profile/[username]/[dropSlug]",
+            query: {
+              ...router.query,
+              initialScrollItemId: data[e.activeIndex].nft_id,
```
intergalacticspacehighway, line 97: root issue was combination of setTimeout and using activeIndex state instead of e.activeIndex here. Other changes in the PR are for replacing index with nft_id

packages/app/screens/swipe-list.tsx
```diff
@@ -85,6 +89,10 @@ const ProfileSwipeList = () => {
   );
   const { bottom: safeAreaBottom } = useSafeAreaInsets();
 
+  const initialScrollIndex = data?.findIndex(
+    (item) => item.nft_id == initialScrollItemId
+  );
+
```
hirbod: this should useMemo

hirbod (changes requested):
One small issue I see, otherwise LGTM

packages/app/screens/swipe-list.tsx
hirbod: And there might be circumstances, where findIndex return -1 (nft deleted, hidden, server etc), which should be handled gracefully with 0. FlashList breaks with a negative index.

packages/app/components/swipe-list.web.tsx
hirbod: The timer most likely had a purpose, right? Are we running into a regression maybe?

alantoa: Yeah, the timer is for fixing the CSS swipe list flicker issue. But this is the swipe.js version, so removing or adding it are both good options.

packages/app/screens/swipe-list.tsx
```diff
@@ -55,12 +55,16 @@ export const SwipeListScreen = withColorScheme(() => {
 const FeedSwipeList = () => {
   const { useParam } = createParam<Query>();
   const { data } = useFeed();
-  const [initialScrollIndex] = useParam("initialScrollIndex");
+  const [initialScrollItemId] = useParam("initialScrollItemId");
   const { bottom: safeAreaBottom } = useSafeAreaInsets();
+  const initialScrollIndex = data?.findIndex(
+    (item) => item.nft_id == initialScrollItemId
+  );
+
```
hirbod: why is this duplicated?

hirbod (changes requested):

packages/app/screens/swipe-list.tsx
hirbod: Already in line 92

intergalacticspacehighway: yeah, they are in different components. do you mean something else?

hirbod: Missed that, just saw the same file name. So this should also be handled as remarked before (useMemo and handle -1 cases)

packages/app/screens/swipe-list.tsx
```diff
@@ -89,9 +91,10 @@ const ProfileSwipeList = () => {
   );
   const { bottom: safeAreaBottom } = useSafeAreaInsets();
 
-  const initialScrollIndex = data?.findIndex(
-    (item) => item.nft_id == initialScrollItemId
-  );
+  const initialScrollIndex = useMemo(() => {
+    const index = data?.findIndex((item) => item.nft_id == initialScrollItemId);
```
hirbod, line 95: how you handle data if undefined? Or initialScrollItemId was tampered with?
The optional chain would fail

intergalacticspacehighway: if data is undefined `initialScrollIndex` would be undefined and won't be passed in the list so it'll default to 0 in swipe-list and if `initialScrollItemId` is tampered then it'll be -1 and so will default to 0 so seems okay to me.

hirbod: But if data is undefined, findIndex would fail with "data is undefined".
It should be 

```
const index = data ? data.findIndex.. : 0
```

hirbod: Or will SWR make sure to always return an empty array?

hirbod (approved):
LGTM

packages/app/screens/swipe-list.tsx
intergalacticspacehighway: oh try this. 
x = undefined; x?.findIndex(); 
this won't fail but will be returned as undefined. The only time it can fail is if data is not an array and not undefined. SWR will return undefined or an array.

hirbod: LGTM

hirbod: I forgot about the hook definition, brain 404 today

PR #2326, opened 2023-08-02
chore: move KV into SWR

alantoa:
# Why

Currently, we are using the "@vercel/kv" library to fetch data from our kv server. However, it is not synchronized with SWR, resulting in some loading issues. 

# How

- Move KV request into SWR. 
- Remove the "@vercel/kv" library, we don't need it anymore.

# Test Plan

https://github.com/showtime-xyz/showtime-frontend/assets/37520667/823241ab-c10f-4ff4-a181-b532d46c8834


hirbod (approved):
LGTM

hirbod:
Preview seems to fail @alantoa 
https://showtime-git-chore-move-kv-into-swr-showtime.vercel.app/

PR #2345, opened 2023-08-11
fix: collect to unlock changes

intergalacticspacehighway:
# Why
https://linear.app/showtime/issue/SHOW2-2007/collect-to-unlock-channel-small-product-changes

# How
Gets NFT details, edition by slug, makes sure that performance is only affected for users who are not subscribed in the channel. We won't need API call once we receive edition in the API. Reported to backend!

# Test Plan
- Try collecting to unlock button in channel


packages/app/hooks/use-infinite-list-query.ts
```diff
@@ -42,7 +42,7 @@ export const useInfiniteListQuerySWR = <T>(
       // suspense: true,
       refreshInterval,
       revalidateOnMount: true,
-      dedupingInterval: 30000,
+      dedupingInterval: 5000,
```
intergalacticspacehighway, line 45: this was required or else doing mutate doesn't perform refetch (we need to refetch once user has collected the star drop from default selected card). I think 5 seconds is good enough for dedup interval. cc - @hirbod let me know what you think!

hirbod: Thats fine. I didn't know it had that many implications

packages/app/components/creator-channels/messages.tsx
```diff
@@ -536,7 +545,7 @@ export const Messages = memo(() => {
           editMessage={editMessage}
           isUserAdmin={isUserAdmin}
           keyboard={keyboard}
-          latestPaidNFTSlug={latest_paid_nft_slug}
+          edition={edition}
```
hirbod, line 548: why did you remove latestPaidNFTSlug? This is required for the different states inside the MessageInput component

hirbod: Ok I have seen the changes, but is that safe enough?

intergalacticspacehighway: yeah should be good. Not perfect. Ideal solution would be to wait for BE to add edition in the collection detail API. We can re-iterate once they do. wdyt?

packages/app/hooks/use-infinite-list-query.ts
intergalacticspacehighway: yeah, it is not obvious. So if mutate is triggered within 30 secs SWR counts it as a duplicate request and doesn't refetch. Took a while to debug

packages/app/components/creator-channels/messages.tsx
hirbod: But I am using the channel by detail endpoint to determine if there is any paid nft, why should there be the need to use the edition for it?

intergalacticspacehighway: We need the edition to trigger the pay flow directly instead of going to the drop page on `Collect to unlock` button tap. The edition is required by our Pay button component. If the channel by detail endpoint can also return star drop edition that would save us 2 requests that I am [making.](https://github.com/showtime-xyz/showtime-frontend/pull/2345/files#diff-35e8b62f9b70f65e462c40a883cc47ea668799d3de4585b0d46b0c098b8b6b8fR189)

alantoa (approved):
LGTM 🔥

PR #2351, opened 2023-08-14
feat: add some delay for claim call

intergalacticspacehighway:
https://showtime-xyz.slack.com/archives/C02QD3J7DJM/p1691982390108209?thread_ts=1691980287.122139&cid=C02QD3J7DJM

https://linear.app/showtime/issue/SHOW2-2017/frontend-wait-between-the-end-of-start-payment-and-calling-claim


hirbod:
I don't think that this is a silver bullet to the issue because 3s could also fail. Magic numbers are a bad idea most of the time.

intergalacticspacehighway:
> I don't think that this is a silver bullet to the issue because 3s could also fail. Magic numbers are a bad idea most of the time.

yup, status API would be the ideal solution.

hirbod:
@intergalacticspacehighway can we kill this PR with the new polling?

PR #2360, opened 2023-08-15
Fix/usd price conversion drops

hirbod:
# Why

This PR adds the output of USD if `usd_price` is available and fallback to the provided one if `null` (rare case that should not happen)

# How

# Test Plan


alantoa (approved):
LGTM

hirbod:
FYI: no changes to the create pricing flow (remains in the user currency)

PR #2363, opened 2023-08-15
fix: add status api in star drops

intergalacticspacehighway:
# Why
- Poll status API before claiming the drop.

### Claim can happen after below cases. Tested below cases.
- 409 in /start API - User already paid but didn't claim - **Don't Poll**
- Hard redirect/refresh by stripe - **Poll**
- By selecting default card - no hard redirects - **Poll**

Tested above flow, there is surely some opportunity to clean things up, but pushing this for now to unblock backend.


alantoa (approved):
Lgtm 🔥

PR #2366, opened 2023-08-15
fix: only apply EAS updates instantly on iOS and exclude android

hirbod:
# Why

Since Android updates crash when the runtime reloads, I've excluded it for Android. This change will still download the newest update but will only apply it after the next app reload.

While this is not the ideal solution I am looking for, our first-time app crash rate is so extremely high on Android that we need to take this path.

# How

# Test Plan


Nasseratic (dismissed):
🎉 ❤️

alantoa (approved):
LGTM

intergalacticspacehighway (approved):

PR #2369, opened 2023-08-17
Fix/star drop close modal

intergalacticspacehighway:
# Why
- Star drop form close button doesn't work on native.

# How
- Replace modalContext.pop with router.pop as we're using native modal on iOS. On Android, it uses modalContext.pop. We can revisit the internal modal to handle such a case. Pushing this right now.

### DIFF
--- a/packages/app/screens/payouts/setup.tsx
+++ b/packages/app/screens/payouts/setup.tsx
@@ -10,6 +10,4 @@ export const PayoutsSetupScreen = withModalScreen(PayoutsSetup, {
   disableBackdropPress: true,
   snapPoints: ["100%"],
   headerShown: false,
-  enableContentPanningGesture: false,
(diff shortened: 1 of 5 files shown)


packages/app/screens/payouts/setup.tsx
```diff
@@ -10,6 +10,4 @@ export const PayoutsSetupScreen = withModalScreen(PayoutsSetup, {
   disableBackdropPress: true,
   snapPoints: ["100%"],
   headerShown: false,
-  enableContentPanningGesture: false,
```
intergalacticspacehighway, line 13: enabled closing with gesture as we're saving the form state now.

alantoa (approved):
LGTM 👍

hirbod (approved):
LGTM

PR #2370, opened 2023-08-17
Feature/show2 2057 not a big deal but when i saw 2m my first thought was that

alantoa:
# Why
When I saw 2m, my first thought was that it meant “2 minutes ago”. 

# How 

Add the "isDisplayCompleteUnit" parameter to support displaying a full time unit instead of a short one.
# Test Plan

### DIFF
--- a/packages/app/components/creator-channels/components/message-item.tsx
+++ b/packages/app/components/creator-channels/components/message-item.tsx
@@ -176,7 +176,7 @@ export const MessageItem = memo(
 
               <View tw="flex-row items-center">
                 <Text tw={["text-xs text-gray-700 dark:text-gray-200"]}>
-                  {formatDateRelativeWithIntl(channel_message.created_at)}
+                  {formatDateRelativeWithIntl(channel_message.created_at, true)}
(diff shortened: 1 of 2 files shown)


hirbod:
No, this will have too many implications, as it might break in too many lists like comments, notifications etc.
We need shorter conventions

hirbod (approved):
Oh you actually added a param to control this.
LGTM

packages/app/components/creator-channels/components/message-item.tsx
```diff
@@ -176,7 +177,10 @@ export const MessageItem = memo(
 
               <View tw="flex-row items-center">
                 <Text tw={["text-xs text-gray-700 dark:text-gray-200"]}>
-                  {formatDateRelativeWithIntl(channel_message.created_at)}
+                  {formatDateRelativeWithIntl(
+                    channel_message.created_at,
+                    !isMobileWeb()
```
alantoa, line 182: @hirbod this param is false by default, and here i set the desktop only. so they will not have much effect.

PR #2371, opened 2023-08-17
fix copy exteeds

alantoa:
# Why

The 'Announce Your Raffle' copy is broken when the drop is paid gating because the other side has a 'View Channel' button.

 
image

# How

Make the copy shorter when it is a paid drop. 
# Test Plan

### DIFF
--- a/packages/app/components/claim/claim-button.tsx
+++ b/packages/app/components/claim/claim-button.tsx
@@ -179,7 +179,7 @@ export const ClaimButton = ({
       <Button {...buttonProps} onPress={handleRaffleResultPress}>
         <>
           <Text tw={["font-semibold text-white", LABEL_SIZE_TW[size]]}>
-            Announce your raffle
+            {isPaidGated ? "Announce" : "Announce your raffle"}
           </Text>
         </>
       </Button>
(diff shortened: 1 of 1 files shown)


packages/app/components/claim/claim-button.tsx
```diff
@@ -179,7 +179,7 @@ export const ClaimButton = ({
       <Button {...buttonProps} onPress={handleRaffleResultPress}>
         <>
           <Text tw={["font-semibold text-white", LABEL_SIZE_TW[size]]}>
-            Announce your raffle
+            {isPaidGated ? "Announce" : "Announce your raffle"}
```
hirbod:
```suggestion
            {isPaidGated ? "Announce winner" : "Announce your raffle"}
```

hirbod: Would announce winner fit?

hirbod (approved):
LGTM, with a tiny question

packages/app/components/claim/claim-button.tsx
alantoa: yeah, that's would be great!

PR #2372, opened 2023-08-17
feat: onchain home carousel

hirbod:
# Why

We need a temp onchain summer carousel at top.

# How

Added a dedicated component for the on-chain summer. Furthermore, adjusted our feed to inject the trending carousel every 25% of scroll percentage.

Need to wait for @jorgeavaldez because /v1/onchain-summer returns 404 - so couldn't test it yet

# Test Plan


hirbod:
FYI @alantoa @intergalacticspacehighway - just in case I can't make it. This needs to be merged by tomorrow, 18th of August, at 2:30pm CET, and released by 2:45pm CET.

I will most likely handle it myself, but I'm informing you just in case.

PR #2373, opened 2023-08-17
fix: prevent onboarding trigger for star drop collectors

intergalacticspacehighway:
# Why
Add `latest_star_drop_collected` check


hirbod (approved):
LGTM

PR #2384, opened 2023-08-18
fix: showing onboarding modal on collecting

intergalacticspacehighway:
# Why

# How

# Test Plan


packages/app/components/checkout/checkout-return-for-paid-nft.tsx
```diff
@@ -29,19 +32,27 @@ const { useParam } = createParam<{
 export const CheckoutReturnForPaidNFT = () => {
   const [contractAddress] = useParam("contractAddress");
   const { data: edition } = useCreatorCollectionDetail(contractAddress);
-  if (!edition)
+  const { data: nft } = useNFTDetailByTokenId({
```
intergalacticspacehighway, line 35: had to fetch the nft as the latest_star field needs some values from it.

hirbod (approved):
LGTM

PR #2391, opened 2023-08-22
feat: kill claim modal for pre-save drop

alantoa:
# Why

As Alex said, we want to remove the claim modal to reduce unnecessary clicks for users. 

# How

Moving all claim button logic into the `useClaimDrop` hooks.

# Test Plan

https://github.com/showtime-xyz/showtime-frontend/assets/37520667/2f5cd37d-66e6-46fc-a683-ced77c33558c

### DIFF
--- a/packages/app/components/claim/claim-button.tsx
+++ b/packages/app/components/claim/claim-button.tsx
@@ -20,20 +20,17 @@ import { View } from "@showtime-xyz/universal.view";
 import { ButtonGoldLinearGradient } from "app/components/gold-gradient";
 import { ClaimContext } from "app/context/claim-context";
 import { useAppleMusicGatedClaim } from "app/hooks/use-apple-music-gated-claim";
+import { useClaimDrop } from "app/hooks/use-claim-drop";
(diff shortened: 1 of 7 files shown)


alantoa:
@intergalacticspacehighway You should have more context about this flow. Could you please help to take a look?

packages/app/components/claim/claim-button.tsx
```diff
@@ -336,7 +298,7 @@ export const ClaimButton = ({
         {edition?.apple_music_track_url ? (
           <Button
             {...buttonProps}
-            onPress={() => handleCollectPress("appleMusic")}
+            onPress={handleClaimNFT}
```
intergalacticspacehighway: > handleCollectPress("appleMusic")

this button explicitly passes appleMusic. i.e. the user has pressed apple music so it triggers apple music flow because a pre save drop can have apple music URL as well as spotify URL. I think we'll have to handle it in handleClaimNFT

alantoa: Oh yes, that makes sense. I forgot about that.

alantoa: fixed!

PR #2395, opened 2023-08-23
fix: add a primary wallet check for OG users to prevent breaking claim flow

hirbod:
# Why

We have OG users who haven't been online for some time. Some of them don't have a primary wallet set, which we require for both paid and free music NFTs.

# How

Added a check to our buttons to show an alert when there is no primary wallet.

# Test Plan


alantoa (approved):
LGTM

PR #2399, opened 2023-08-24
Fix/default payment

alantoa:
# Why

# How

# Test Plan

### DIFF
--- a/packages/app/components/checkout/checkout-claim-form.tsx
+++ b/packages/app/components/checkout/checkout-claim-form.tsx
@@ -117,17 +117,17 @@ const CheckoutFormLayout = ({
   });
   const router = useRouter();
   const paymentMethods = usePaymentsManage();
-  const defaultPaymentMethod = useMemo(
-    () => paymentMethods.data?.find((method) => method.is_default),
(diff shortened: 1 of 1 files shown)


packages/app/components/checkout/checkout-claim-form.tsx
```diff
@@ -117,17 +117,17 @@ const CheckoutFormLayout = ({
...
   );
-  const [savedPaymentMethodId, setSavedPaymentMethodId] = useState(
-    defaultPaymentMethod?.id
+  const defaultPaymentMethod = useMemo(
+    () => paymentMethods.data?.find((method) => method.is_default),
+    [paymentMethods.data]
   );
+  const [savedPaymentMethodId, setSavedPaymentMethodId] = useState<
+    undefined | string
+  >(defaultPaymentMethod?.id);
```
intergalacticspacehighway, line 130: Do we need to call `setSavedPaymentMethodId` when the `paymentMethods` API call succeeds to set the initial state?

alantoa:
@intergalacticspacehighway I have fixed it in another commit. This is not a correct PR, so just ignore it

PR #2401, opened 2023-08-25
feat: support video on home v2(web & iOS only)

alantoa:
# Why

Currently, we haven't supported video on home v2 yet. This PR enables auto play of videos on home v2. However, this change is only for the Web and iOS platforms because there is a performance issue on Android. We will add it back once Hirbod solves the Android player issue. 

# How  

Use React context and the `ViewabilityInfiniteScrollList` component to automatically play videos when scrolling through the current video content.

# Test Plan

I have checked the web and iOS, and it works well! and this change doesn't affect Android.

https://github.com/showtime-xyz/showtime-frontend/assets/37520667/f843ba09-4d36-4dfc-a311-2c42942e908f


intergalacticspacehighway (approved):
LGTM ❇️

PR #2404, opened 2023-08-29
fix: add more onboardinginitiated checks

intergalacticspacehighway:
# Why
Add new fields that make sure user has submitted stripe details


alantoa (approved):
LGTM

PR #2406, opened 2023-08-29
Remove the onchain summer section from homev2.

alantoa:
# Why 

This PR is just to remove the onchain summer section from homev2 because all of the drops have expired. 

# How

# Test Plan


hirbod (commented):
Have you also removed the getItemType for index 1?

alantoa:
@hirbod oh yeah nice catch I forgot that. will update

PR #2407, opened 2023-08-30
fix: tooltip issues

alantoa:
# Why

I just released a new [preversion](https://github.com/alantoa/universal-tooltip/releases/tag/1.0.0-2) of the tooltip component to fix some previous issues:

- The tooltip can't close when scrolling the view.

- The tooltip component duplicates itself when switching apps. 

And also, I switched the iOS library to https://github.com/aheze/Popovers, which has a better UX and API.

Besides, I have updated a few native components, which benefits us to do some OTA updates. 
# How

- Upgrade the universal-tooltip component version and switch to a new API. 

# Test Plan

I tested Android and iOS, and they both work well!

### DIFF
--- a/apps/expo/package.json
+++ b/apps/expo/package.json
@@ -172,7 +172,7 @@
     "tinycolor2": "^1.6.0",
     "tldts": "^6.0.12",
     "tty-browserify": "0.0.1",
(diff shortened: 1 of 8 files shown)


intergalacticspacehighway (approved):
LGTM 🔥

PR #2410, opened 2023-08-31
fix: remove lagacy code

alantoa:
# Why

This PR is just for removing some legacy code because we don't need it anymore. 

# How

# Test Plan


intergalacticspacehighway (approved):
LGTM ❇️

PR #2423, opened 2023-09-05
reassign wallet on error code 409

intergalacticspacehighway:
# Why
If a wallet is already added, we receive an error code of 409. On receiving that we need to show a confirmation to reassign the wallet to the new profile.


alantoa (approved):
LGTM! Btw will this be removed once we get rid of Magic, right?

PR #2427, opened 2023-09-11
fix: canCreateMusicDrop expression was broken

hirbod:
# Why
Fix a regression caused by #2426 

# How

The expression was broken.

# Test Plan


intergalacticspacehighway (approved):
LGTM ❇️

PR #2431, opened 2023-09-16
chore: upgrade important packages

hirbod:
# Why
Upgrades some important packages (specially Reanimated)

# How

# Test Plan


hirbod:
Seems like reanimated upgrade breaks our Tab view component - or its Expo Image. I'll have to make a dedicated test.

PR #2432, opened 2023-09-16
fix: Text and View rendering perf

hirbod:
# Why

According to [https://twitter.com/hirbod_dev/status/1703023347477434440](https://twitter.com/hirbod_dev/status/1703023347477434440) (referencing [https://twitter.com/hirbod_dev/status/1703025832518664404](https://twitter.com/hirbod_dev/status/1703025832518664404)), we can achieve up to 90% faster text rendering performance. The downside is that there is no visual feedback for onPress, but accessibility is currently not a key focus for us. I have introduced a new StockText which retains the old behavior if needed.

We need to use StockText when we have multiline text (like notifications or channels) only for @ mentions, all others use the new lean one. Furthermore, I have changed how `View` works by just using a RCTView, this reduces the rendering by another 30%.

# How
Bildschirmfoto 2023-09-16 um 17 43 51

Bildschirmfoto 2023-09-16 um 22 12 35

# Test Plan


hirbod:
Closing this, because the approach is too aggressive. I've added a LeanText and LeanView, which should be used on a case-by-case basis, and maybe just for lists, because this is where they are the most impactful.

PR #2444, opened 2023-09-29
feat: update release date

intergalacticspacehighway:
# Why
- Adds update release date feature back.
- Uses a [different](https://showtime-xyz.slack.com/archives/C02QD3SU4BT/p1695920477017039?thread_ts=1695913935.666369&cid=C02QD3SU4BT) API, so not a great solution but works for now.

# Test Plan
Tested on iOS, web and android


hirbod (approved):
LGTM

PR #2451, opened 2023-10-09
fix: no excessive sessionStorage reads

### LINKED ISSUE #2450: Reading `sessionStorage` on every render
https://github.com/showtime-xyz/showtime-frontend/blob/c15e7af2eee3bd73828744ff28dd2d42ad52a54b/packages/design-system/infinite-scroll-list/infinite-scroll-list.web.tsx#L100

This code is reading `sessionStorage` on every render when `preserveScrollPosition` is true fyi. Not ideal for performance.

hirbod:
# Why

Fixes #2450 

# How

# Test Plan


intergalacticspacehighway (approved):
LGTM ❇️

PR #2459, opened 2023-10-22
fix: add root shared value to as dep instead of .value

### LINKED ISSUE #2458: Remove `.value` from deps array
https://github.com/showtime-xyz/showtime-frontend/blob/7ed6e633118995be284fd619828e8abc824bb8e2/packages/app/components/creator-channels/components/message-item.tsx#L117C38-L117C38

Adding `.value` here breaks the Web implementation of Reanimated. The deps array should have the root shared value.

hirbod:
# Why
The `.value` should be removed from the `deps` array when using `useSharedValue`. This is necessary for it to work on the web.

Fixes #2458 

Thanks, @nandorojo!

# How

Removed .value and use the root shared value.

# Test Plan

### DIFF
--- a/packages/app/components/creator-channels/components/message-item.tsx
+++ b/packages/app/components/creator-channels/components/message-item.tsx
@@ -114,7 +114,7 @@ export const MessageItem = memo(
         backgroundColor: "transparent",
         borderRadius: 0,
       };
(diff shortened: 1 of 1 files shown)


intergalacticspacehighway (approved):
LGTM ❇️

alantoa (approved):
LGTM!

PR #2465, opened 2023-10-27
fix: open edit wallet nickname modal

alantoa:
# Why
For the Modal component, if it is opened, you need to pass a `visible` field. 
Currently, if it doesn't pass this field, it will cause an error because the `Transition` component only accepts `true` or `false`. 

# How

I just made "visible" false by default. This change will no longer cause any broken.

# Test Plan


hirbod (approved):
LGTM

PR #2467, opened 2023-10-27
fix: redirect to settings on set primary on creating creator token

intergalacticspacehighway:
Workaround until we replace magic with privy


hirbod (approved):
LGTM - good workaround for now

PR #2471, opened 2023-10-30
Use the ChatGPT generated gas price padding function when purchasing a creator token

jorgeavaldez:
# Why

The simulated contract function request payload seems to have discrepancies with the quoted gas price between the Alchemy rpc and the user wallet rpc.

# How

https://showtime-xyz.slack.com/archives/C02PXGK3V8D/p1698699732510189

I took whatever chat gpt did here and added it in the `useCreatorTokenBuy` hook.

This doesn't seem correct though. Because this is an eip 1559 transaction but gas price is [not an eip 1559](https://viem.sh/docs/contract/writeContract.html#gasprice-optional) thing.

We probably need to do this to the following:
- https://viem.sh/docs/contract/writeContract.html#maxfeepergas-optional
- https://viem.sh/docs/contract/writeContract.html#maxpriorityfeepergas-optional
- Potentially https://viem.sh/docs/contract/writeContract.html#gas-optional the raw gas param, but this turns off the auto estimate which is also not good...

# Test Plan

Staging

### DIFF
--- a/packages/app/hooks/creator-token/use-creator-token-buy.ts
+++ b/packages/app/hooks/creator-token/use-creator-token-buy.ts
@@ -24,6 +24,24 @@ import {
 import { useSwitchChain } from "./use-switch-chain";
 import { baseChain } from "./utils";
 
+type BigIntish = string | number | bigint; // Define BigIntish type
+const paddedGasPrice = (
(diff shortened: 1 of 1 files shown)


hirbod:
There is no gasPrice returned from simulation, this has to be done differently

PR #2478, opened 2023-10-31
fix: disable truncation of usernames

alantoa:
# Why

We don't want usernames to be truncated. So, this PR is just to disable the truncation of usernames on the profile screen and improve a bit of the UI in the profile top section. 

# How

# Test Plan


intergalacticspacehighway (approved):
LGTM ❇️

hirbod (approved):

PR #2484, opened 2023-11-01
feat: auto-pop CT modal if user is in allowlist

alantoa:
# Why 

This PR just made two changes: 
- Show the Creator Token explanation banner every time, even if the user clicks the X button.
- Auto-pop the Creator Token self-serve explainer modal if the user is in the allowlist. 

# How

# Test Plan


intergalacticspacehighway (approved):
LGTM ❇️

PR #2490, opened 2023-11-01
Feat/creator tokens invite system

hirbod:
# Why

The info circle was opening the wrong modal instead of showing the designed tooltip

# How

Replaced modal opening against tooltip

# Test Plan


hirbod:
Damn, wrong branch

PR #2492, opened 2023-11-01
Show channel messages button on profile screen even if empty

jorgeavaldez:
# Why

Screenshot 2023-11-01 at 11 24 45 AM

I deleted a channel message locally but it was the only message in the channel. Then it looked like the channel did not exist. I did see it in the channels list though.

This is a problem only because i fixed https://linear.app/showtime/issue/SHOW2-2303/deleted-messages-are-included-in-profile-channel-message-count

Now we can see it:
Screenshot 2023-11-01 at 11 32 17 AM

# How

Validate message count is `>= 0` to display

# Test Plan

Locally and in staging

### DIFF
--- a/packages/app/components/creator-channels/types.ts
+++ b/packages/app/components/creator-channels/types.ts
@@ -154,15 +154,17 @@ export type Channel = {
   owner: ChannelProfile;
 };
 
+export type ChannelPermissions = {
+  can_send_messages: boolean;
+  can_upload_media: boolean;
+  can_view_creator_messages: boolean;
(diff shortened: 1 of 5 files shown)


hirbod:
@jorgeavaldez I actually did this on purpose. There is no need to show 0 channel messages imho.

jorgeavaldez:
Screenshot 2023-11-01 at 12 27 10 PM

Changed it to this at Alex's request

hirbod (approved):

PR #2494, opened 2023-11-01
feat: reenable import allowlist to channel

alantoa:
# Why

# How

# Test Plan


hirbod:
Added loading state, re-enabled AccessTicket Icon in channel header

PR #2501, opened 2023-11-03
feat: top creator token

alantoa:
# Why

This PR is for adding the Top creator token to the profile and home screen,  and made some improvements to the UI of the creator token. 

# How

# Test Plan

  

⚠️ Need to merge it after the backend has deployed it!

CleanShot 2023-11-06 at 17 08 48@2x
CleanShot 2023-11-06 at 17 08 52@2x


intergalacticspacehighway (approved):
LGTM ❇️

PR #2508, opened 2023-11-07
Automatically adjust top creator tokens quantity

alantoa:
# Why
this PR includes: 

- Automatically adjust the quantity of top creator tokens following the screen height.
- Made some adjustments to the UI of Top Creator Tokens, as reported by @hirbod.

# How

# Test Plan


hirbod (approved):
Love it, much cleaner now, and better UX

PR #2520, opened 2023-11-09
fix: connect wallet options are broken

alantoa:
# Why
The style of the connect wallet option item is broken, so I will fix it! 

# Before
CleanShot 2023-11-09 at 23 20 03@2x
# After 
CleanShot 2023-11-09 at 23 19 49@2x

# How

# Test Plan


intergalacticspacehighway (approved):
LGTM 🫡

PR #2542, opened 2023-11-20
fix: redirect auth bug

intergalacticspacehighway:
# Why
- Better fix for - https://github.com/showtime-xyz/showtime-frontend/pull/2541
- Adds a `useAuthScreen` hook, that can be used in screens that require authentication.
- Cleans up `useUser` hook

### DIFF
--- a/packages/app/components/settings/setting.web.tsx
+++ b/packages/app/components/settings/setting.web.tsx
@@ -23,10 +23,7 @@ const SettingsTabs = () => {
     WalletAddressesV2 | undefined
   >(undefined);
   const { width } = useWindowDimensions();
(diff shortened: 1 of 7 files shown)


packages/app/components/settings/setting.web.tsx
```diff
@@ -23,10 +23,7 @@ const SettingsTabs = () => {
     WalletAddressesV2 | undefined
   >(undefined);
   const { width } = useWindowDimensions();
-  const { isAuthenticated } = useUser({
-    redirectTo: "/login",
```
intergalacticspacehighway, line 27: moved the redirection to screens/settings.tsx like everywhere else

PR #2553, opened 2023-11-24
chore: hide connected apps tab

intergalacticspacehighway:
Hide connected apps tab and move twitter/ig disconnect to edit profile

### DIFF
--- a/packages/app/components/edit-profile.tsx
+++ b/packages/app/components/edit-profile.tsx
@@ -1,4 +1,10 @@
-import React, { useEffect, useMemo, useState, useRef } from "react";
+import React, {
+  useEffect,
+  useMemo,
+  useState,
+  useRef,
+  useCallback,
+} from "react";
 import {
   Platform,
   useWindowDimensions,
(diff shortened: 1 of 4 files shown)


packages/app/components/edit-profile.tsx
```diff
@@ -261,6 +265,14 @@ export const EditProfile = () => {
...
     [width]
   );
+  const accounts = useListSocialAccounts();
+
+  const instagramProviderId = accounts.data?.find?.(
+    (v) => v.provider === "instagram"
+  )?.provider_account_id;
+  const twitterProviderId = accounts.data?.find?.(
+    (v) => v.provider === "twitter"
+  )?.provider_account_id;
```
hirbod: I'd wrap this in useMemo, even though the screen is not very expensive, but every button press re-renders, so its better to do

packages/app/components/edit-profile.tsx
```diff
@@ -623,25 +637,62 @@ const ConnectButton = ({
...
   return (
     <PressableHover
       onPress={async () => {
-        try {
-          await trigger({
-            type,
-          });
-        } catch {
-          // do nothing
+        if (isConnected) {
```
hirbod: same

hirbod (changes requested):

hirbod (approved):
LGTM

